PR #7, opened 2016-11-23
18: Config whitelist

katiefenn:
- Created whitelist for client-side whitelist
- Moved config-related functions to a single helper file


katiefenn:
I've also renamed the client-side config file:

`src/common/config/index.js` -> `src/common/helpers/config.js`

This reflects the server-side config helper, and is meant to convince unwary developers not to try to import the server-side helper in client-side. This isn't actually possible at present (trying to include nconf on the client throws all sorts of errors), but I think it's worth making this clear.

src/server/helpers/config.js
```diff
@@ -0,0 +1,40 @@
...
+nconf.env({
+  separator: '__'
+});
+
+export const conf = nconf;
+
+export const getClientConfig = () => {
+  let configForClient = {};
+
+  CLIENT_SIDE_WHITELIST.forEach(item => {
```
bartaz: Do we need to build new `configForClient` each time `getClientConfig` is called? Are we expecting config to change in runtime after this file is loaded?

katiefenn: Yes, configForClient is declared and items assigned to it each time getClientConfig is run. I didn't think too hard about optimising this at the time because config is managed in memory. 

However, we could declare configForClient and assign items once when the module is created, and memoize the result.

PR #8, opened 2016-11-24
GitHub repository input component

bartaz:
### DONE: 

- [x] add simple component
- [x] basic input parsing (to get user/repo pair and repo URL)
- [x] talk to GH API to verify repo existence
- [x] move component state to redux store
- [x] connect store to component, remove component state
- [x] tests of the component

- [x] talk to GH API to verify snapcraft.yaml existence (can be one step with checking repo on GH?)

### TODO (postponed to separate branch)
- better styles / reuse form components from MU

### DIFF
--- a/src/common/actions/repository-input.js
+++ b/src/common/actions/repository-input.js
@@ -42,11 +42,8 @@ export function verifyGitHubRepository(repository) {
         payload: gitHubRepo.repo
       });
 
-      return fetch(`https://api.github.com/repos/${repo}`)
+      return fetch(`https://api.github.com/repos/${repo}/contents/snapcraft.yaml`)
         .then(checkStatus)
-        // TODO:
-        //.then(response => response.json())
-        // use json.clone_url from response instead of building our own?
         .then(() => dispatch(verifyGitHubRepositorySuccess(`https://github.com/${repo}.git`)))
         .catch(error => dispatch(verifyGitHubRepositoryError(error)));
     } else {
--- a/src/common/actions/repository-input.js
+++ b/src/common/actions/repository-input.js
@@ -3,9 +3,17 @@ import parseGitHubUrl from 'parse-github-url';
 
 export const CHANGE_REPOSITORY_INPUT = 'CHANGE_REPOSITORY_INPUT';
 export const UPDATE_STATUS_MESSAGE = 'UPDATE_STATUS_MESSAGE';
+export const SET_GITHUB_REPOSITORY = 'SET_GITHUB_REPOSITORY';
 export const VERIFY_GITHUB_REPOSITORY = 'VERIFY_GITHUB_REPOSITORY';
 export const VERIFY_GITHUB_REPOSITORY_SUCCESS = 'VERIFY_GITHUB_REPOSITORY_SUCCESS';
 export const VERIFY_GITHUB_REPOSITORY_ERROR = 'VERIFY_GITHUB_REPOSITORY_ERROR';
+
+export function updateInputValue(value) {
+  return (dispatch) => {
+    dispatch(changeRepositoryInput(value));
+    dispatch(validateGitHubRepository(value));
+  };
+}
 
 export function changeRepositoryInput(value) {
   return {
@@ -14,11 +22,22 @@ export function changeRepositoryInput(value) {
   };
 }
 
-export function updateStatusMessage(message) {
+export function setGitHubRepository(repository) {
   return {
-    type: UPDATE_STATUS_MESSAGE,
-    payload: message
+    type: SET_GITHUB_REPOSITORY,
+    payload: repository
   };
+}
+
+export function validateGitHubRepository(repository) {
+  const gitHubRepo = parseGitHubUrl(repository);
+  const repo = gitHubRepo ? gitHubRepo.repo : null;
+
+  if (repo) {
+    return setGitHubRepository(repo);
+  } else {
+    return verifyGitHubRepositoryError(new Error('Invalid repository URL.'));
+  }
 }
 
 function checkStatus(response) {
@@ -33,22 +52,16 @@ function checkStatus(response) {
 
 export function verifyGitHubRepository(repository) {
   return (dispatch) => {
-    const gitHubRepo = parseGitHubUrl(repository);
-    const repo = gitHubRepo ? gitHubRepo.repo : null;
-
-    if (repo) {
+    if (repository) {
       dispatch({
         type: VERIFY_GITHUB_REPOSITORY,
-        payload: gitHubRepo.repo
+        payload: repository
       });
 
-      return fetch(`https://api.github.com/repos/${repo}/contents/snapcraft.yaml`)
+      return fetch(`https://api.github.com/repos/${repository}/contents/snapcraft.yaml`)
         .then(checkStatus)
-        .then(() => dispatch(verifyGitHubRepositorySuccess(`https://github.com/${repo}.git`)))
+        .then(() => dispatch(verifyGitHubRepositorySuccess(`https://github.com/${repository}.git`)))
         .catch(error => dispatch(verifyGitHubRepositoryError(error)));
-    } else {
-      // TODO: above we return promise, here nothing - inconsistent and harder to test
-      dispatch(verifyGitHubRepositoryError(new Error('Invalid repository URL.')));
     }
   };
 }
--- a/src/common/actions/repository-input.js
+++ b/src/common/actions/repository-input.js
@@ -1,44 +1,14 @@
 import 'isomorphic-fetch';
-import parseGitHubUrl from 'parse-github-url';
 
-export const CHANGE_REPOSITORY_INPUT = 'CHANGE_REPOSITORY_INPUT';
-export const UPDATE_STATUS_MESSAGE = 'UPDATE_STATUS_MESSAGE';
 export const SET_GITHUB_REPOSITORY = 'SET_GITHUB_REPOSITORY';
 export const VERIFY_GITHUB_REPOSITORY = 'VERIFY_GITHUB_REPOSITORY';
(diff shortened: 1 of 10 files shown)


src/common/actions/repository-input.js
```diff
@@ -0,0 +1,69 @@
...
+        type: VERIFY_GITHUB_REPOSITORY,
+        payload: gitHubRepo.repo
+      });
+
+      return fetch(`https://api.github.com/repos/${repo}/contents/snapcraft.yaml`)
+        .then(checkStatus)
+        .then(() => dispatch(verifyGitHubRepositorySuccess(`https://github.com/${repo}.git`)))
+        .catch(error => dispatch(verifyGitHubRepositoryError(error)));
+    } else {
+      // TODO: above we return promise, here nothing - inconsistent and harder to test
```
bartaz: Not sure how to best handle this.

The problem is that `verifyGitHubRepository` does 2 steps of verification.
One is synchronous (parse the input with `parseGitHubUrl` and check if it's valid GH repo URL) and the other is asynchronous (call GH API for snapcraft.yaml).

In async branch we return the fetch promise (so it can be tested). The sync branch just dispatches failure action immediately. This means both branches need different types of tests and the return value of this function is inconsistent.

I'm not sure what would be the best way to solve it? Wrapping sync branch in immediately resolved Promise works, but feels unnecessary and bit dirty. Maybe splitting that into 2 separate actions (one sync for just parsing the input and the other for async calling API)? Any other ideas?

src/common/components/repository-input/index.js
```diff
@@ -0,0 +1,70 @@
...
+  changeRepositoryInput,
+  verifyGitHubRepository
+} from '../../actions/repository-input';
+
+export class RepositoryInput extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  getStatusMessage() {
```
bartaz: I'm bit inconsistent here.

At first I added `statusMessage` in reducer and `updateStatusMessage` action to keep the validation/status message in the store too. But once I got to that it seemed more natural to construct status message in the component based on the state from the store.

Not sure which way to follow? Keep messages in the component (view) and remove it from reducer/action? Or move messages to the store (so the actions themselves will update the messages in the store and component will just display it?

src/common/actions/repository-input.js
earnubs: Maybe redux-batched-actions? https://github.com/reactjs/redux/issues/911#issuecomment-149361073 & https://github.com/tshelburne/redux-batched-actions

earnubs: Aside from the batched actions suggestion (which I've never used) it looks like it should be 2 distinct actions, from the perspective of a reviewer ;)

src/common/actions/repository-input.js
```diff
@@ -0,0 +1,69 @@
+import 'isomorphic-fetch';
+import parseGitHubUrl from 'parse-github-url';
+
+export const CHANGE_REPOSITORY_INPUT = 'CHANGE_REPOSITORY_INPUT';
```
earnubs: ... or SET_GITHUB_REPO_URL ?

src/common/actions/repository-input.js
bartaz: Yes, it seems it will be cleaner and easier to test to have them as separate actions.

src/common/actions/repository-input.js
katiefenn: `CHANGE_REPOSITORY_INPUT` sounds alright to me, given that it's going to be called every time the value is changed.

Suggested alternatives: `REPOSITORY_INPUT_ON_CHANGE`, `UPDATE_REPOSITORY_INPUT`

src/common/actions/repository-input.js
katiefenn: Agree, validity of input should be checked in a different action. This could be initiated by the `changeRepositoryInput` action creator.

src/common/components/repository-input/index.js
katiefenn: I think this is good example to discuss the limits of presentation component logic. I think this amount of switching is okay, but if it became any more complex it would be better to move it into the store.

From a testing perspective, it's nice to pass props like "message" and "status" to presentational components and test whether they are used. But I'm borderline in this instance.

src/common/components/repository-input/index.js
```diff
@@ -0,0 +1,70 @@
...
+    let message;
+
+    if (!input.repository) {
+      message = 'Invalid repository URL.';
+    } else if (input.repository && input.isFetching) {
+      message = `Verifying ${input.repository} on GitHub...`;
+    } else if (input.success && input.repositoryUrl) {
+      message = `Repository ${input.repository} is valid.`;
+    } else if (input.errors) {
+      message = `Repository ${input.repository} is invalid.`;
```
katiefenn: I think we need to distinguish here between an invalid GitHub repository name ("081 811 8181" / "katie.fenn@googlemail.com") and an inaccessible / non-existant GitHub repository ("katiefenn/thisrepodoesntexist").

Suggested message: "Repository either does not exist, or access not authorised."

src/common/components/repository-input/index.js
```diff
@@ -0,0 +1,70 @@
...
+      </div>
+    );
+  }
+
+  onInputChange(event) {
+    this.props.dispatch(changeRepositoryInput(event.target.value));
+  }
+
+  onButtonClick() {
+    this.props.dispatch(verifyGitHubRepository(this.props.repositoryInput.inputValue));
```
katiefenn: Is clicking the button the only way the user can initiate the verification of the GitHub repository? What happens when they enter a URL and hit enter?

src/common/components/repository-input/index.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+RepositoryInput.propTypes = {
+  repositoryInput: PropTypes.object.isRequired,
+  dispatch: PropTypes.func.isRequired
+};
+
+function mapStateToProps(state) {
```
katiefenn: This would normally be way too low to bind components to state. This should really be done in the container. Is this a timesaver?

src/common/components/repository-input/index.js
bartaz: It's a submit button, so I assumed it would work on enter press too. I'll have a look why it's not the case.

src/common/components/repository-input/index.js
bartaz: Yes, messages are not very helpful ATM. Moving sync parsing/validation to input update function separately from calling GH API should help with that.

src/common/components/repository-input/index.js
katiefenn: I think it's because there's no `<form>` element with an action or onSubmit handler.

src/common/components/repository-input/index.js
katiefenn: @bartaz: Maybe "container" wasn't the right word in this case. I think presentational components (such as RepositoryInput) should contain logic only for changing their appearance and defer responding to user input to their parent component. [This article](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.c0azv4409) sums up what I mean ("presentational components have no dependencies on the rest of the app, such as Flux actions or stores").

In this case, I think it would be better to pass functions in as props like we do with the [InputField component in JavanRhino](https://github.com/canonical-ols/javan-rhino/blob/master/src/common/components/forms/inputField.js#L52).

I think there's a value in all our form components behaving in similar ways and having similar interfaces. It would also allow us to have two instances of RepositoryInput on the same page, which is an unlikely situation but could prevent some tricky bugs.

bartaz: @katiefenn at this stage making RepositoryInput just a presentational component would mean more or less that it's just an input with a button and all of the other logic is moved out of it so some other layer that actually does data binding, updates state on keypress etc.

I understand the point of doing it this way, but seems a little overkill on such a small thing this component is ATM.

Maybe it will all make more sense once we take form components from MU and use it here for styling. This will mean that components like InputField, etc will be the dumb presentational components, while RepositoryInput will become the binding between the redux state and props of form components?

katiefenn: > This will mean that components like InputField, etc will be the dumb presentational components, while RepositoryInput will become the binding between the redux state and props of form components?

That sounds like it'd work 👍

bartaz: OK, let's do it in the next branch.

src/common/actions/repository-input.js
earnubs: Perhaps a bikeshed, but the action is not to change the repository input, it's to change the repository url (or 'set', but my argument is against the word 'input', not the word 'change').

earnubs: There's nothing about the action that's linked to inputs, or change events, or anything at a component level, naming the action with reference to inputs or events needlessly couples it to the component here.

katiefenn: @earnubs Ah - I see. `CHANGE_REPOSITORY` sounds fine in that case.

src/common/actions/repository-input.js
```diff
@@ -0,0 +1,84 @@
...
+export function setGitHubRepository(repository) {
+  return {
+    type: SET_GITHUB_REPOSITORY,
+    payload: repository
+  };
+}
+
+export function validateGitHubRepository(repository) {
+  return (dispatch) => {
+    const gitHubRepo = parseGitHubUrl(repository);
```
earnubs: I wonder if this should be moved to the reducer? And that might help my argument on the number of and naming of actions to set the url?

src/common/components/repository-input/index.js
```diff
@@ -17,27 +17,29 @@ export class RepositoryInput extends Component {
     const input = this.props.repositoryInput;
     let message;
 
-    if (input.repository && input.isFetching) {
+    if (input.inputValue.length > 2 && !input.repository) {
```
earnubs: minor ux nit, too early reporting of "invalid" input is annoying, is 2 the smallest this can be?

PR #9, opened 2016-11-24
Add Launchpad API client

cjwatson:
This is based on the JavaScript client in Launchpad itself
(lib/lp/app/javascript/client.js).  Changes from the Launchpad version
are as follows:

 * Drop a number of functions specific to the Launchpad web application;
   this is now a pure API client usable from a Node.js server.
 * Remove HostedFile support, at least for now.
 * Port to ES2015 and remove all use of YUI.
 * Convert to fetch (using isomorphic-fetch) and promises.
 * Require a base URI.  (In pages served from launchpad.net, it makes
   sense to use relative paths; rather less so from a different server.)
 * Convert tests to mocha, expect, and nock, and extend them somewhat.


src/server/launchpad/client.js
```diff
@@ -0,0 +1,347 @@
+/* Copyright 2009-2016 Canonical Ltd.  This software is licensed under the
+ * GNU Affero General Public License version 3 (see the file LICENSE).
+ *
+ * Utility methods and classes to deal with the Launchpad API using
+ * Javascript.
+ */
+
+require('es6-promise/auto');
```
earnubs: why require here, not import?

src/server/launchpad/client.js
```diff
@@ -0,0 +1,347 @@
...
+
+/**
+ * Converts an absolute URI into a relative URI.
+ * Prepends the root to a relative URI that lacks the root.
+ * Does nothing to a relative URI that includes the root.
+ */
+export function normalizeURI(base_uri, uri) {
+  const service_base = '/api/devel';
+  const base_parsed = url.parse(base_uri);
+  let parsed = url.parse(uri);
```
earnubs: maybe const?

src/server/launchpad/client.js
```diff
@@ -0,0 +1,347 @@
...
+
+// The resources that come together to make Launchpad.
+
+/** The base class for objects retrieved from Launchpad's web service. */
+export class Resource {
+  /** Initialize a resource with its representation and URI. */
+  init(client, representation, uri) {
+    this.lp_client = client;
+    this.uri = uri;
+    for (const key of Object.keys(representation)) {
```
earnubs: let, not const, since you rebind?

src/server/launchpad/client.js
```diff
@@ -0,0 +1,347 @@
...
+    this.init(client, representation, uri);
+  }
+}
+
+/** A grouped collection of objects from the Launchpad web service. */
+export class Collection extends Resource {
+  constructor(client, representation, uri) {
+    super();
+    this.init(client, representation, uri);
+    for (let i = 0; i < this.entries.length; i++) {
```
earnubs: for consistency, maybe stick to `for ... of` loops?

src/server/launchpad/client.js
```diff
@@ -0,0 +1,347 @@
...
+  constructor(client, representation, uri) {
+    super();
+    this.lp_client = client;
+    this.uri = uri;
+    this.lp_attributes = {};
+    this.dirty_attributes = [];
+
+    // Copy the representation keys into our own set of attributes, and add
+    // an attribute-change event listener for caching purposes.
+    for (const key of Object.keys(representation)) {
```
earnubs: let, not const (i'm doubting myself, if this hasn't already thrown an error)

src/server/launchpad/client.js
```diff
@@ -0,0 +1,347 @@
...
+    }
+  }
+
+  parsed.protocol = base_parsed.protocol;
+  parsed.host = base_parsed.host;
+
+  return url.format(parsed);
+}
+
+class ResourceError extends Error {
```
bartaz: Nit picking a bit but it seems like it could be split into multiple files (class per file or at least some meaningful chunks).

Nothing necessary for now, but may help later in maintenance and testing.

src/server/launchpad/client.js
```diff
@@ -0,0 +1,347 @@
...
+    this.init(client, representation, uri);
+  }
+}
+
+/** A grouped collection of objects from the Launchpad web service. */
+export class Collection extends Resource {
+  constructor(client, representation, uri) {
+    super();
+    this.init(client, representation, uri);
+    for (const i in this.entries.keys()) {
```
bartaz: Any reason for iterating over `keys` instead of simply doing good'ol `for (const i = 0; i < this.entries.length)` or use `this.entries.forEach()`?

src/server/launchpad/client.js
```diff
@@ -0,0 +1,347 @@
...
+    if (config === undefined) {
+      config = {};
+    }
+    const start = config.start;
+    const size = config.size;
+    const headers = {
+      ...this.makeHeaders(),
+      'Accept': config.accept || 'application/json'
+    };
+    let parameters = { ...config.parameters };
```
bartaz: Just a detail, but `parameters` can be a const here it seems. As long as we don't do any other `parameters = ...` assignment parameters object can be changed with `parameters.something = ...` or `parameters['something'] = ...`.

src/server/launchpad/index.js
```diff
@@ -0,0 +1,14 @@
+/* Copyright 2016 Canonical Ltd.  This software is licensed under the
+ * GNU Affero General Public License version 3 (see the file LICENSE).
+ */
+
+import { conf } from '../helpers/config';
+import { Launchpad } from './client';
+
+export function getLaunchpad() {
+  return new Launchpad(
+    conf.get('LP_API_URL'), conf.get('LP_CONSUMER_KEY'),
```
bartaz: As it looks we need more env variables than just `LP_API_URL` that is added to dev env file we should have them documented at least.

I would suggest file similar to https://github.com/canonical-ols/javan-rhino/blob/master/docs/environment-variables.md (sorry we don't have it created yet - we should!).

test/unit/src/server/launchpad/t_client.js
```diff
@@ -0,0 +1,415 @@
...
+  Collection,
+  Entry,
+  normalizeURI,
+  wrapResourceOnSuccess
+} from '../../../../../src/server/launchpad/client';
+import getLaunchpad from '../../../../../src/server/launchpad';
+
+const LP_API_URL = conf.get('LP_API_URL');
+
+describe('normalizeURI', () => {
```
bartaz: It would be best to have one root `describe` per test file. If needed split into several `context` groups for testing different cases.

It goes in line with comment about splitting the lib into several files.
I think that `describe` groups as seen in this test are good candidates to be in separate source files. Or at least in separate test files.

bartaz: It would be best to have one root `describe` per test file. If needed split into several `context` groups for testing different cases.

It goes in line with comment about splitting the lib into several files.
I think that `describe` groups as seen in this test are good candidates to be in separate source files. Or at least in separate test files.

src/server/launchpad/index.js
cjwatson: Done.  Other people can add theirs.

(The reason I only added LP_API_URL to the dev env was because that's enough for minimal anonymous access; the rest are part of actual authentication credentials and shouldn't be committed.)

src/server/launchpad/client.js
cjwatson: I've split out some reasonably-sized pieces now.

test/unit/src/server/launchpad/t_client.js
cjwatson: Done.

src/server/launchpad/client.js
cjwatson: Done, here and elsewhere.

PR #18, opened 2016-11-30
mirroring commit from mu; don't ship production npm deps

earnubs:
This will delete devDeps from node_modules, reducing churn in staging and production branches and preventing shipping unused code in deployments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,7 @@ $(DIST):
 	mkdir -p $(DISTDIR)
 	npm install || (cat npm-debug.log && exit 1)
 	npm run build
+	npm install --production || (cat npm-debug.log && exit 1)
 	touch $@
 
 $(PAYLOAD): $(CHARM) $(DIST) version-info build-exclude.txt $(SRC) $(SRC)/* $(SRC_PREQS)


Makefile
```diff
@@ -62,6 +62,7 @@ $(DIST):
 	mkdir -p $(DISTDIR)
 	npm install || (cat npm-debug.log && exit 1)
 	npm run build
+	npm install --production || (cat npm-debug.log && exit 1)
```
bartaz: I'm not sure why, but it seems it doesn't work here - meaning it doesn't remove dev dependencies.

I checked it out in MU and it seems to work there (calling `npm install` followed by `npm install --production` removes unnecessary dependencies). But in here it seems to do nothing.

bartaz:
Because of https://github.com/canonical-ols/javan-rhino/issues/217 this won't work.

We need to move some needed devDependencies to prod dependencies or make sure they are required only in dev environment.

Makefile
bartaz: OK, I think I found why it's not working. We lack npm-shrinkwrap.json.

It seems that only with npm-shrinkwrap.json being present `npm install` followed by `npm install --production` actually removes dev dependencies.

bartaz:
We should first separate dev and prod servers and fix the dependencies as in https://github.com/canonical-ols/javan-rhino/pull/226

PR #23, opened 2016-12-02
13: Webhook endpoints (part 2)

katiefenn:
- Added new incoming webhook endpoint
- Added "ping" request to github mock service
- Added tests for webhook endpoint


environments/development.env
```diff
@@ -4,7 +4,7 @@ BASE_URL=http://localhost:8000
 WEBPACK_DEV_URL=http://localhost:8001
 NODE_ENV=development
 LP_API_URL=https://api.staging.launchpad.net
-WEBHOOK_ENDPOINT=https://build.staging.snapcraft.io/webhook/notify
+WEBHOOK_ENDPOINT=http://localhost:8000/api/webhook/notify
```
katiefenn: Default to the mock service.

src/server/handlers/webhook.js
```diff
@@ -0,0 +1,9 @@
+const logging = require('../logging/').default;
+const logger = logging.getLogger('express');
+
+export const notify = (req, res) => {
+  // Acknowledge webhook
+  res.status(200).send();
```
katiefenn: Acknowledge, but do nothing.

Handling push events is outside of the scope of this PR; need to find out from @cjwatson how the payload should be handled.

bartaz: We could maybe log something here at least until we have some functionality?

earnubs: Even with functionality, this seems like it might be good for `logger.debug`?

cjwatson: Yeah, just logging it for now is fine, and I agree we ought to retain logging even once full integration is in place.  The webhook will basically need to do `lp.snaps.findByURL` on the repository URL (with memcache) and then `snap.requestAutoBuilds` on that, but I'm still putting together the pieces of the LP interaction here.

mocks/service/github/index.js
```diff
@@ -14,6 +16,21 @@ responses['okay-new-hook-created'] = (req, res) => {
...
   // No response body implemented
   res.send();
+
+  // Send ping to webhook endpoint
+  request.post({
+    url: req.body.config.url,
+    headers: {
+      'X-Github-Delivery': '72d3162e-cc78-11e3-81ab-4c9367dc0958',
+      'User-Agent': 'GitHub-Hookshot/044aadd',
+      'X-GitHub-Event': 'issues'
```
cjwatson: [push](https://developer.github.com/v3/activity/events/types/#pushevent) would be a more realistic choice of event.

environments/development.env
```diff
@@ -6,7 +6,7 @@ NODE_ENV=development
 UBUNTU_SSO_URL=https://login.ubuntu.com
 OPENID_VERIFY_URL=http://localhost:8000/login/verify
 LP_API_URL=https://api.launchpad.net
-WEBHOOK_ENDPOINT=https://build.staging.snapcraft.io/webhook/notify
+WEBHOOK_ENDPOINT=http://localhost:8000/api/webhook/notify
```
cjwatson: Could you add a suitable `WEBHOOK_ENDPOINT` to the staging and production environments as well?  Looks like they should be `https://build.staging.snapcraft.io/api/webhook/notify` and `https://build.snapcraft.io/api/webhook/notify` respectively.

PR #26, opened 2016-12-07
Builds view

bartaz:
screen shot 2016-12-12 at 17 37 26
screen shot 2016-12-12 at 17 37 42

### TODO

- [x] get mocked data in format compatible with LP (messages, dates, user, etc)
- [x] test snap builds store reducers and actions
- [x] builds listing loading (waiting for builds data to fetch)
- [x] open builds view from GH repo component
- [x] builds listing view/route
- [x] mocked builds listing data
- [x] basic builds listing styling
- [x] single build view/route
- [x] display build log in build view
- [x] remove commit info from view (as it may not be available from API yet)

#### Review TODO

- [x] copy/paste in tests names x2
- [x] drop dateFirstDispatched
- [x] `CANCELING` status should be error rather then pending
- [x] dateCompleted || dateStarted confusion
- [x] arch_tag TODO comment
- [x] unnecessary TODO about loading details
- [x] ~~fetch in connect? http://jamesknelson.com/join-the-dark-side-of-the-flux-responding-to-actions-with-actors/~~
- [x] unnecessary BuildStatusLP (?)
- [x] pending/success/error as class names in views (should use enum?)


src/common/components/build-log/index.js
```diff
@@ -0,0 +1,18 @@
+import React, { PropTypes } from 'react';
+
+import styles from './build-log.css';
+
+const BuildLog= (props) => {
+
+  return (
+    <iframe className={styles.frame} src={props.logUrl}>
```
bartaz: Most basic component for displaying build log as an iframe.

It's obviously very limited, but the quickest thing we can have ATM.

src/common/components/repository-input/index.js
```diff
@@ -33,6 +34,14 @@ export class RepositoryInput extends Component {
     return message;
   }
 
+  componentWillReceiveProps(nextProps) {
```
bartaz: I'm not sure if it's the best way to do it, but it was simplest solution I could find.

So currently it's the repository input component that redirects to `/user/repo/builds` once the repository entered is verified.

Maybe there should be a separate redux action to do that? But who should dispatch it?

src/common/containers/build-details.js
```diff
@@ -0,0 +1,74 @@
...
+  render() {
+    const { account, repo, fullName, buildId, build } = this.props;
+
+    return (
+      <div className={ styles.container }>
+        <Helmet
+          title={`${fullName} builds`}
+        />
+        {/* TODO: make into title component? */}
+        <h1>{fullName} build #{buildId}</h1>
```
bartaz: It may be worth making it a separate component (it's used in two views already), but I wasn't sure if making separate module just to render h1 is worth it ATM. Maybe if we have more complex layout or specific styles for it...

src/common/containers/build-details.js
```diff
@@ -0,0 +1,74 @@
...
+const mapStateToProps = (state, ownProps) => {
+  const account = ownProps.params.account.toLowerCase();
+  const repo = ownProps.params.repo.toLowerCase();
+  const buildId = ownProps.params.buildId.toLowerCase();
+
+  const fullName = `${account}/${repo}`;
+  // TODO:
+  // find build by id in builds list
+  // but it should also fetch it if necessary
+  const build = state.snapBuilds.builds.filter((build) => build.buildId === buildId)[0];
```
bartaz: Very basic 'find by id' implementation to find specific build details in list of the builds.

Currently it works on assumption that we have a store with list of builds of given repository. It gets bit tricky when user goes directly to details page (when listing was not yet fetched). Now, with mocked data, I just 'fetch' whole mocked list of builds, but in the end we should probably only request single builds details and fetch listing only when user goes to listing page.

src/common/helpers/snap-builds.js
```diff
@@ -0,0 +1,91 @@
...
+  [BuildStatusLP.UPLOADING]: BuildStatus.PENDING,
+  [BuildStatusLP.CANCELLING]: BuildStatus.PENDING,
+  [BuildStatusLP.CANCELLED]: BuildStatus.ERROR
+};
+
+function getLastPartOfUrl(url) {
+  return url ? url.substr(url.lastIndexOf('/') + 1) : null;
+}
+
+export function snapBuildFromAPI(entry) {
```
bartaz: Transforming snap build details as returned from LP API into values we use in our views (parsing id and arch from URLs, mapping build status, coping dates).

src/common/reducers/snap-builds.js
```diff
@@ -0,0 +1,31 @@
...
+    case ActionTypes.FETCH_BUILDS:
+      return {
+        ...state,
+        isFetching: true
+      };
+    case ActionTypes.FETCH_BUILDS_SUCCESS:
+      return {
+        ...state,
+        isFetching: false,
+        builds: action.payload.map(snapBuildFromAPI),
```
bartaz: Most of the 'logic' of transforming of API returned entry into values in the store is done in `snapBuildFromAPI` helper.

package.json
```diff
@@ -42,13 +42,14 @@
...
     "memcached": "^2.2.2",
+    "moment": "^2.17.1",
     "nconf": "^0.8.4",
     "nconf-yaml": "^1.0.2",
     "normalize.css": "^5.0.0",
     "oauth": "^0.9.14",
     "openid": "^2.0.6",
-    "qs": "^6.3.0",
     "parse-github-url": "^0.3.2",
+    "qs": "^6.3.0",
```
earnubs: why npm, why?

src/common/components/repository-input/index.js
earnubs: initial suggestion here http://jamesknelson.com/join-the-dark-side-of-the-flux-responding-to-actions-with-actors/ is to do it in connect? then goes on to talk about actors ...

src/common/containers/build-details.js
earnubs: -1 to new component, for the reasons you give :)

src/common/helpers/snap-builds.js
```diff
@@ -0,0 +1,91 @@
...
+  PENDING: 'pending', // for build currently running or in any way in progress
+  ERROR: 'error'      // for builds failed for any reason
+};
+
+// TODO:
+// export enum with build status success/pending/error
+// or export full list of statuses (like in API) and let view do the mapping?
+
+// mapping between build status from LP and pending/success/error internal status
+const BuildStatusMapping = {
```
cjwatson: Can we avoid duplicating information between comments in BuildStatusLP and values in BuildStatusMapping?  I'd be inclined to drop BuildStatusLP and just map directly from string statuses to BuildStatus elements.

Also isn't the `[]` around each key redundant?

src/common/helpers/snap-builds.js
```diff
@@ -0,0 +1,91 @@
...
+  [BuildStatusLP.NEEDSBUILD]: BuildStatus.PENDING,
+  [BuildStatusLP.FULLYBUILT]: BuildStatus.SUCCESS,
+  [BuildStatusLP.FAILEDTOBUILD]: BuildStatus.ERROR,
+  [BuildStatusLP.MANUALDEPWAIT]: BuildStatus.PENDING,
```
cjwatson: This is debatable, since it might make progress on its own at some point in the future; but on balance I'd be inclined to render this as an error state.

src/common/helpers/snap-builds.js
```diff
@@ -0,0 +1,91 @@
...
+  [BuildStatusLP.CHROOTWAIT]: BuildStatus.ERROR,
+  [BuildStatusLP.SUPERSEDED]: BuildStatus.ERROR,
+  [BuildStatusLP.BUILDING]: BuildStatus.PENDING,
+  [BuildStatusLP.FAILEDTOUPLOAD]: BuildStatus.ERROR,
+  [BuildStatusLP.UPLOADING]: BuildStatus.PENDING,
+  [BuildStatusLP.CANCELLING]: BuildStatus.PENDING,
```
cjwatson: Should be error.  Launchpad may not have finished actually killing off the build at this point, but it has no path to success from here.

src/common/helpers/snap-builds.js
```diff
@@ -0,0 +1,91 @@
...
+    buildLogUrl: entry.build_log_url,
+
+    architecture: getLastPartOfUrl(entry.distro_arch_series_link),
+
+    status: BuildStatusMapping[entry.buildstate],
+    statusMessage: entry.buildstate,
+
+    dateCreated: entry.datecreated,
+    dateStarted: entry.date_started,
+    dateFirstDispatched: entry.date_first_dispatched,
```
cjwatson: You don't use `dateFirstDispatched`, and I think every possible use of it on this site should probably use `dateStarted` instead, so I'd drop this.

I think `dateCreated` is currently unused too, but I can imagine reasonable uses of that here.

test/unit/src/common/actions/t_snap-builds.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  context('fetchBuildsError', () => {
+    let payload = 'Something went wrong!';
+
+    beforeEach(() => {
+      action = fetchBuildsError(payload);
+    });
+
+    it('should create an action to store github repo error on failure', () => {
```
cjwatson: s/github repo/Launchpad/?

test/unit/src/common/actions/t_snap-builds.js
```diff
@@ -0,0 +1,98 @@
...
+      return store.dispatch(fetchBuilds('foo/bar'))
+        .then(() => {
+          expect(store.getActions()).toHaveActionOfType(
+            ActionTypes.FETCH_BUILDS_SUCCESS
+          );
+        });
+    });
+
+    // TODO: pending - mocked actions never fail
+    xit('should store error on GitHub verification failure', () => {
```
cjwatson: s/GitHub/Launchpad/?

test/unit/src/common/actions/t_snap-builds.js
bartaz: Good catch. Copy/paste ;)

src/common/helpers/snap-builds.js
bartaz: Yes, I've put them all here to make them available for UI if we want to show them.

But I'll take you suggestion and drop 'first dispatched', while keeping date created for possible future use.

@cjwatson I'm not sure if `datecreated` vs `date_started` case mismatch was intended (backwards compatibility or something?) but it's quite confusing to have different props using different naming format.

src/common/helpers/snap-builds.js
bartaz: My idea behind having `BuildStatusLP` was to have possibility of using `BuildStatusLP.NEEDSBUILD` instead of hardcoding `"Needs building"` string around the code. So the `BuildStatusLP` would be equivalent of `BuildStatus` enum from Python code to be used around the code with the string values only in one place.

But you are right, that at this point we don't use it anywhere else and I can't think of the case where we may need it if we map it to internal `PENDING/SUCCESS/ERROR` status here, so I can simplify the mapping and drop it.

@cjwatson the `[]` is needed to compute the keys: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names (it's a new JS syntax).

package.json
bartaz: I wouldn't blame npm on that. I guess `qs` was added manually instead of using `npm --save` so it slipped a bit in sorting.

src/common/components/build-row/index.js
```diff
@@ -0,0 +1,55 @@
...
+    <div className={ `${styles.buildRow} ${styles[status]}` }>
+      <div className={ styles.item }><Link to={`/${account}/${repo}/builds/${buildId}`}>{`#${buildId}`}</Link> {statusMessage}</div>
+      <div className={ styles.item }>
+        {architecture}
+      </div>
+      <div className={ styles.item }>
+        {moment.duration(duration).humanize()}
+      </div>
+      <div className={ styles.item }>
+        <span title={moment(dateCompleted || dateStarted).format('DD-MM-YYYY HH:mm:ss')}>{ moment(dateCompleted || dateStarted).fromNow() }</span>
```
bartaz: Agreed. We could show both (only started for those in progress) or add some label to make it clear.

I'll look into what is quickest and looks good. We would hopefully fix that better once we have designs.

src/common/helpers/snap-builds.js
cjwatson: Ah, of course, I even knew about that syntax for my recent async iterator work and forgot about it.

src/common/helpers/snap-builds.js
cjwatson: Yes, it is somewhat unfortunate.  We're pretty much stuck with it now for compatibility though.

src/common/helpers/snap-builds.js
bartaz: @cjwatson So it's fine to leave it as `PENDING`? I guess 'PENDING' may not be a perfect name here as in many cases it simply means 'IN_PROGRESS'.

src/common/components/repository-input/index.js
katiefenn: There is an intermediate step I am working on that will call on a Redux action: webhook creation. As soon as the webhook is successfully installed, it can redirect to the repo builds page.

src/common/containers/builds.js
```diff
@@ -0,0 +1,59 @@
...
+
+  render() {
+    const { account, repo, fullName } = this.props;
+
+    return (
+      <div className={ styles.container }>
+        <Helmet
+          title={`${fullName} builds`}
+        />
+        {/* TODO: make into title component? */}
```
katiefenn: Remove TODO

src/common/helpers/snap-builds.js
```diff
@@ -0,0 +1,91 @@
...
+  CHROOTWAIT: 'Chroot problem',     // error
+  SUPERSEDED: 'Build for superseded Source', // error
+  BUILDING: 'Currently building',   // pending
+  FAILEDTOUPLOAD: 'Failed to upload', // error
+  UPLOADING: 'Uploading build',     // pending
+  CANCELLING: 'Cancelling build',   // pending
+  CANCELLED: 'Cancelled build'      // error
+};
+
+export const BuildStatus = {
```
katiefenn: I think a helper is the right place for this feature. The build status is the fundamental state of the application - the underlying LaunchPad build status is less prominent (if it is even shown to the user at all).

Keeping it separate from the UI also makes it more testable.

test/unit/src/common/helpers/t_snap-builds.js
```diff
@@ -0,0 +1,211 @@
...
+    it('should map UPLOADING into PENDING', () => {
+      const entry = {
+        ...SNAP_BUILD_ENTRY,
+        buildstate: BuildStatusLP.UPLOADING
+      };
+
+      expect(snapBuildFromAPI(entry).status).toEqual(BuildStatus.PENDING);
+    });
+
+    it('should map CANCELLING into PENDING', () => {
```
katiefenn: Expectation is testing for "ERROR", not "PENDING"

bartaz: Good catch!

src/common/components/build-row/index.js
bartaz: I removed `dateCompleted`. I guess showing one common started date will be less confusing.
Later if we want we may show both dates in build details, but I guess date started and duration give enough information.

src/common/components/repository-input/index.js
bartaz: I'll leave it as it is for now. I guess it's not worth putting to much work into that before we make all pieces work together as a flow.

src/common/components/build-row/index.js
```diff
@@ -0,0 +1,59 @@
...
+  account: PropTypes.string,
+  repo: PropTypes.string,
+
+  // build properties
+  buildId:  PropTypes.string,
+  architecture: PropTypes.string,
+  status:  PropTypes.string,
+  statusMessage: PropTypes.string,
+  dateStarted: PropTypes.string,
+  dateCompleted: PropTypes.string,
```
cjwatson: This line looks like a leftover.

src/common/helpers/snap-builds.js
```diff
@@ -0,0 +1,71 @@
...
+//   architecture: 'i386', //'arch_tag'
+//
+//   // https://git.launchpad.net/launchpad/tree/lib/lp/buildmaster/enums.py#n22
+//   status:  'error', // parsed based on 'buildstate' -> success, error, pending
+//   statusMessage: 'Failed to build', // 'buildstate'
+//
+//   dateCreated: '2016-12-01T17:08:36.317805+00:00', // 'datecreated'
+//   dateStarted: '2016-12-01T17:08:36.317805+00:00', // 'date_started'
+//   dateFirstDispatched: '2016-12-01T17:08:36.317805+00:00', // 'date_first_dispatched'
+//   dateCompleted: '2016-12-01T17:10:36.317805+00:00', // 'datebuilt'
```
cjwatson: Not quite in sync with the code below.

test/unit/src/common/reducers/t_snap-builds.js
```diff
@@ -0,0 +1,153 @@
...
+import { snapBuildFromAPI } from '../../../../../src/common/helpers/snap-builds';
+
+describe('snapBuilds reducers', () => {
+  const initialState = {
+    isFetching: false,
+    builds: [],
+    error: false
+  };
+
+  const SNAP_ENTRIES = [{
```
cjwatson: These mock entries should possibly have `arch_tag` as well.

bartaz: Good catch. It seems the tests are not perfect if they didn't fail on that...

PR #28, opened 2016-12-08
17: GitHub auth

katiefenn:
- Added endpoint for redirecting users to GitHub auth login page
- Added endpoint for verifying returning users from GitHub login
- Added route tests for new endpoint
- Added mock login form and actions
- Removed GitHub username and password config items
- Added session helper (mostly copied from Mu)


docs/environment-variables.md
```diff
@@ -63,25 +63,38 @@ The OAuth token secret, generated by Launchpad.
 ## GitHub settings
 
 ### GITHUB_API_ENDPOINT
-- **Example**: `http://localhost:4000/github`, `https://api.github.com`
+- **Example**: `http://localhost:4000/github`
+- **Example**: `https://api.github.com`
```
katiefenn: No big change here, just re-formatting examples over two lines to match changes later in the file.

docs/environment-variables.md
```diff
@@ -63,25 +63,38 @@ The OAuth token secret, generated by Launchpad.
...
 
 The base URL for the GitHub API.
 
-### GITHUB_USERNAME
-- **Example:** katiefenn
-- **DEPRECATED**
```
katiefenn: `GITHUB_USERNAME` and `GITHUB_PASSWORD` finally deprecated in favour of Oauth tokens 💃

docs/environment-variables.md
```diff
@@ -63,25 +63,38 @@ The OAuth token secret, generated by Launchpad.
...
+-- **Example**: `http://localhost:8000/login/verify`
+-- **Example**: `https://build.snapcraft.io/login/verify`
+
+The local page users should be redirected to after logging in with GitHub.
+
+### GITHUB_AUTH_CLIENT_ID
+-- **Example:** `389a7b4c2ade662c0bf4`
+
+The GitHub application id. Created using the form at GitHub > Your profile > Oauth Applications > Register a new application. Shared credentials for dev, staging and production sites are stored on LastPass. Access to credentials can be obtained by asking a developer.
+
```
katiefenn: TODO: Request staging and production client id and client secret.

Open question: Which GitHub account should these keys be associated with? canonical-ols?

mocks/service/github/login-form.js
```diff
@@ -0,0 +1,18 @@
...
+
+      <input type='submit' value='Log In' style={ { fontSize: 16 } } />
+    </form>
+  );
+}
+
+LoginForm.propTypes = {
+  redirectUrl: PropTypes.string,
+  sharedSecret: PropTypes.string
+};
```
katiefenn: A simple, server-side only login form to stand in for GitHub.

src/server/handlers/github-auth.js
```diff
@@ -0,0 +1,59 @@
+import request from 'request';
+import logging from '../logging';
+
+const AUTH_SCOPE = 'admin:repo_hook';
```
katiefenn: This level of scope works for adding webhooks, we'll probably need other scopes for accessing private repos.

test/routes/src/server/routes/t_github-auth.js
```diff
@@ -0,0 +1,166 @@
...
+        sharedSecret: 'bar'
+      };
+      next();
+    });
+    app.use(login);
+    app.use((err, req, res, next) => {
+      res.status(500).send();
+      next();
+    });
+  });
```
katiefenn: Lots of mocks needed :-(

test/routes/src/server/routes/t_github-auth.js
```diff
@@ -0,0 +1,166 @@
...
+        .expect('location', new RegExp(GITHUB_AUTH_REDIRECT_URL))
+        .end(done);
+    });
+
+    it('should supply a shared secret', (done) => {
+      supertest(app)
+        .get('/auth/authenticate')
+        .expect('location', /state/)
+        .end(done);
+    });
```
katiefenn: Given (a lot of) more time, I would have liked to add some statistical tests to measure randomness of the shared secret. Checking for its presence will have to do for now.

docs/environment-variables.md
cjwatson: Let's not share accounts; we should have a specific account just for this service.  (Possibly one for staging and one for production, though that may be overkill since they're both devops.)

cjwatson: ... but since it is devops, we can take care of setting up the account ourselves; no need to get IS to do it.

src/server/handlers/github-auth.js
```diff
@@ -0,0 +1,59 @@
...
+const logger = logging.getLogger('login');
+
+export const authenticate = (conf) => {
+  return (req, res) => {
+    // Generate unguessable shared secret
+    require('crypto').randomBytes(48, function(err, buffer) {
```
earnubs: do you need to handle the error?

src/server/handlers/github-auth.js
```diff
@@ -0,0 +1,59 @@
...
+export const verify = (conf) => {
+  return (req, res, next) => {
+    // Check shared secret matches our copy
+    if (req.query.state !== req.session.sharedSecret) {
+      // Warn about potential CSRF attack
+      logger.info('Authentication shared secret mismatch');
+      return next(new Error('Returned shared secret does not match generated shared secret'));
+    }
+
+    let options = {
```
earnubs: const?

src/server/handlers/github-auth.js
```diff
@@ -0,0 +1,59 @@
...
+      url: conf.get('GITHUB_AUTH_VERIFY_URL'),
+      json: {
+        client_id: conf.get('GITHUB_AUTH_CLIENT_ID'),
```
earnubs: could we do the conf.get()'s once at the top of the module?

src/server/handlers/github-auth.js
```diff
@@ -0,0 +1,59 @@
...
+    // Exchange code for token
+    request.post(options, (error, response, body) => {
+      if (error || response.statusCode !== 200) {
+        logger.info('Authentication token exchange failed', error);
+        return next(new Error('Authentication token exchange failed'));
+      }
+
+      // Save auth token to session
+      req.session.token = body.access_token;
+      logger.info('User successfully authenticated');
```
earnubs: maybe this should be debug rather than info, it's seems like it would just be noise in the long term.

src/server/handlers/github.js
```diff
@@ -76,16 +75,13 @@ export const newIntegration = (req, res) => {
   });
 };
 
-const getRequest = (account, repo) => {
+const getRequest = (account, repo, token) => {
   return {
     url: GITHUB_API_ENDPOINT + `/repos/${account}/${repo}/hooks`,
```
earnubs: why string concat and template literal? couldn't this just be `${GITHUB_API_ENDPOINT}/repos/...`

test/routes/src/server/routes/t_github-auth.js
earnubs: yeah, this *might* be better in acceptance tests, we'll know better when i've completed the mu tests

test/routes/src/server/routes/t_github-auth.js
```diff
@@ -0,0 +1,166 @@
...
+          .get('/auth/verify')
+          .query({ code: 'foo', state: 'bar' })
+          .send()
+          .expect(500, done);
+      });
+    });
+  });
+});
+
+function getHost(url) {
```
earnubs: url in stdlib https://nodejs.org/api/url.html

docs/environment-variables.md
katiefenn: @cjwatson: 👍

src/server/handlers/github-auth.js
katiefenn: I should do, yes 👍

src/server/handlers/github-auth.js
katiefenn: @earnubs: @bartaz: Anyone in favour of adding a "warn" level of logging for unrecoverable problems such as token exchange failure?

docs/environment-variables.md
earnubs: Interesting, what assurance does the user have that a Github account is "us"?

src/server/handlers/github-auth.js
bartaz: Not sure if private repos are in scope. Even if we would get access to them here, still LP would need to be authorised separately to access them to build.

src/server/handlers/github-auth.js
earnubs: Talisker docs seemed to suggest we only use info and debug, which is how the logger is configured: http://talisker.readthedocs.io/en/latest/logging.html

test/routes/src/server/routes/t_github-auth.js
bartaz: Without digging into complex tests a simpler thing to check would be if two secrets are not equal and that they are long enough.

But not sure if such tests are needed here ATM.

src/server/handlers/github-auth.js
katiefenn: 👍

docs/environment-variables.md
bartaz: @earnubs Yep, some dummy empty account made just for auth may look bit suspicious.

Using @canonical-ols at least lets them verify that it's where the repository of build.snapcraft.io is, or maybe we should use more 'official' account such as @snapcore where all snapcraft related repos are, or @ubuntu or @canonicalltd?

Ubuntu/Canonical presence is scattered all around different accounts/orgs on GH...

test/routes/src/server/routes/t_github-auth.js
katiefenn: @bartaz: I don't think so. We could do basic checks such as the ones you suggested, but I think they can't actually check what actually needs checking - entropy.

bartaz: Yep, question is how deeply we actually need to test it. I guess at this point we are good as it is. Especially that we are not programming random generator just relying on what `crypto` provides us with.

PR #29, opened 2016-12-09
Make way for non-session users of memcached

cjwatson:
We're likely to want to use memcached for other things, such as caching
the GitHub repository URL -> Launchpad snap URL lookup.  Rename some
environment variables to be less session-specific, and add a "session:"
prefix to session-related keys.

### DIFF
--- a/src/server/helpers/session.js
+++ b/src/server/helpers/session.js
@@ -33,11 +33,12 @@ export default function sessionStorageConfig(config) {
     settings.secret = 'dont-use-me-in-production';
   }
 
-  if(config.get('SESSION_MEMCACHED_HOST') && config.get('SESSION_MEMCACHED_SECRET')) {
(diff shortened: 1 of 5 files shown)


src/server/helpers/session.js
```diff
@@ -33,11 +33,12 @@ export default function sessionStorageConfig(config) {
     settings.secret = 'dont-use-me-in-production';
   }
 
-  if(config.get('SESSION_MEMCACHED_HOST') && config.get('SESSION_MEMCACHED_SECRET')) {
+  if(config.get('MEMCACHED_HOST') && config.get('MEMCACHED_SESSION_SECRET')) {
     logger.info('Starting with memcached session store');
```
bartaz: Just for the consistency we should remove 'session' from logger message too.

cjwatson: Done.

PR #30, opened 2016-12-09
Switch to production LP/SSO for all environments

cjwatson:
Launchpad's staging instance doesn't have an attached Git hosting
service, and its qastaging instance doesn't have an attached build farm,
so production is really the only instance we can make any sensible use
of here.  We'll deal with environment separation by using different
Launchpad usernames.

This implies using production SSO too, as that will eventually need to
match what's used by the Launchpad instance we're using.


cjwatson:
Yeah, if there were an alternative I'd take it (I'm actually more concerned about it pushing to the production store than about it using production LP, though at least it only pushes to edge), but there isn't one right now.  At some point in the new year hopefully we'll find time to attach a git service to staging.launchpad.net, at which point we can switch back to that.

PR #31, opened 2016-12-13
Add /launchpad/snaps routes

cjwatson:
POST /launchpad/snaps creates a Snap object on Launchpad with the given
URL; GET /launchpad/snaps returns a Snap object on Launchpad owned by
this service's user, if any.

There's no front-end code for this yet, and it isn't yet possible to choose which architectures to enable; but this should be a good start and we can refine it later.


src/server/routes/launchpad.js
```diff
@@ -0,0 +1,12 @@
+import { Router } from 'express';
+import { json } from 'body-parser';
+
+import { newSnap, findSnap } from '../handlers/launchpad';
+
+const router = Router();
+
+router.use('/launchpad/snaps', json());
+router.post('/launchpad/snaps', newSnap);
+router.get('/launchpad/snaps', findSnap);
```
bartaz: @cjwatson Do we have any tests/examples/docs about how to use these routes?

Digging through the code it's possible but quite hard to find what kind of parameters and in what format they expect, and what they return/

src/server/handlers/launchpad.js
```diff
@@ -0,0 +1,236 @@
...
+  if (parsed === null || parsed.owner === null || parsed.name === null) {
+    logger.info(`Cannot parse "${repositoryUrl}"`);
+    return res.status(400).send(RESPONSE_GITHUB_BAD_URL);
+  }
+
+  const uri = `/repos/${parsed.owner}/${parsed.name}/contents/snapcraft.yaml`;
+  const options = {
+    headers: {
+      'Authorization': `token ${req.session.token}`,
+      'Accept': 'application/vnd.github.v3.raw'
```
bartaz: By default GH API returns JSON response - this header is changing that so raw contents of snapcraft.yaml will be returned?

src/server/handlers/launchpad.js
```diff
@@ -0,0 +1,236 @@
...
+      return res.status(400).send(RESPONSE_SNAPCRAFT_YAML_PARSE_FAILED);
+    }
+    return callback(snapcraftYaml);
+  });
+};
+
+// XXX cjwatson 2016-12-13: We should set an appropriate set of
+// architectures.
+export const newSnap = (req, res) => {
+  getSnapcraftYaml(req, res, snapcraftYaml => {
```
bartaz: It's not a huge deal (especially now in first phase), but we have a bit of duplication. We already call GH API for snapcraft.yaml when verifying user input in https://github.com/canonical-ols/build.snapcraft.io/blob/master/src/common/actions/repository-input.js#L33

Up there we don't parse the response (we just check if the file exists), so for sure we need this raw response to parse it here, but ideally we should only ask the API once.

@cjwatson @katiefenn We are getting to the point where we have some 'loose' pieces without having a bigger picture how all of them would work together in the end.
We have GH and SSO authentication, GH repo input and verification, getting data from LP and creating new snap/build, then listing builds...

It would be worth having some kind of flow chart of the whole user path through the app. It would clarify both user path and all the internal and external (to LP, GH) API calls we need on the way.

src/server/handlers/launchpad.js
```diff
@@ -0,0 +1,236 @@
...
+        store_series: `/+snappy-series/${STORE_SERIES}`,
+        store_name: snapcraftYaml.name
+      }
+    }).then(result => {
+      logger.info(`Authorizing ${result.self_link}`);
+      lp_client.named_post(result.self_link, 'beginAuthorization', {
+        parameters: { success_url: req.body.success_url }
+      }).then(authUrl => {
+        logger.info(`Authorized ${result.self_link}`);
+        return res.redirect(authUrl);
```
bartaz: Got to this point when trying to put this into flow chart.

As far as I understand we first call for `/+snaps` `new`, and once that returns we do the SSO authorisation (based on `self_link` returned from previous call).
And the redirect url (after successful auth) is what we pass as `success_url` from request params?
This is supposed to be redirect back to BSI? Would it be directly one of our views (like /user/repo/builds?) or does auth need to redirect to some our internal endpoint to (like OAuth /verify to store the token)?

src/server/routes/launchpad.js
cjwatson: I've added `docs/routes.md`.

src/server/handlers/launchpad.js
cjwatson: Yes.  We could get it from the default JSON response, but it's more decoding for no gain.

src/server/handlers/launchpad.js
cjwatson: `/+snaps?ws.op=new` must send the user through to SSO to authorise the upload of snap builds to the store on their behalf.  This process is necessarily initiated by Launchpad (because it needs to get a macaroon from SCA first, behind the scenes), but we ultimately want to end up back in BSI, hence the new `snap.beginAuthorization` endpoint.  Yes, `success_url` would therefore be something in BSI.

We don't need to store the token on our side; that's done in Launchpad before redirecting to `success_url`.

cjwatson:
I realised a fatal flaw in my earlier plan, and have switched over to a slightly different authorization arrangement that uses two Launchpad API calls and requires this site to do the OpenID exchange itself (not implemented in this PR, but we have almost all the necessary code in the tree already).  This relies on [this Launchpad MP](https://code.launchpad.net/~cjwatson/launchpad/snap-api-complete-authorization/+merge/313396).

PR #33, opened 2016-12-18
Fix and enable login route tests

cjwatson:
The login route tests need to be in a file matching ./test/routes/**/*
in order to actually be run.  Getting them to work properly requires a
few gymnastics in order to simulate a successful OpenID exchange, and I
made them a little more specific as well.


test/routes/src/server/routes/t_login.js
```diff
@@ -0,0 +1,105 @@
...
+  githubAuth: () => {},
+  universal: () => {},
+  // tell proxyquire not to try to load stubbed module
+  '@noCallThru': true
+};
+const stubDependencies = {
+  './routes/': routesStub
+};
+
+const app = proxyquire(
```
bartaz: It was me who used proxyquire + routes stub pattern in [t_login.js](https://github.com/canonical-ols/javan-rhino/blob/master/test/routes/t_login.js) but I've seen @katiefenn testing routes in simpler and more isolated way, where we just imports the route to test and create an express app with that route only.

Like:
https://github.com/canonical-ols/build.snapcraft.io/blob/master/test/routes/src/server/routes/t_github.js#L9-L15

It's not that the way with stubbed server here is wrong, but because it starts the whole app server it may 'test' more than it should.

cjwatson: Yeah, that makes sense and it's definitely not ideal to have to explicitly list all the routes we're not testing.  I've converted these tests to the isolated style.

PR #41, opened 2016-12-20
Authentication and build UI

katiefenn:
- [x] Add test for Login with GitHub button
- [x] Add test for progress icons
- [x] Replace emoji progress icons with something more appropriate
- [x] Add test for webhook reducer
- [x] Add test for webhook actions
- [x] Fix linting errors
- [x] Rebase changes from master
- [x] Merge PR #23

PR #42, opened 2016-12-20
Hook up snap creation in the frontend

cjwatson:
If LP_API_USERNAME is set, then we now create a snap based on the given
GitHub repository rather than merely verifying it.  This uses the
/launchpad/snaps API to create the snap in Launchpad, then redirects
through OpenID to authorise uploads of it to the store, then redirects
to the builds page for that repository.

If LP_API_USERNAME is not set, then we behave as before and only verify
the repository.  This is partly because it's handy in development, and
partly because we still need the verify actions to support future plans.

### DIFF
--- a/src/common/actions/repository-input.js
+++ b/src/common/actions/repository-input.js
@@ -2,6 +2,8 @@ import 'isomorphic-fetch';
 
 import conf from '../helpers/config';
 
+const BASE_URL = conf.get('BASE_URL');
+
 export const SET_GITHUB_REPOSITORY = 'SET_GITHUB_REPOSITORY';
 export const VERIFY_GITHUB_REPOSITORY = 'VERIFY_GITHUB_REPOSITORY';
 export const VERIFY_GITHUB_REPOSITORY_SUCCESS = 'VERIFY_GITHUB_REPOSITORY_SUCCESS';
@@ -65,7 +67,6 @@ export function createSnap(repository, location) {
         payload: repository
       });
 
-      const BASE_URL = conf.get('BASE_URL');
       return fetch(`${BASE_URL}/api/launchpad/snaps`, {
         method: 'POST',
         headers: { 'Content-Type': 'application/json' },
--- a/src/common/actions/repository-input.js
+++ b/src/common/actions/repository-input.js
@@ -1,9 +1,13 @@
 import 'isomorphic-fetch';
 
+import conf from '../helpers/config';
+
 export const SET_GITHUB_REPOSITORY = 'SET_GITHUB_REPOSITORY';
 export const VERIFY_GITHUB_REPOSITORY = 'VERIFY_GITHUB_REPOSITORY';
 export const VERIFY_GITHUB_REPOSITORY_SUCCESS = 'VERIFY_GITHUB_REPOSITORY_SUCCESS';
 export const VERIFY_GITHUB_REPOSITORY_ERROR = 'VERIFY_GITHUB_REPOSITORY_ERROR';
+export const CREATE_SNAP = 'CREATE_SNAP';
+export const CREATE_SNAP_ERROR = 'CREATE_SNAP_ERROR';
 
 export function setGitHubRepository(value) {
   return {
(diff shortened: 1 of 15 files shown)


src/common/actions/repository-input.js
```diff
@@ -52,3 +56,49 @@ export function verifyGitHubRepositoryError(error) {
...
+
+export function createSnap(repository, location) {
+  return (dispatch) => {
+    if (repository) {
+      dispatch({
+        type: CREATE_SNAP,
+        payload: repository
+      });
+
+      const BASE_URL = conf.get('BASE_URL');
```
katiefenn: This could live with the other constants the top of the file.

src/common/actions/repository-input.js
```diff
@@ -52,3 +56,49 @@ export function verifyGitHubRepositoryError(error) {
...
+                result.payload.code !== 'snap-created') {
+              const error = new Error(response.statusText);
+              error.response = response;
+              throw error;
+            }
+            const startingUrl = `${BASE_URL}/${repository}/builds`;
+            (location || window.location).href =
+              `${BASE_URL}/login/authenticate` +
+              `?starting_url=${encodeURIComponent(startingUrl)}` +
+              `&caveat_id=${encodeURIComponent(result.payload.message)}`;
```
katiefenn: Location can be changed using `react-router-redux`, which loads a different route without a full page request. There's an example of this in my PR [here](https://github.com/canonical-ols/build.snapcraft.io/pull/41/files#diff-4643f606293beb5e4781b5aa11c170e0R30).

This requires the router middleware to be added to the store, which has not yet been merged into master. An example of this can be found [here](https://github.com/canonical-ols/build.snapcraft.io/pull/41/files#diff-6a5ca9964ee5cf49b3b87af20da794e4).

src/common/actions/repository-input.js
cjwatson: I've been getting out of the habit of doing that for configuration items, because it makes testing harder in cases where we need to vary the configuration; but it's true that in this case that doesn't matter.  I've moved this one.

src/common/components/repository-input/index.js
```diff
@@ -48,6 +50,24 @@ export class RepositoryInput extends Component {
...
+      );
+    } else {
+      // If the Launchpad API isn't configured, fall back to just verifying
+      // the repository.  This is handy in development.
+      submitButton = (
+        <Button type='submit' disabled={!isValid || input.isFetching}>
+          { input.isFetching ? 'Verifying...' : 'Verify' }
+        </Button>
+      );
+    }
```
katiefenn: Breaking snippets of JSX out into separate methods generally results in cleaner code. There's a few examples [here](https://github.com/canonical-ols/build.snapcraft.io/pull/41/files#diff-b287bbebb1e8c31d3df3068963f0d3e9R55).

bartaz: Yes, it would be good to separate those, maybe even into smaller components when it makes sense, but at this point we kind of discover how flow of the UI should look like, so let's keep it all here as simple as possible and we will refactor once we have everything in place.

cjwatson: I've split out a separate `submitButton` method.

Yes, I agree this overlaps with your PR, and I don't mean this to supersede yours; I just wanted to get to the point of being able to actually test that we have working end-to-end snap creation as quickly as possible.  I don't mind if you then replace it with your flow.  (Note though that we don't yet have a backend method to dispatch a build ...)

test/routes/src/server/routes/t_login.js
```diff
@@ -60,32 +60,115 @@ describe('login routes', () => {
...
-          }
-        })
-        .expect(302, done);
+    context('with no options', () => {
+      it('should redirect from /login/authenticate to SSO', (done) => {
+        supertest(app)
+          .get('/login/authenticate')
+          .expect(res => {
+            const loc = res.header['location'];
+            if (!loc.startsWith(UBUNTU_SSO_URL)) {
```
bartaz: Instead of "manual" comparisons and throwing errors it's possible to use regular `expect` as in other (unit) tests.

As in: https://github.com/canonical-ols/build.snapcraft.io/blob/master/test/routes/src/server/routes/t_github-auth.js#L111-L115

```js
.expect((res) => {
  expect(res.statusCode).toEqual(302);
  expect(res.headers.location).toEqual('/');
});
```

Not a blocking change, but can make some of the checks more readable.
Also applies to rest of the tests in the file.

src/common/components/repository-input/index.js
katiefenn: @cjwatson: Yep - my comment was more to give you the information should you wish to abort that change. Continuing is also fine. 👍

src/common/components/repository-input/index.js
```diff
@@ -42,6 +44,25 @@ export class RepositoryInput extends Component {
     }
   }
 
+  submitButton(input, isValid) {
+    if (conf.get('LP_API_USERNAME')) {
+      // Main path, for use in production.
+      return (
+        <Button type='submit' disabled={!isValid || input.isFetching}>
+          { input.isFetching ? 'Creating...' : 'Create' }
```
katiefenn: ...or even creating a method called "getButtonLabel", or similar.

src/server/helpers/config.js
```diff
@@ -9,7 +9,8 @@ import dotenv from 'dotenv';
  */
 const CLIENT_SIDE_WHITELIST = [
   'NODE_ENV',
-  'BASE_URL'
+  'BASE_URL',
+  'LP_API_USERNAME'
```
katiefenn: Is there any use-case where it would not be configured?

src/common/actions/repository-input.js
```diff
@@ -52,3 +58,48 @@ export function verifyGitHubRepositoryError(error) {
...
+        .then(checkStatus)
+        .then(response => {
+          return response.json().then(result => {
+            if (result.status !== 'success' ||
+                result.payload.code !== 'snap-created') {
+              const error = new Error(response.statusText);
+              error.response = response;
+              throw error;
+            }
+            const startingUrl = `${BASE_URL}/${repository}/builds`;
```
cjwatson: I think there's probably no point at the moment, since at this point we have to redirect through OpenID so we're about to have to reload the store from scratch anyway.

src/server/helpers/config.js
cjwatson: It means that things wouldn't work during development unless you supplied Launchpad credentials.  That's fine by me (and I could make the errors in that case a little more graceful), but are you two OK with that?

src/common/components/repository-input/index.js
```diff
@@ -42,6 +44,25 @@ export class RepositoryInput extends Component {
     }
   }
 
+  submitButton(input, isValid) {
+    if (conf.get('LP_API_USERNAME')) {
```
cjwatson: Moot now that I've removed the feature switch.

src/common/components/repository-input/index.js
cjwatson: Moot now that I've removed the feature switch.

src/server/helpers/config.js
cjwatson: I've added a script to create Launchpad credentials, referred to it from `docs/environment-variables.md`, and removed this feature switch.

test/routes/src/server/routes/t_login.js
cjwatson: Done, thanks.

PR #43, opened 2016-12-20
Added API route to get snap builds from LP

bartaz:
As a first step to get real data into builds views I added a new `/api/launchpad/builds` route to get builds list from LP for given snap (by `self_link`).


src/server/handlers/launchpad.js
```diff
@@ -301,3 +301,57 @@ export const completeSnapAuthorization = async (req, res) => {
...
 };
+
+export const getSnapBuilds = (req, res) => {
+  const snap_link = req.query.snap_link;
+
+  if (snap_link) {
+    return getLaunchpad().get(snap_link).then(snap => {
+      // TODO: bartaz: currently just limiting to 10 recent builds
+      // we will need to process it properly later for paging/loading more
+      return getLaunchpad().get(snap.builds_collection_link, { start: 0, size: 10 })
```
bartaz: Currently just fetching 10 latest builds from LP.

Eventually we need to properly handle loading/paging here.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -569,6 +570,214 @@ describe('The Launchpad API endpoint', () => {
...
+
+    });
+
+    context('when snap fails to fetch', () => {
+
+      beforeEach(() => {
+        // when getting snap data from API (via self_link)
+        nock(lp_api_url)
+          .get(lp_snap_path)
+          .reply(404, 'Not found');
```
bartaz: @cjwatson  Are there any specific scenarios/cases we would like to cover here?

What possible errors can we get when trying to get snap?

src/server/handlers/launchpad.js
```diff
@@ -301,3 +301,57 @@ export const completeSnapAuthorization = async (req, res) => {
...
+        .then(builds => {
+          return res.status(200).send({
+            status: 'success',
+            payload: {
+              code: 'snap-builds-found',
+              message: builds.entries
```
bartaz: @katiefenn @cjwatson I wanted to follow the response convention we use here (to have `code` and `message` in payload), but I'm not sure if array of builds actually should go to message here? Or should it be rather separate field?

src/server/handlers/launchpad.js
cjwatson: It wouldn't be much effort to make `start` and `size` be query parameters to this route, but this is OK for now.

src/server/handlers/launchpad.js
cjwatson: I'd be inclined to make it a separate field so that we don't end up with `PropTypes` confusion or similar later on.  `builds` should be fine.

src/server/handlers/launchpad.js
```diff
@@ -301,3 +301,57 @@ export const completeSnapAuthorization = async (req, res) => {
...
+          payload: {
+            code: 'internal-error',
+            message: error.message
+          }
+        });
+      }
+    });
+  }
+  else {
+    return res.status(404).send({
```
cjwatson: I'm generally inclined to put this sort of "is the request well-formed?" check up at the top with an early return, rather than buried down at the bottom of an if/else chain.  This makes it easier to see what condition the error is associated with, and it reduces indentation depth.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -569,6 +570,214 @@ describe('The Launchpad API endpoint', () => {
...
+              {
+                'self_link': `${lp_api_url}${lp_snap_path}/+build/12345`,
+                'date_started': '2016-12-06T10:40:17.751937+00:00',
+                'datebuilt': '2016-12-06T10:47:59.384231+00:00',
+                'duration': '0:07:41.632294',
+                'datecreated': '2016-12-06T10:30:29.285058+00:00',
+                'buildstate': 'Successfully built',
+                'resource_type_link': `${lp_api_url}/devel/#snap_build`,
+                'build_log_url': `${lp_api_url}${lp_snap_path}/+build/12345/+files/buildlog_snap_ubuntu_xenial_armhf_${lp_snap_name}_BUILDING.txt.gz`,
+                'arch_tag': 'armhf',
```
cjwatson: You don't need fully-formed test data here.  For readability, I'd trim this down to something pretty minimal, maybe just `resource_type_link` and `self_link`.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -569,6 +570,214 @@ describe('The Launchpad API endpoint', () => {
...
+          });
+
+        // when getting builds list (via builds_collection_link)
+        nock(lp_api_url)
+          .get(lp_builds_path)
+          .query({
+            'ws.start': 0,
+            'ws.size': 10
+          })
+          .reply(404, 'Not found');
```
cjwatson: The structure of the route tests should generally follow the structure of the route itself, so I think it's only important that they test that errors returned by Launchpad are passed through properly, as you do here.  The details of how the Launchpad client behaves belong in the unit tests for the Launchpad client itself.

test/routes/src/server/routes/t_launchpad.js
cjwatson: Same comment as above.  I think this level of detail is fine here.

src/server/handlers/launchpad.js
katiefenn: My intention was for "code" to be a unique identifier for an outcome that could be matched to an error message to be shown to the user. The "message" was to be a human-readable alternative to describe what happened to a developer reading the network logs in devtools. These two fields (with the status field) are fine for describing errors. "code" might even be sufficiently readable on its own.

I think we should pass builds as its own property in this case.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -569,6 +570,214 @@ describe('The Launchpad API endpoint', () => {
...
+          .expect(hasMessage('snap-builds-found'))
+          .end(done);
+      });
+
+      it('should return builds list as payload message', (done) => {
+        supertest(app)
+          .get('/launchpad/builds')
+          .query({ snap_link: lp_snap_link })
+          .end((err, res) => {
+            expect(res.body.payload.message.length).toEqual(1);
```
katiefenn: Can the content of the builds list and the statusCode be tested as well? The endpoint could return a 404 and an error payload and I _think_ this may report a false positive.

src/server/handlers/launchpad.js
```diff
@@ -301,3 +301,57 @@ export const completeSnapAuthorization = async (req, res) => {
...
+          logger.info('Launchpad API error:', text);
+          return res.status(error.response.status).send({
+            status: 'error',
+            payload: {
+              code: 'lp-error',
+              message: text
+            }
+          });
+        });
+      } else {
```
katiefenn: Another opportunity for early return here.

bartaz: Not really sure about that.

There are 2 branches of exactly the same size. So the only change would turning `if (error.response)` to `if (!error.response)` and changing the order of returns.

test/routes/src/server/routes/t_launchpad.js
bartaz: Error message would possibly be grater than one, but agreed.
Together with other changes updated this to check for `self_link` in the build props.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -569,6 +570,272 @@ describe('The Launchpad API endpoint', () => {
...
+          .end((err, res) => {
+            const build = res.body.payload.builds[0];
+            // XXX bartaz
+            // Wanted to use `expect(build).toEqual(test_build)`
+            // but LP client adds more propeties to the object...
+            // also `expect(build).toContain(test_build)` doesnt work
+            // because `expect` fails to iterate over keys on the object
+            // for some reason (async maybe).
+            // To be investigated later...
+            expect(build.self_link).toEqual(test_build.self_link);
```
bartaz: @cjwatson @katiefenn One thing I noticed (may be worth remembering if you ever have similar issue) is that objects returned from LP client are not pure JSON objects, but have classes defined in `launchpad/resources.js`.

They have additional properties (not just what is returned by API) and for some reason `expect` has some trouble iterating over the keys in object when using `expect().toContain()`.

It's nothing really blocking ATM, but it may hit us again. It may be worth investigating when we have more time why it happens and if we can/need to do something about it.

katiefenn:
LGTM

PR #44, opened 2016-12-21
Create snaps with more appropriate architectures

cjwatson:
Launchpad defaults to [amd64, i386], which is possibly not ideal for
snaps.  Let's go for [amd64, armhf] instead; we need this to be
user-selectable in the long run, but this will be better for demo
purposes.

In the process, I fixed some buggy use of nock in tests: the Launchpad
client passes parameters to POST requests using the request body, not
the query string.

### DIFF
--- a/test/routes/src/server/routes/t_launchpad.js
+++ b/test/routes/src/server/routes/t_launchpad.js
@@ -78,8 +78,12 @@ describe('The Launchpad API endpoint', () => {
           .reply(200, 'name: test-snap\n');
         const lp_api_url = conf.get('LP_API_URL');
         nock(lp_api_url)
-          .post('/devel/+snaps')
-          .query({ 'ws.op': 'new' })
+          .post('/devel/+snaps', {
+            'ws.op': 'new',
+            git_repository_url: 'https://github.com/anaccount/arepo',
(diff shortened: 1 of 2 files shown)


test/routes/src/server/routes/t_launchpad.js
```diff
@@ -78,8 +78,12 @@ describe('The Launchpad API endpoint', () => {
           .reply(200, 'name: test-snap\n');
         const lp_api_url = conf.get('LP_API_URL');
         nock(lp_api_url)
-          .post('/devel/+snaps')
-          .query({ 'ws.op': 'new' })
+          .post('/devel/+snaps', {
+            'ws.op': 'new',
+            git_repository_url: 'https://github.com/anaccount/arepo',
+            processors: ['/+processors/amd64', '/+processors/armhf'],
```
bartaz: I was thinking that maybe it would be worth adding a test to check if expected default architectures are properly passed to LP.

But I guess this test would simply check if this nock mock expectations are fulfilled which in theory it already does - if different archs are passed nock shouldn't catch this request and tests should fail loudly...

cjwatson: Yep, exactly that.  I think it's probably best not to duplicate the expectations.

PR #45, opened 2016-12-21
Getting snap builds from LP into views

bartaz:
screen shot 2016-12-22 at 13 38 22

Going to `/user/repo/builds/` now properly asks LP for snap for given repository and then for builds of this snap.

To test is locally you need to have LP_ env variables properly set up with your user name and auth tokens. Also you need to have a snap created from GH url in LP (for your user) to make it available for your app.

### TODO
- [x] action to get builds from LP API (with hardcoded snap URL)
- [x] action to get snap URL for repository (with mocked response)
- [x] really fetch snap URL from LP API (needs proper LP credentials and some build to test)

### DIFF
--- a/src/common/actions/snap-builds.js
+++ b/src/common/actions/snap-builds.js
@@ -39,19 +39,13 @@ export function fetchSnap(repository) {
         type: FETCH_BUILDS
       });
 
-      return new Promise((resolve) => {
-        setTimeout(() => {
-          resolve({
-            status: 'success',
-            payload: {
-              code: 'snap-found',
-              message: 'https://api.launchpad.net/devel/~snappy-dev/+snap/core'
-            }
-          });
-        }, 1000);
-      })
-      .then(json => dispatch(fetchBuilds(json.payload.message)))
-      .catch(error => dispatch(fetchBuildsError(error)));
+      const repositoryUrl = encodeURIComponent(`https://github.com/${repository}.git`);
+      const url = `${BASE_URL}/api/launchpad/snaps?repository_url=${repositoryUrl}`;
+      return fetch(url)
+        .then(checkStatus)
+        .then(response => response.json())
+        .then((json) => dispatch(fetchBuilds(json.payload.message)))
+        .catch( error => dispatch(fetchBuildsError(error)));
     }
   };
 }
--- a/src/common/actions/snap-builds.js
+++ b/src/common/actions/snap-builds.js
@@ -1,3 +1,7 @@
+import conf from '../helpers/config';
+
+const BASE_URL = conf.get('BASE_URL');
+
 export const FETCH_BUILDS = 'FETCH_BUILDS';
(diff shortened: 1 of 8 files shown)


src/common/actions/snap-builds.js
```diff
@@ -17,169 +21,52 @@ export function fetchBuildsError(error) {
   };
 }
 
-export function fetchBuilds(repository) {
+function checkStatus(response) {
+  if (response.status >= 200 && response.status < 300) {
+    return response;
+  } else {
+    return response.json().then(json => {
```
katiefenn: `json` is missing brackets / parens.

src/common/containers/build-details.js
```diff
@@ -26,6 +27,9 @@ class BuildDetails extends Component {
         { this.props.isFetching &&
           <span>Loading...</span>
         }
+        { this.props.error &&
+          <Message status='error'>{ this.props.error.message || this.props.error }</Message>
```
katiefenn: The PropTypes validation states `error` can be a bool. What is shown here when it is?

src/common/actions/snap-builds.js
cjwatson: Aren't they optional?  I've certainly been treating them as such.  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

src/common/actions/snap-builds.js
```diff
@@ -17,169 +21,52 @@ export function fetchBuildsError(error) {
...
-};
+export function fetchBuilds(snapLink) {
+  return (dispatch) => {
+    if (snapLink) {
+      dispatch({
+        type: FETCH_BUILDS
+      });
+
+      snapLink = encodeURIComponent(snapLink);
+      const url = `${BASE_URL}/api/launchpad/builds?snap_link=${snapLink}`;
```
cjwatson: I know the server handler is already in place, but I'd be inclined to call the query parameter here just `snap` rather than `snap_link`.  The Launchpad API sprinkles `_link` suffixes everywhere, but they're a bit cumbersome to deal with directly, and they're normally intended to be wrapped automatically by language bindings (even though our JS binding doesn't yet do so; the Python binding does).  It's probably best not to replicate that here.

src/common/actions/snap-builds.js
bartaz: They are optional, but I guess it would be good to be consistent about it from code style point of view.

Because you have to use parens when there are no args or multiple args (`() => { ... }`) I would agree to have a guideline to use parens always (for consistency).

src/common/containers/build-details.js
bartaz: Ah yes, it's a tricky one.

Currently in the store error can be `false` or error object. That's why I needed to add `PropTypes.bool` to make both valid values. Not sure if I can limit it to object and `false` only.

But I just realised using `null` instead of `false` would make it much less confusing.

PR #55, opened 2016-12-23
Check for admin permissions on repository

cjwatson:
We can only install a webhook if we have admin permissions, and in
general we only want to allow users to manipulate their own repositories
anyway.

### DIFF
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -225,8 +225,8 @@ const getSnapcraftYaml = (owner, name, token) => {
 
 export const newSnap = (req, res) => {
   const repositoryUrl = req.body.repository_url;
-  const lp_client = getLaunchpad();
-  let self_link;
+  const lpClient = getLaunchpad();
+  let snapUrl;
   // We need admin permissions in order to be able to install a webhook later.
   checkAdminPermissions(req)
     .then(([owner, name, token]) => getSnapcraftYaml(owner, name, token))
@@ -236,7 +236,7 @@ export const newSnap = (req, res) => {
       }
       const username = conf.get('LP_API_USERNAME');
       logger.info(`Creating new snap for ${repositoryUrl}`);
-      return lp_client.named_post('/+snaps', 'new', {
+      return lpClient.named_post('/+snaps', 'new', {
         parameters: {
           owner: `/~${username}`,
           distro_series: `/${DISTRIBUTION}/${DISTRO_SERIES}`,
@@ -256,12 +256,12 @@ export const newSnap = (req, res) => {
       });
     })
     .then((result) => {
-      self_link = result.self_link;
-      logger.info(`Authorizing ${self_link}`);
-      return lp_client.named_post(self_link, 'beginAuthorization');
+      snapUrl = result.self_link;
+      logger.info(`Authorizing ${snapUrl}`);
+      return lpClient.named_post(snapUrl, 'beginAuthorization');
     })
     .then((caveatId) => {
-      logger.info(`Began authorization of ${self_link}`);
+      logger.info(`Began authorization of ${snapUrl}`);
       return res.status(201).send({
         status: 'success',
         payload: {
@@ -319,12 +319,12 @@ const internalFindSnap = async (repositoryUrl) => {
 
 export const findSnap = (req, res) => {
   internalFindSnap(req.query.repository_url)
-    .then((snap_link) => {
+    .then((snapUrl) => {
       return res.status(200).send({
         status: 'success',
         payload: {
           code: 'snap-found',
-          message: snap_link
+          message: snapUrl
         }
       });
     })
@@ -332,22 +332,22 @@ export const findSnap = (req, res) => {
 };
 
 export const completeSnapAuthorization = async (req, res) => {
-  let snap_link;
+  let snapUrl;
   checkAdminPermissions(req)
     .then(() => internalFindSnap(req.body.repository_url))
     .then((result) => {
-      snap_link = result;
-      return getLaunchpad().named_post(snap_link, 'completeAuthorization', {
+      snapUrl = result;
+      return getLaunchpad().named_post(snapUrl, 'completeAuthorization', {
         parameters: { discharge_macaroon: req.body.discharge_macaroon },
       });
     })
     .then(() => {
-      logger.info(`Completed authorization of ${snap_link}`);
+      logger.info(`Completed authorization of ${snapUrl}`);
       return res.status(200).send({
         status: 'success',
         payload: {
           code: 'snap-authorized',
-          message: snap_link
+          message: snapUrl
         }
       });
     })
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -97,11 +97,24 @@ export const setMemcached = (value) => {
   memcached = value;
 };
 
+class PreparedError extends Error {
+  constructor(status, body) {
+    super();
+    this.status = status;
+    this.body = body;
+  }
+}
+
 const responseError = (res, error) => {
-  if (error.response) {
+  if (error.status && error.body) {
+    // The error comes with a prepared representation.
+    return res.status(error.status).send(error.body);
+  } else if (error.response) {
     // if it's ResourceError from LP client at least for the moment
     // we just wrap the error we get from LP
-    return error.response.text().then(text => {
+    // XXX cjwatson 2016-12-22: Perhaps refactor to use the PreparedError
+    // system above?
+    return error.response.text().then((text) => {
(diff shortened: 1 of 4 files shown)


src/server/helpers/github.js
```diff
@@ -11,31 +11,35 @@ import { conf } from '../helpers/config';
...
+    const params = {
+      ...options,
+      proxy: HTTP_PROXY
+    };
+    params.uri = GITHUB_API_ENDPOINT + params.uri;
+    if (!params.headers) {
+      params.headers = {};
+    }
+    params.headers['User-Agent'] = 'SnapcraftBuild';
+    request(params, (err, response) => {
```
bartaz: I was wondering if using fetch from `isomorphic-fetch` wont be simpler and more consistent (as it is already based on promises), but as it's server side only and we have it set up to work with proxy it's probably not worth changing.

src/server/helpers/github.js
```diff
@@ -11,31 +11,35 @@ import { conf } from '../helpers/config';
...
-    ...options,
-    proxy: HTTP_PROXY
-  };
-  params.uri = GITHUB_API_ENDPOINT + params.uri;
-  if (!params.headers) {
-    params.headers = {};
-  }
-  params.headers['User-Agent'] = 'SnapcraftBuild';
-  return request(params, callback);
+export const requestGitHub = options => {
```
bartaz: I would be for using `(options) =>` as it makes it a little bit clearer that is a function with an argument.

src/server/helpers/github.js
```diff
@@ -11,31 +11,35 @@ import { conf } from '../helpers/config';
...
+  });
 };
 
-requestGitHub.get = (uri, options, callback) => {
-  const params = { ...options };
-  params.uri = uri;
-  params.method = 'GET';
-  return requestGitHub(params, callback);
+requestGitHub.get = (uri, options) => {
+  const params = { ...options, uri: uri, method: 'GET' };
```
bartaz: When you pass params with the same name you may even omit the value, like:

```js
const params = { ...options, method: 'GET', uri }
```

src/server/handlers/launchpad.js
```diff
@@ -97,11 +105,24 @@ export const setMemcached = (value) => {
...
 const responseError = (res, error) => {
-  if (error.response) {
+  if (error.status && error.body) {
+    // The error comes with a prepared representation.
+    return res.status(error.status).send(error.body);
+  } else if (error.response) {
     // if it's ResourceError from LP client at least for the moment
     // we just wrap the error we get from LP
-    return error.response.text().then(text => {
+    // XXX cjwatson 2016-12-22: Perhaps refactor to use the PreparedError
```
bartaz: I agree that it's a bit confusing right now with all the 'layers' of error we have (ResourceError, PreparedError, vanilla Errors...).

As ResourceError already seems to have a reference to response, maybe it could already have status and body set (be 'PreparedError' or at least in compatible format).

src/server/handlers/launchpad.js
```diff
@@ -122,107 +143,137 @@ const responseError = (res, error) => {
...
-  const uri = `/repos/${parsed.owner}/${parsed.name}/contents/snapcraft.yaml`;
+  const uri = `/repos/${parsed.owner}/${parsed.name}`;
+  const options = {
+    headers: { 'Authorization': `token ${token}` },
+    json: true
+  };
+  logger.info(`Checking permissions for ${parsed.owner}/${parsed.name}`);
+  return requestGitHub.get(uri, options)
+    .then(checkStatus)
+    .then(response => {
```
bartaz: Again, parens around param. For consistency.

src/server/handlers/launchpad.js
```diff
@@ -122,107 +143,137 @@ const responseError = (res, error) => {
   }
 };
 
-const makeSnapName = url => {
-  return createHash('md5').update(url).digest('hex');
+const checkStatus = (response) => {
```
bartaz: If it's only for GH (as it seems from possible error responses) it would be good to name it to reflect that. So it's not used elsewhere as a generic 'checkStatus'.

src/server/handlers/launchpad.js
```diff
@@ -122,107 +143,137 @@ const responseError = (res, error) => {
...
+const makeSnapName = (url) => {
+  return createHash('md5').update(url).digest('hex');
+};
+
+const getSnapcraftYaml = (owner, name, token) => {
+  const uri = `/repos/${owner}/${name}/contents/snapcraft.yaml`;
   const options = {
     headers: {
-      'Authorization': `token ${req.session.token}`,
+      'Authorization': `token ${token}`,
```
bartaz: We add the 'Authorization' header to every GH request, so maybe `requestGitHub` could do it? It probably can't take token from session on its own, but we can just pass it as param or part of options (not building whole header value each time).

src/server/handlers/launchpad.js
```diff
@@ -122,107 +143,137 @@ const responseError = (res, error) => {
...
-        auto_build_archive: `/${DISTRIBUTION}/+archive/primary`,
-        auto_build_pocket: 'Updates',
-        processors: ARCHITECTURES.map(arch => {
-          return `/+processors/${arch}`;
-        }),
-        store_upload: true,
-        store_series: `/+snappy-series/${STORE_SERIES}`,
-        store_name: snapcraftYaml.name
+  const repositoryUrl = req.body.repository_url;
+  const lp_client = getLaunchpad();
```
bartaz: It already spread out a little bit in other places too as I noticed (I myself followed it in lauchpad tests I guess), but we should use camelCase consistently for JS variables.

src/server/handlers/launchpad.js
```diff
@@ -122,107 +143,137 @@ const responseError = (res, error) => {
...
+    json: true
+  };
+  logger.info(`Checking permissions for ${parsed.owner}/${parsed.name}`);
+  return requestGitHub.get(uri, options)
+    .then(checkStatus)
+    .then(response => {
+      if (!response.body.permissions || !response.body.permissions.admin) {
+        throw new PreparedError(401, RESPONSE_GITHUB_NO_ADMIN_PERMISSIONS);
+      }
+      return [parsed.owner, parsed.name, token];
```
bartaz: I think it would be more explicit to return it as an object with named keys `{ owner: parsed.owner, name: parsed.name, token }` instead of 'anonymous' values in array.

src/server/handlers/launchpad.js
```diff
@@ -122,107 +143,137 @@ const responseError = (res, error) => {
...
+      });
+    })
+    .catch((error) => responseError(res, error));
 };
 
-const internalFindSnap = async repositoryUrl => {
+const internalFindSnap = async (repositoryUrl) => {
   const cacheId = `url:${repositoryUrl}`;
 
   return new Promise((resolve, reject) => {
```
bartaz: Why returning `new Promise` when `named_get` is already a Promise?

Just realised `named_get` should also be camelCase in theory... but we don't have to update it all at once.

src/server/handlers/launchpad.js
```diff
@@ -243,77 +311,47 @@ const internalFindSnap = async repositoryUrl => {
...
 export const completeSnapAuthorization = async (req, res) => {
-  // XXX cjwatson 2016-12-15: Limit to only repositories the user owns.
-  if (!req.session || !req.session.token) {
-    return res.status(401).send(RESPONSE_NOT_LOGGED_IN);
-  }
-
-  internalFindSnap(req.body.repository_url).then(self_link => {
-    if (self_link !== null) {
-      return getLaunchpad().named_post(self_link, 'completeAuthorization', {
+  let snap_link;
```
bartaz: camelCase

src/server/helpers/github.js
cjwatson: I tried using `fetch`, but it doesn't support proxies at all, so that turned out to be infeasible.

src/server/handlers/launchpad.js
cjwatson: Definitely worth refactoring at some point, although it's not clear how much we should push down to the LP client since ideally that should be split off to its own package and not hardcode BSI's error formats.  I'm going to defer this one.

src/server/handlers/launchpad.js
cjwatson: `memcached` doesn't return a `Promise`, unfortunately, and I think I'm doing something wrong when I try to simplify this any further.  In the new year perhaps we can track down a suitable promise-enabled memcached client library and use that instead.

src/server/handlers/launchpad.js
cjwatson: Good suggestion, thanks - done.

PR #57, opened 2016-12-23
Trigger builds on snap creation and webhook notify

cjwatson:


test/routes/src/server/routes/t_webhook.js
```diff
@@ -10,7 +12,45 @@ describe('The WebHook API endpoint', () => {
...
+    const overrides = conf.stores['test-overrides'];
+    overrides.clear('LP_API_URL');
+    overrides.clear('LP_API_USERNAME');
+    overrides.clear('LP_API_CONSUMER_KEY');
+    overrides.clear('LP_API_TOKEN');
+    overrides.clear('LP_API_TOKEN_SECRET');
+  });
+
+  beforeEach(() => {
+    const memcachedStub = { cache: {} };
```
bartaz: It would be good if we can make memcached optional, so we don't need to set it up in tests when not necessary.
Especially here, webhook tests shouldn't really worry about setting memcached for launchpad handler...

But that's something we can fix separately and doesn't affect this branch.

bartaz: Made an issue so we don't lose track of that: https://github.com/canonical-ols/build.snapcraft.io/issues/59

PR #58, opened 2016-12-23
Tidy up return value of checkAdminPermissions

cjwatson:


cjwatson:
This was left over from a review comment in https://github.com/canonical-ols/build.snapcraft.io/pull/55#discussion_r93777504 that I missed by accident, so self-approving.

PR #60, opened 2017-01-02
Making memcached optional (for development and tests)

bartaz:
If `MEMCACHED_HOST` is not configured launchpad handler will just pass through memcached calls without caching anything.

This allows to run dev server without running memcached server and test without stubbing memcached.

### DIFF
--- a/src/server/helpers/memcached.js
+++ b/src/server/helpers/memcached.js
@@ -1,0 +1,30 @@
+import Memcached from 'memcached';
+
+import { conf } from '../helpers/config';
+
+let memcached = null;
+
+const getMemcachedStub = () => {
+  return {
+    get: (key, callback) => callback(),
+    set: (key, value, lifetime, callback) => callback()
+  };
+};
+
+export const getMemcached = () => {
+  const host = conf.get('MEMCACHED_HOST') ? conf.get('MEMCACHED_HOST').split(',') : null;
+
+  if (memcached === null) {
+    if (host) {
+      memcached = new Memcached(host, { namespace: 'lp:' });
+    } else {
+      memcached = getMemcachedStub();
+    }
+  }
+  return memcached;
+};
+
+// Test affordance.
+export const setMemcached = (value) => {
+  memcached = value;
+};
--- a/src/server/helpers/memcached.js
+++ b/src/server/helpers/memcached.js
@@ -11,6 +11,24 @@ const getMemcachedStub = () => {
   };
 };
 
+const getInMemoryMemcachedStub = () => {
+  const memcachedStub = { cache: {} };
+
+  memcachedStub.get = (key, callback) => {
+    if (callback) {
+      callback(undefined, memcachedStub.cache[key]);
+    }
+  };
+  memcachedStub.set = (key, value, lifetime, callback) => {
+    memcachedStub.cache[key] = value;
+    if (callback) {
+      callback(undefined, true);
+    }
+  };
+
+  return memcachedStub;
+};
+
 export const getMemcached = () => {
   const host = conf.get('MEMCACHED_HOST') ? conf.get('MEMCACHED_HOST').split(',') : null;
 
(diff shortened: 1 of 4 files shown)


src/server/helpers/memcached.js
```diff
@@ -0,0 +1,30 @@
+import Memcached from 'memcached';
+
+import { conf } from '../helpers/config';
+
+let memcached = null;
+
+const getMemcachedStub = () => {
+  return {
+    get: (key, callback) => callback(),
+    set: (key, value, lifetime, callback) => callback()
```
cjwatson: I'd suggest making the default stub use a trivial in-memory cache object, the way that some tests do.  That would make it a more accurate stub and would simplify tests further.

src/server/helpers/memcached.js
```diff
@@ -0,0 +1,30 @@
+import Memcached from 'memcached';
+
+import { conf } from '../helpers/config';
+
+let memcached = null;
+
+const getMemcachedStub = () => {
+  return {
+    get: (key, callback) => callback(),
```
bartaz: Just a simple 'pass-through' stub - nothing is cached, only callback is called (with undefined value).

src/server/helpers/memcached.js
```diff
@@ -0,0 +1,30 @@
...
+    set: (key, value, lifetime, callback) => callback()
+  };
+};
+
+export const getMemcached = () => {
+  const host = conf.get('MEMCACHED_HOST') ? conf.get('MEMCACHED_HOST').split(',') : null;
```
bartaz: Should we require `MEMCACHED_HOST` to be defined in production environment?
Or is assuming that config will be complete is good enough?

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -473,25 +458,58 @@ describe('The Launchpad API endpoint', () => {
...
+          callback(undefined, true);
+        };
+        setMemcached(memcachedStub);
+
+        memcachedStub.cache[`url:${repositoryUrl}`] = snapUrl;
+      });
+
+      afterEach(() => {
+        setMemcached(null);
+      });
```
cjwatson: This pair is unnecessarily wordy given that the containing `describe` already does almost the same thing.  I think you could just have:

```
beforeEach(() => {
  getMemcached().set(`url:${repositoryUrl}`, snapUrl);
});
```

... and omit `afterEach` entirely.

test/routes/src/server/routes/t_webhook.js
```diff
@@ -30,20 +29,7 @@ describe('The WebHook API endpoint', () => {
     overrides.clear('LP_API_TOKEN_SECRET');
   });
 
-  beforeEach(() => {
```
bartaz: Webhook tests should not be concerned about memcached at all.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -315,21 +315,6 @@ describe('The Launchpad API endpoint', () => {
   });
 
   describe('find snap route', () => {
-    beforeEach(() => {
-      const memcachedStub = { cache: {} };
```
bartaz: Removed unnecessary memcached stubs.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -473,25 +458,58 @@ describe('The Launchpad API endpoint', () => {
...
+        memcachedStub.get = (key, callback) => {
+          callback(undefined, memcachedStub.cache[key]);
+        };
+        memcachedStub.set = (key, value, lifetime, callback) => {
+          memcachedStub.cache[key] = value;
+          callback(undefined, true);
+        };
+        setMemcached(memcachedStub);
+
+        memcachedStub.cache[`url:${repositoryUrl}`] = snapUrl;
```
bartaz: Testing if values from memcached are properly returned.

src/server/helpers/memcached.js
cjwatson: I think it's OK to leave this up to the config.

PR #61, opened 2017-01-03
Autorefresh builds list

bartaz:
screen shot 2017-01-03 at 14 56 07

Builds list view is now auto fetching builds list every 5sec to keep builds status up to date.

### DIFF
--- a/src/common/containers/builds.js
+++ b/src/common/containers/builds.js
@@ -25,6 +25,7 @@ class Builds extends Component {
 
   render() {
     const { account, repo, fullName } = this.props;
+    const isLoading = this.props.isFetching && !this.props.success;
 
     return (
       <div className={ styles.container }>
@@ -33,7 +34,7 @@ class Builds extends Component {
         />
         <h1>{fullName} builds</h1>
         <BuildHistory account={account} repo={repo}/>
-        { this.props.isFetching &&
+        { isLoading &&
           <span>Loading...</span>
         }
         { this.props.error &&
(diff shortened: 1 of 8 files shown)


src/common/containers/builds.js
```diff
@@ -25,7 +36,7 @@ class Builds extends Component {
         <h1>{fullName} builds</h1>
         <BuildHistory account={account} repo={repo}/>
         { this.props.isFetching &&
-          <span>Loading...</span>
+          <div className={styles.spinner}><Spinner /></div>
```
bartaz: Not sure if we want to keep spinner visible every time we fetch (even if list of builds is already displayed), but at least now it's a nice way to show that auto refresh works (and something is going on in background).

src/common/containers/builds.js
```diff
@@ -4,14 +4,25 @@ import Helmet from 'react-helmet';
...
 
 class Builds extends Component {
+  fetchInterval = null
 
   componentWillMount() {
     this.props.dispatch(fetchSnap(this.props.fullName));
+
+    this.fetchInterval = setInterval(() => {
+      this.props.dispatch(fetchSnap(this.props.fullName));
+    }, 5000);
```
bartaz: Is 5s OK or too often? We don't want to flood LP with requests, but on the other hand we want to keep user view up to date.

cjwatson: LP's buildd-manager cycle is 15 seconds, so there's no point in making requests any more frequently than that as they can't possibly have new data.  15 seconds should be OK.

src/common/containers/builds.js
cjwatson: Yeah, I think repeatedly popping up the spinner will get pretty distracting.  I'd stick to devtools for demonstrating that auto-refresh works.

PR #62, opened 2017-01-03
Finish hooking up completeSnapAuthorization

cjwatson:
It appears that I never actually arranged for completeSnapAuthorization
to be called on creating a new snap, so let's fix that.  The
authorization workflow goes through the server-rendered page
/login/verify after returning from SSO, so the simplest way to do things
is just to have that call completeSnapAuthorization directly rather than
via a route.  We need to add the GitHub repository URL to the parameters
passed through the OpenID dance in order for this to work.


src/server/handlers/launchpad.js
```diff
@@ -387,5 +383,6 @@ export const requestSnapBuilds = (req, res) => {
         }
       });
     })
-    .catch((error) => responseError(res, error));
+    .catch((error) => prepareError(error))
+    .catch((error) => res.status(error.status).send(error.body));
```
bartaz: As `prepareError` seems to be used mostly just to catch it again and return, maybe we could have `prepareErrorResponse(error, res)` instead that would format the error if necessary and return it as the response? This would avoid this a bit strange double `catch` pattern where we throw prepared error just to catch and return it in response.

test/unit/src/server/handlers/t_launchpad.js
```diff
@@ -0,0 +1,130 @@
...
+import nock from 'nock';
+
+import { completeSnapAuthorization } from '../../../../../src/server/handlers/launchpad';
+import { conf } from '../../../../../src/server/helpers/config.js';
+
+describe('completeSnapAuthorization', () => {
+  const session = { 'token': 'secret' };
+
+  before(() => {
+    const overrides = conf.stores['test-overrides'];
```
bartaz: We seem to do the same test config overrides in couple of places already.

Not something to fix in this branch, but maybe it would be worth extracting it to some helper functions or maybe we could just make it a part of test setup somehow.

src/server/handlers/launchpad.js
cjwatson: I've refactored it a bit.  Is this any better?  I had to extract at least part of the error preparation logic in order that `completeSnapAuthorization` could just throw errors rather than sending a response.

test/unit/src/server/handlers/t_launchpad.js
cjwatson: I've filed #64 for this.

src/server/handlers/launchpad.js
```diff
@@ -310,30 +316,27 @@ export const findSnap = (req, res) => {
...
-        payload: {
-          code: 'snap-authorized',
-          message: snapUrl
-        }
-      });
-    })
-    .catch((error) => responseError(res, error));
+    .then(() => logger.info(`Completed authorization of ${snapUrl}`))
+    .catch((error) => {
+      return prepareError(error)
```
bartaz: I hoped it would be not necessary to call `prepareError` outside of `sendError`, but as `completeSnapAuthorization` is called from different part of flow it doesn't actually use `sendError` and needs special treatment, right?

cjwatson: Indeed.  I imagine we'll come back to this as a target for further refactoring ...

PR #65, opened 2017-01-04
Properly showing error messages from APIs in repository input

bartaz:


src/common/components/repository-input/index.js
```diff
@@ -12,16 +12,18 @@ import { Form, InputField, Message } from '../forms';
...
+      message = webhook.error.message;
     } else if (input.error) {
-      if (input.repository) {
-        message = `Repository ${input.repository} doesn't exist, is not public or doesn't contain snapcraft.yaml file.`;
-      } else {
-        message = input.error.message;
-      }
+      message = input.error.message;
+    } else {
+      message = 'Unexpected error. Please make sure you are entering valid GitHub repository and try again.';
```
evandandrea: "Please make sure you are entering a valid GitHub repository…"

test/unit/src/common/actions/t_repository-input.js
```diff
@@ -54,9 +54,9 @@ describe('repository input actions', () => {
       expect(store.getActions()).toInclude(expectedAction);
     });
 
-    it('should create a valid flux standard action', () => {
-      expect(isFSA(action)).toBe(true);
-    });
+    // it('should create a valid flux standard action', () => {
+    //   expect(isFSA(action)).toBe(true);
+    // });
```
cjwatson: Include a comment about why this test is disabled?  (And maybe use `xit` or similar instead.)

test/unit/src/common/components/repository-input/t_repository-input.js
```diff
@@ -1,14 +1,14 @@
...
-
+import configureMockStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
 // Workaround for Redux-wrapped components
 // See: https://github.com/airbnb/enzyme/issues/472
 import Component from '../../../../../../src/common/components/repository-input';
 const RepositoryInput = Component.WrappedComponent;
 
-let dispatch = expect.createSpy();
+//let dispatch = expect.createSpy();
```
cjwatson: A fair bit of commented-out code in this file; I'd like to see that either explained or removed.

bartaz: Good catch.

`setGitHubRepository` that used to just return action object now returns function that dispatches 2 other actions... But it may all be irrelevant now that webhook errors are displayed on `/setup` view rather than in repository input.
https://github.com/canonical-ols/build.snapcraft.io/pull/65/files#diff-890863cbc28566af9b5f84d509b1d749R19

PR #66, opened 2017-01-04
Create webhook after creating snap

cjwatson:
We should only install a webhook on the repository once we have a snap
for it to operate on.  This also reduces rollback complexity in the case
where something goes wrong while creating the snap.

I added a new client-side /:account/:repo/setup route that deals with
setting up the repository (currently only installing the webhook) and
then redirects to /:account/:repo/builds.  This gives us somewhere we
can run repository setup code after emerging from SSO authorisation, and
seems a better idea than doing it directly on /:account/:repo/builds.

### DIFF
--- a/src/common/containers/repository-setup.js
+++ b/src/common/containers/repository-setup.js
@@ -22,7 +22,7 @@ class RepositorySetup extends Component {
     const { fullName, isFetching, success, error } = this.props;
 
     if (success) {
-      this.props.router.push(`/${fullName}/builds`);
+      this.props.router.replace(`/${fullName}/builds`);
     } else {
       return (
         <div className={styles.container}>
--- a/src/common/containers/repository-setup.js
+++ b/src/common/containers/repository-setup.js
@@ -29,7 +29,6 @@ class RepositorySetup extends Component {
           <Helmet
             title={`Setting up ${fullName}`}
           />
-          <h1>Setting up {fullName}</h1>
           { isFetching &&
             <div className={styles.spinner}><Spinner /></div>
           }
--- a/src/common/containers/repository-setup.js
+++ b/src/common/containers/repository-setup.js
@@ -11,15 +11,15 @@ import styles from './container.css';
 
 class RepositorySetup extends Component {
   componentDidMount() {
-    const { account, repo, isPending } = this.props;
+    const { account, repo, isFetching } = this.props;
 
-    if (!isPending) {
+    if (!isFetching) {
       this.props.dispatch(createWebhook(account, repo));
     }
   }
 
   render() {
-    const { fullName, isPending, success, error } = this.props;
+    const { fullName, isFetching, success, error } = this.props;
 
     if (success) {
       this.props.router.push(`/${fullName}/builds`);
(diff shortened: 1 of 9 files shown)


src/common/reducers/webhook.js
```diff
@@ -12,12 +12,26 @@ const ERROR_MESSAGES = {
...
-export function webhook(state = INITIAL_STATE, code) {
-  switch (code.type) {
-    case WEBHOOK_FAILURE:
+export function webhook(state = INITIAL_STATE, action) {
+  switch (action.type) {
+    case ActionTypes.WEBHOOK:
       return {
         ...state,
-        error: { message:  ERROR_MESSAGES[code.code] }
+        isPending: true
```
bartaz: I believe it has the same meaning as `isFetching` that we use in other reducers, so I would rename it to `isFetching` too to keep it consistent.

src/common/containers/repository-setup.js
```diff
@@ -0,0 +1,75 @@
...
+import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
+
+import { createWebhook } from '../actions/webhook';
+import { Message } from '../components/forms';
+import Spinner from '../components/spinner';
+
+import styles from './container.css';
+
+class RepositorySetup extends Component {
```
bartaz: At first I was wondering if we need this new view at all, as maybe `createSnap` could dispatch `createWebhook` (just as `createWebhook` dispatches `createSnap` in current flow), but then I realised that creating snap goes through redirects, so having view separate from `/builds` is good choice even if ATM all it does is triggering webhook action and redirecting again to `/builds`.

src/common/containers/repository-setup.js
```diff
@@ -0,0 +1,75 @@
...
+
+    if (success) {
+      this.props.router.push(`/${fullName}/builds`);
+    } else {
+      return (
+        <div className={styles.container}>
+          <Helmet
+            title={`Setting up ${fullName}`}
+          />
+          <h1>Setting up {fullName}</h1>
```
bartaz: This view is only visible for very short time (while webhook is being set up), so it maybe hard to even notice this title. But it's quite visible that it shows up and disappears again to be replaced by `/builds` view.

So I think I would remove the title and only keep loading spinner there, so the 'flash' of content is not so visible.

src/common/containers/repository-setup.js
```diff
@@ -0,0 +1,75 @@
...
+    if (!isPending) {
+      this.props.dispatch(createWebhook(account, repo));
+    }
+  }
+
+  render() {
+    const { fullName, isPending, success, error } = this.props;
+
+    if (success) {
+      this.props.router.push(`/${fullName}/builds`);
```
bartaz: I guess `/setup` is kind of 'internal' route and we probably don't want to keep it in history (so users don't navigate to it directly), so I would use [`.replace`](https://github.com/ReactTraining/react-router/blob/master/docs/API.md#replacepathorloc) instead of `.push` here.

src/common/containers/repository-setup.js
cjwatson: Indeed, that was pretty much my reasoning.  The setup page will shortly also start triggering initial builds.

cjwatson:
Thanks; I've addressed all your comments.

PR #70, opened 2017-01-05
Request initial builds from repository setup page

cjwatson:
We can do this in parallel with installing the webhook, and it means
that we get a satisfying response from the builds page right away.

### DIFF
--- a/src/common/reducers/request-snap-builds.js
+++ b/src/common/reducers/request-snap-builds.js
@@ -1,0 +1,34 @@
+import * as ActionTypes from '../actions/request-snap-builds';
+import { snapBuildFromAPI } from '../helpers/snap-builds';
+
+export function requestSnapBuilds(state = {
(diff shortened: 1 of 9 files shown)


src/common/reducers/request-snap-builds.js
```diff
@@ -0,0 +1,34 @@
+import * as ActionTypes from '../actions/request-snap-builds';
+import { snapBuildFromAPI } from '../helpers/snap-builds';
+
+export function requestSnapBuilds(state = {
+  isFetching: false,
+  builds: [],
```
bartaz: There seems to be a bit of redundancy, as we already request and keep builds list in `snapBuilds` reducer (for data fetched for builds list).

As discussed on IRC, let's think about refactoring it later.

cjwatson: Yep - filed as #71.

PR #72, opened 2017-01-05
Unset proxy when running unit tests

cjwatson:
Our Jenkins environment sets a proxy, and that shouldn't be used for
tests because it isn't going to be able to contact mock services.  This
was already done for the route tests, but now we have unit tests that
exercise some relevant code paths.


cjwatson:
I think if we ever wanted that then we would be mistaken and should be corrected :-)  (For the usual reasons - offline-safety, isolation from external resources vanishing, and so on.)

PR #74, opened 2017-01-05
Verify if snap name is registered before requesting new snap

bartaz:
screen shot 2017-01-05 at 14 25 31


environments/development.env
```diff
@@ -6,6 +6,7 @@ NODE_ENV=development
 UBUNTU_SSO_URL=https://login.ubuntu.com
 OPENID_VERIFY_URL=http://localhost:8000/login/verify
 LP_API_URL=https://api.launchpad.net
+STORE_MACAROON_API_URL=https://myapps.developer.ubuntu.com/dev/api
```
bartaz: As we are talking to production LP only, I guess it's fine to talk to production store to check name registration.

src/server/handlers/launchpad.js
```diff
@@ -210,6 +218,47 @@ const getSnapcraftYaml = (owner, name, token) => {
...
+  }).then((response) => response.json().then((json) => {
+    if (response.status === 200 && json.macaroon) {
+      return name;
+    } else {
+      throw new PreparedError(400, RESPONSE_SNAP_NAME_NOT_REGISTERED);
+    }
+  }));
+};
+
+const requestNewSnap = (name, repositoryUrl) => {
```
bartaz: Extracted it from body of `newSnap` for easier promise chaining.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -74,113 +74,163 @@ describe('The Launchpad API endpoint', () => {
       });
     });
 
-    context('when snap does not exist', () => {
-      const caveatId = 'dummy caveat';
+    context('when repo contains valid snapcraft.yaml', () => {
```
bartaz: Quite a lot of changes here due to moving tests around.

I moved some tests that shared same mocked requests (valid snapcraft.yaml, name registered) into common nested contexts to avoid duplicating same calls to `nock`.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -74,113 +74,163 @@ describe('The Launchpad API endpoint', () => {
...
+      const snapName = 'dummy-test-snap';
 
       beforeEach(() => {
         nock(conf.get('GITHUB_API_ENDPOINT'))
           .get('/repos/anaccount/arepo')
           .reply(200, { permissions: { admin: true } });
         nock(conf.get('GITHUB_API_ENDPOINT'))
           .get('/repos/anaccount/arepo/contents/snapcraft.yaml')
-          .reply(200, 'name: test-snap\n');
```
bartaz: Fun fact: the tests were actually passing with `test-snap` (even though macaroon requests were not mocked), because `test-snap` seems to be actually a real registered name in public store, so when tests called real API they got valid macaroon response.

docs/environment-variables.md
```diff
@@ -67,6 +67,15 @@ The OAuth access token, generated by Launchpad.
...
 The OAuth token secret, generated by Launchpad.
 
+
+## Ubuntu Store settings
+
+### STORE\_MACAROON\_API\_URL
+
+- **Example**: `https://myapps.developer.ubuntu.com/dev/api`
+
+The URL of the Ubuntu Store Macaroon API service.
```
cjwatson: I'd call this just `STORE_API_URL` and "the Ubuntu Store API service", I think; there's more than just the macaroon API under there, and we might want some other bit of it in future.

environments/development.env
cjwatson: Indeed - you have no choice.

src/server/handlers/launchpad.js
```diff
@@ -210,6 +218,47 @@ const getSnapcraftYaml = (owner, name, token) => {
     });
 };
 
+const verifySnapNameRegistered = (name) => {
+  return fetch(`${conf.get('STORE_MACAROON_API_URL')}/acl/`, {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
```
cjwatson: I can't remember if `fetch` does this by itself, but we should make sure to pass `Accept: application/json` as well, even if the store doesn't enforce that today.

test/routes/src/server/routes/t_launchpad.js
cjwatson: This can wait until later, but I wonder if we should have some global test setup that puts some kind of catch-all request mock in place so that requests don't slip through to production services if we forget something.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -74,113 +74,163 @@ describe('The Launchpad API endpoint', () => {
...
-          .post('/launchpad/snaps')
-          .send({ repository_url: 'https://github.com/anaccount/arepo' })
-          .expect(hasStatus('success'))
-          .end(done);
-      });
+        beforeEach(() => {
+          nock(conf.get('STORE_MACAROON_API_URL'))
+            .post('/acl/', {
+              'packages': [{ 'name': snapName, 'series': '16' }],
+              'permissions':['package_upload']
```
cjwatson: Space after colon?  (Similarly below.)

src/server/handlers/launchpad.js
```diff
@@ -83,6 +83,14 @@ const RESPONSE_SNAPCRAFT_YAML_NO_NAME = {
   }
 };
 
+const RESPONSE_SNAP_NAME_NOT_REGISTERED = {
+  status: 'error',
+  payload: {
+    code: 'snap-name-not-registered',
+    message: 'Snap name is not registered in the store'
```
evandandrea: It could wait until later, but would it be quick to change this to "The name provided in the snapcraft.yaml ($name) is not registered in the store. Please register it \<a href="https://myapps.developer.ubuntu.com/dev/click-apps/register-name/?name=$name">here</a> before trying again."

cjwatson: I'd suggest that we should do this by handling that payload code specially in the frontend component, rather than by injecting HTML from the server-side handler.

bartaz: Yes, it should be handled in views, as HTML in such messages would be escaped by React.

Sure, ideally we want actionable error message like this, but for MVP demo this should be enough. And it would be great to have some designs if we start to look into details like this.

cjwatson: I'm going to do this in a separate PR, rather than adding it to Bartek's.

PR #77, opened 2017-01-06
Link to store for name registration

cjwatson:
Requested by Evan Dandrea.

The link CSS is from vanilla-framework, but limited to descendants of
errorMsg in order to be as unintrusive as possible.  Without something
like this it's hard to tell that the link is present, because the error
colour is too similar to the link colour.


cjwatson:
Self-approving since I'm on my own today and would like to get this in place in time for next week's demo.

PR #78, opened 2017-01-06
Tidy up duration/date display on builds page

cjwatson:
duration and dateStarted may be null, so just leave the corresponding
cells empty in that case rather than having it come out as "a few
seconds" and "Invalid date" respectively.

While I'm here, render dates using ISO 8601 order in the dateStarted
tooltip.


cjwatson:
Self-approving as a simple but rather visible fix for next week's demo.

PR #82, opened 2017-01-09
Various logging improvements

cjwatson:
Right now, we only log to the console.  This is fine for development, but cumbersome for the staging/production deployments, where the only way to debug anything is to log into the unit and use `systemctl status --lines=1000 snap-build`; this is made worse by switch deploys, since the unit that contains the logging information is likely to go away.

Part of fixing this will be sending logs to ELK, which is outside the scope of this PR since it'll be done by the Mojo spec.  However, as a prerequisite for that, we need to log to files.  This PR arranges to write logs to files under `LOGS_PATH` (already set by the charm), and tries to reserve the `express-error` logger only for actual errors.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -865,6 +865,9 @@
         }
       }
     },
+    "winston-daily-rotate-file": {
+      "version": "1.4.2"
+    },
     "wrap-ansi": {
       "version": "2.1.0"
     },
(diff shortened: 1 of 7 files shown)


cjwatson:
I've refactored this to use `logger.error()` rather than the separate `express-error` logger; there's no need for these to actually go to separate log files, as long as the levels are distinct.  Does this still look OK?

bartaz:
@cjwatson Sure it's OK - as long as it gives you output you want in log files. Looks much better with `logger.error`. Thanks!

PR #83, opened 2017-01-09
Common test setup

### LINKED ISSUE #64: Refactor common test config overrides
We set up some common config overrides in various tests, notably `LP_API_URL` et al, and it's a bit repetitive.  We should look into refactoring this into either some helper functions or test setup code.

bartaz:
Fixes #64

Moved tests setup to one common file (shared between unit and route tests) and test config into test.env file that is passed to tests when mocha runs.


environments/test.env
```diff
@@ -0,0 +1,26 @@
+NODE_ENV=test
+HTTP_PROXY=
```
bartaz: This should be enough to set it here instead of whole command.
It should be set in test setup (so before any of our code runs in tests)

environments/test.env
```diff
@@ -0,0 +1,26 @@
...
+HTTPS_PROXY=
+
+BASE_URL=http://localhost:8000
+
+UBUNTU_SSO_URL=https://login.ubuntu.com
+OPENID_VERIFY_URL=http://localhost:8000/login/verify
+
+LP_API_URL=http://localhost:4000/launchpad
```
bartaz: All config that was previously set in tests can is now here.

PR #87, opened 2017-01-10
Using only one reducer to store snap builds

### LINKED ISSUE #71: Share builds data between snapBuilds and requestSnapBuilds
`requestSnapBuilds` (from #70) could save a request when `/:account/:repo/builds` is first loaded by storing the requested builds in the same bit of state as `snapBuilds`.

bartaz:
Fixes #71 

Reducer for request builds actions was exactly the same as the one used to fetch builds for given snap, so we simply merge them to one to avoid unnecessary duplication.


src/common/containers/repository-setup.js
```diff
@@ -4,13 +4,21 @@ import { connect } from 'react-redux';
...
 import { createWebhook } from '../actions/webhook';
-import { requestBuilds } from '../actions/request-snap-builds';
+import { requestBuilds } from '../actions/snap-builds';
 import { Message } from '../components/forms';
 import Spinner from '../components/spinner';
 
 import styles from './container.css';
 
 class RepositorySetup extends Component {
+  componentWillReceiveProps(nextProps) {
```
bartaz: Moved redirecting out of `render` to make sure `render` always returns valid component and to trigger redirect even before anything is rendered.

PR #90, opened 2017-01-13
Only fetch snap by repo URL when needed

bartaz:
Currently we fetch snap by repo URL every time we fetch builds (every 15s on builds page). It's unnecessarily doubles numbers of requests made from the client.

With this update snap link is saved to the store and only list of builds is being fetched every 15 seconds to keep it up to date.

### DIFF
--- a/src/common/actions/snap-builds.js
+++ b/src/common/actions/snap-builds.js
@@ -6,6 +6,7 @@ import getGitHubRepoUrl from '../helpers/github-url';
 const BASE_URL = conf.get('BASE_URL');
 
 export const FETCH_BUILDS = 'FETCH_BUILDS';
+export const FETCH_SNAP_SUCCESS = 'FETCH_SNAP_SUCCESS';
 export const FETCH_BUILDS_SUCCESS = 'FETCH_BUILDS_SUCCESS';
 export const FETCH_BUILDS_ERROR = 'FETCH_BUILDS_ERROR';
 
@@ -59,7 +60,13 @@ export function fetchSnap(repository) {
       return fetch(url)
         .then(checkStatus)
         .then(response => response.json())
-        .then((json) => dispatch(fetchBuilds(json.payload.message)))
+        .then((json) => {
+          const snapLink = json.payload.message;
+          dispatch({
+            type: FETCH_SNAP_SUCCESS,
+            payload: snapLink
+          });
+        })
         .catch((error) => dispatch(fetchBuildsError(error)));
     }
   };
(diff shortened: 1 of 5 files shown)


src/common/actions/snap-builds.js
```diff
@@ -59,7 +60,13 @@ export function fetchSnap(repository) {
       return fetch(url)
         .then(checkStatus)
         .then(response => response.json())
-        .then((json) => dispatch(fetchBuilds(json.payload.message)))
```
bartaz: `fetchSnap` is not starting fetching builds anymore, just triggers an action to   save snap link in the store.

`fetchBuilds` is called directly from builds container once snap link is in the store

src/common/containers/builds.js
```diff
@@ -6,36 +6,52 @@ import BuildHistory from '../components/build-history';
...
 
-import { fetchSnap } from '../actions/snap-builds';
+import { fetchBuilds, fetchSnap } from '../actions/snap-builds';
 
 import styles from './container.css';
 
 class Builds extends Component {
   fetchInterval = null
 
+  fetchData({ snapLink, repoFullName }) {
```
bartaz: If snap link is not available yet, fetch snap for given repo.
If we have the snap, fetch the builds.

test/unit/src/common/reducers/t_snap-builds.js
```diff
@@ -72,93 +73,83 @@ describe('snapBuilds reducers', () => {
...
-        ...initialState,
-        error: 'Previous error'
-      };
-
-      const action = {
-        type: ActionTypes.FETCH_BUILDS_SUCCESS,
-        payload: SNAP_ENTRIES
-      };
-
       expect(snapBuilds(state, action).error).toBe(null);
```
cjwatson: The refactoring is incorrect here, rendering the test ineffective.  The initial state for this test must have `error` set.

bartaz: Well spotted! Thanks

test/unit/src/common/reducers/t_snap-builds.js
```diff
@@ -72,93 +73,84 @@ describe('snapBuilds reducers', () => {
...
+    it('should store snap link', () => {
+      expect(snapBuilds(state, action).snapLink).toEqual(SNAP_LINK);
+    });
+  });
+
+  context('FETCH_BUILDS_SUCCESS', () => {
+    const state = {
+      ...initialState,
+      isFetching: true,
+      error: 'Previous error'
```
bartaz: Added `error` to also cover "it should clean error" test case as mentioned in review comments.

PR #91, opened 2017-01-17
GH repository (owner, name, etc) naming consistency

bartaz:
More consistent naming of GH repo related values.

Currently for example repository `canonical-ols/build.snapcraft.io` we have:

```js
{
  owner: 'canonical-ols',
  name: 'build.snapcraft.io',
  fullName 'canonical-ols/build.snapcraft.io',
  url: 'https://github.com/canonical-ols/buildsnapcraft.io'
}
```

This PR is mostly a lot of renaming. Some other points of interest where commented separately inline (see comments below).

### DIFF
--- a/src/common/actions/repository-input.js
+++ b/src/common/actions/repository-input.js
@@ -1,7 +1,7 @@
 import 'isomorphic-fetch';
 
 import conf from '../helpers/config';
-import getGitHubRepoUrl from '../helpers/github-url';
+import { parseGitHubRepoUrl } from '../helpers/github-url';
 
 const BASE_URL = conf.get('BASE_URL');
 
@@ -34,17 +34,14 @@ function checkStatus(response) {
   }
 }
 
-// TODO: bartaz (accept repository object)?
-export function createSnap(repository, location) {
-  const fullName = repository;
-
+export function createSnap(repositoryUrl, location) { // location for tests
   return (dispatch) => {
-    if (fullName) {
+    if (repositoryUrl) {
+      const { fullName } = parseGitHubRepoUrl(repositoryUrl);
+
       dispatch({
-        type: CREATE_SNAP,
-        payload: fullName
+        type: CREATE_SNAP
       });
-      const repositoryUrl = getGitHubRepoUrl(fullName);
 
       return fetch(`${BASE_URL}/api/launchpad/snaps`, {
         method: 'POST',
--- a/src/common/actions/repository-input.js
+++ b/src/common/actions/repository-input.js
@@ -4,12 +4,8 @@ import conf from '../helpers/config';
 import getGitHubRepoUrl from '../helpers/github-url';
 
 const BASE_URL = conf.get('BASE_URL');
-const GITHUB_API_ENDPOINT = conf.get('GITHUB_API_ENDPOINT');
 
 export const SET_GITHUB_REPOSITORY = 'SET_GITHUB_REPOSITORY';
-export const VERIFY_GITHUB_REPOSITORY = 'VERIFY_GITHUB_REPOSITORY';
-export const VERIFY_GITHUB_REPOSITORY_SUCCESS = 'VERIFY_GITHUB_REPOSITORY_SUCCESS';
-export const VERIFY_GITHUB_REPOSITORY_ERROR = 'VERIFY_GITHUB_REPOSITORY_ERROR';
 export const CREATE_SNAP = 'CREATE_SNAP';
 export const CREATE_SNAP_ERROR = 'CREATE_SNAP_ERROR';
 
@@ -36,38 +32,6 @@ function checkStatus(response) {
       throw getError(response, json);
     });
   }
-}
-
-export function verifyGitHubRepository(repository) {
-  const fullName = repository;
-  return (dispatch) => {
-    if (fullName) {
-      dispatch({
-        type: VERIFY_GITHUB_REPOSITORY,
-        payload: fullName
-      });
-
-      return fetch(`${GITHUB_API_ENDPOINT}/repos/${fullName}/contents/snapcraft.yaml`)
-        .then(checkStatus)
-        .then(() => dispatch(verifyGitHubRepositorySuccess(getGitHubRepoUrl(fullName))))
-        .catch(error => dispatch(verifyGitHubRepositoryError(error)));
-    }
-  };
-}
-
-export function verifyGitHubRepositorySuccess(repositoryUrl) {
-  return {
-    type: VERIFY_GITHUB_REPOSITORY_SUCCESS,
-    payload: repositoryUrl
-  };
-}
-
-export function verifyGitHubRepositoryError(error) {
-  return {
-    type: VERIFY_GITHUB_REPOSITORY_ERROR,
-    payload: error,
-    error: true
-  };
 }
 
 // TODO: bartaz (accept repository object)?
--- a/src/common/actions/repository-input.js
+++ b/src/common/actions/repository-input.js
@@ -39,16 +39,17 @@ function checkStatus(response) {
 }
 
 export function verifyGitHubRepository(repository) {
+  const fullName = repository;
   return (dispatch) => {
-    if (repository) {
+    if (fullName) {
       dispatch({
         type: VERIFY_GITHUB_REPOSITORY,
-        payload: repository
+        payload: fullName
       });
 
-      return fetch(`${GITHUB_API_ENDPOINT}/repos/${repository}/contents/snapcraft.yaml`)
+      return fetch(`${GITHUB_API_ENDPOINT}/repos/${fullName}/contents/snapcraft.yaml`)
         .then(checkStatus)
-        .then(() => dispatch(verifyGitHubRepositorySuccess(getGitHubRepoUrl(repository))))
+        .then(() => dispatch(verifyGitHubRepositorySuccess(getGitHubRepoUrl(fullName))))
         .catch(error => dispatch(verifyGitHubRepositoryError(error)));
     }
   };
@@ -69,14 +70,17 @@ export function verifyGitHubRepositoryError(error) {
   };
 }
 
+// TODO: bartaz (accept repository object)?
 export function createSnap(repository, location) {
+  const fullName = repository;
+
   return (dispatch) => {
-    if (repository) {
+    if (fullName) {
       dispatch({
         type: CREATE_SNAP,
-        payload: repository
+        payload: fullName
       });
-      const repositoryUrl = getGitHubRepoUrl(repository);
+      const repositoryUrl = getGitHubRepoUrl(fullName);
 
       return fetch(`${BASE_URL}/api/launchpad/snaps`, {
         method: 'POST',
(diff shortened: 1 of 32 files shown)


src/common/actions/repository-input.js
```diff
@@ -38,45 +34,14 @@ function checkStatus(response) {
   }
 }
 
-export function verifyGitHubRepository(repository) {
```
bartaz: `verifyGitHubRepository` was not used anymore, we do this on server-side when trying to create a snap.

src/common/containers/build-details.js
```diff
@@ -5,18 +5,37 @@ import Helmet from 'react-helmet';
...
+    if (snapLink) {
+      this.props.dispatch(fetchBuilds(snapLink));
+    } else {
+      this.props.dispatch(fetchSnap(repository.url));
+    }
+  }
+
   componentWillMount() {
-    this.props.dispatch(fetchSnap(this.props.fullName));
+    this.fetchData(this.props);
```
bartaz: Properly fetching snap and build data when user enters build details directly via URL.

src/common/containers/build-details.js
```diff
@@ -43,33 +62,44 @@ class BuildDetails extends Component {
...
   dispatch: PropTypes.func.isRequired
 };
 
 const mapStateToProps = (state, ownProps) => {
-  const account = ownProps.params.account.toLowerCase();
-  const repo = ownProps.params.repo.toLowerCase();
+  const owner = ownProps.params.owner.toLowerCase();
+  const name = ownProps.params.name.toLowerCase();
+  const fullName = `${owner}/${name}`;
+  const url = getGitHubRepoUrl(fullName);
```
bartaz: Because we redirect to `/owner/name/builds` after SSO login or user can enter this URL directly, we don't have repository info in the store (from repository input), so it's constructed here based on `owner` and `name` from URL params.

I wonder if we could reuse functionality of reducer (that already does the same thing for repository input). But that would mean that before we can render builds details (or builds listing) we need to trigger an action to populate the store. Who should trigger this action? Component in `componentWillMount` ?

src/common/containers/builds.js
```diff
@@ -77,9 +81,10 @@ Builds.propTypes = {
 };
 
 const mapStateToProps = (state, ownProps) => {
-  const account = ownProps.params.account.toLowerCase();
-  const repo = ownProps.params.repo.toLowerCase();
-  const repoFullName = `${account}/${repo}`;
+  const owner = ownProps.params.owner.toLowerCase();
+  const name = ownProps.params.name.toLowerCase();
+  const fullName = `${owner}/${name}`;
+  const url = getGitHubRepoUrl(fullName);
```
bartaz: Same question as in build details https://github.com/canonical-ols/build.snapcraft.io/pull/91/files#r96604386

src/common/containers/repository-setup.js
```diff
@@ -73,24 +77,20 @@ RepositorySetup.propTypes = {
 };
 
 const mapStateToProps = (state, ownProps) => {
-  const account = ownProps.params.account.toLowerCase();
-  const repo = ownProps.params.repo.toLowerCase();
-  const fullName = `${account}/${repo}`;
+  const owner = ownProps.params.owner.toLowerCase();
+  const name = ownProps.params.name.toLowerCase();
+  const fullName = `${owner}/${name}`;
+  const url = getGitHubRepoUrl(fullName);
```
bartaz: Third place where we repeat the same as in build details and builds listing (see comment https://github.com/canonical-ols/build.snapcraft.io/pull/91/files#r96604386)

src/common/helpers/github-url.js
```diff
@@ -1,6 +1,16 @@
-const getGitHubRepoUrl = (owner, name) => {
+import parseGitHubUrl from 'parse-github-url';
+
+export const getGitHubRepoUrl = (owner, name) => {
   let repository = !name ? owner : `${owner}/${name}`;
   return `https://github.com/${repository}`;
 };
 
-export default getGitHubRepoUrl;
+export const parseGitHubRepoUrl = (url) => {
```
bartaz: Moved out from repository input reducer.

Now instead of just returning full repo name it returns whole object with owner, name, fullName and url, so in any place we need those we can get these in same way.

src/common/reducers/repository-input.js
```diff
@@ -19,39 +13,15 @@ export function repositoryInput(state = {
       return {
         ...state,
         inputValue: action.payload,
-        repository: parseRepository(action.payload),
+        repository: parseGitHubRepoUrl(action.payload),
```
bartaz: Whole repository object (owner, name, fullName, url) is kept here.

I wonder if it would make sense to move repository info into separate small reducer that would handle this (instead of nested object).
That would also make repository info more independent from repository input component actions and could be reused elsewhere.

src/common/containers/build-details.js
katiefenn: > Who should trigger this action? Component in componentWillMount ?
Yes. This is how we loaded pre-requisite data like user name, balance, payment methods etc on the last React App I worked on.

src/common/helpers/github-url.js
```diff
@@ -1,6 +1,16 @@
...
+  const repo = parseGitHubUrl(url);
+  return (repo && repo.repo) ? {
+    owner: repo.owner,
+    name: repo.name,
+    fullName: repo.repo,
+    url: url
+  } : null;
```
katiefenn: If our repository naming scheme is as close as it is to parse-github-url's, is it worth just using their naming scheme instead?

src/common/reducers/repository-input.js
katiefenn: I think this is fine. On the last React app we worked on, forms were a common part of our stores, and recurring properties like "account number", "amount", "credit card number" and "password" appeared in quite a few places. We extracted common code (usually functions for validating, transforming and decorating data) into helper functions as much as possible. Standardising the shape of the data made this easier.

katiefenn: I think we should **both** create a reducer for repository information and retain repository information inside the repository input form reducer.

A repository info reducer would be good for pages where the repository is an important context - the build details page, for example.

The repository information in the repository input form reducer is responsible for storing the state of the form field. I think the distinction is important, because the form data can be empty, invalid or incomplete. The data describes something which _might be_ a repository, whereas a repository info reducer would represent a _concrete_ repository.

This was a common occurrence on the last app I worked on. We had a "store" (a reducer in Redux terms) for payment methods when they were important for the page being viewed (e.g. making a payment with a saved credit card). We had a separate "store" for the new payment method form. Both used a standardised shape for the payment methods, and we extracted common methods for validating, transforming and decorating credit card data into shared helper functions.

src/common/helpers/github-url.js
bartaz: Personally I find naming repository full name 'repo' a bit confusing. Especially when nested in some "repository" object, so I prefer to stick to naming used by GH in their API.

src/common/containers/build-details.js
```diff
@@ -5,18 +5,46 @@ import Helmet from 'react-helmet';
...
+      this.props.dispatch(fetchBuilds(snapLink));
+    } else if (repository) {
+      this.props.dispatch(fetchSnap(repository.url));
+    }
+  }
+
   componentWillMount() {
-    this.props.dispatch(fetchSnap(this.props.fullName));
+    if (!this.props.repository && this.props.fullName) {
+      this.props.dispatch(setGitHubRepository(this.props.fullName));
```
bartaz: To properly propagate `repository` reducer this bit is needed (and duplicated in repository setup, builds list and build details). 

I wonder if it would be worth having additional layer of container/decorator that would trigger the action (to propagate data in reducer) and pass it to component that needs it.

But it sounds a little bit over-engineered...

earnubs: Higher Order Component? https://facebook.github.io/react/docs/higher-order-components.html

src/common/containers/builds.js
```diff
@@ -6,22 +6,27 @@ import BuildHistory from '../components/build-history';
...
       this.props.dispatch(fetchBuilds(snapLink));
-    } else {
-      this.props.dispatch(fetchSnap(repoFullName));
+    } else if (repository) {
+      this.props.dispatch(fetchSnap(repository.url));
     }
   }
 
-  componentWillMount() {
+  componentDidMount() {
```
bartaz: As a fix for #96 replacing `componentWillMount` with `componentDidMount` to avoid calling it on server.

src/common/containers/with-repository.js
```diff
@@ -0,0 +1,50 @@
...
+      }
+    }
+
+    componentWillUnmount() {
+
+    }
+
+    render() {
+      const { fullName, ...passThroughProps } = this.props;
+      fullName; // just to 'use' this variable
```
bartaz: That's a lousy trick: I want to pass all the props except fullName to WrappedComponent, but I don't need fullName anywhere, so eslint complains about unused variable.

Or am I just tired and there is some obvious better way to do this?

src/common/containers/with-repository.js
```diff
@@ -0,0 +1,46 @@
+import React, { Component, PropTypes } from 'react';
+
+import { setGitHubRepository } from '../actions/repository-input';
+
+function withRepository(WrappedComponent) {
```
bartaz: New HOC component that contains bit of functionality to propagate `repository` reducer with data (needed for wrapped component).

It requires `fullName` prop (in most cases it's filled from URL params) and `repository` that is taken from redux store.

I'm not sure if it's the best way to implement this (it still requires `connect` layer on top to provide `fullName` and `repository`), but it seems to do the trick (removed duplicated code from repository setup, builds list and build details containers).

src/common/containers/repository-setup.js
```diff
@@ -73,25 +77,16 @@ RepositorySetup.propTypes = {
...
-      error: state.snapBuilds.error
-    }
+    repository: state.repository,
+    webhook: state.webhook,
+    builds: state.snapBuilds
   };
 };
 
-export default connect(mapStateToProps)(withRouter(RepositorySetup));
+export default connect(mapStateToProps)(withRepository(withRouter(RepositorySetup)));
```
bartaz: It's getting bit crazy layered in here...

src/common/containers/with-repository.js
earnubs: // eslint-disable-line no-unused-vars

katiefenn: What's the worst that could happen if fullName was passed?

src/common/containers/repository-setup.js
```diff
@@ -5,39 +5,40 @@ import { withRouter } from 'react-router';
```
katiefenn: Non-blocking comment:

I wonder whether this component needs to exist. If the repository information in the store is complete enough for a webhook and a build to be created, these actions should be initiated when the **createSnap** action successfully resolves. The spinner could also be moved to the RepositoryInput component.

Further, we should probably look at creating an API endpoint to aggregate the creation of webhooks and builds.

bartaz: `createSnap` goes through SSO redirect and we want to create webhooks and request builds after this happens (to be sure whole process succeeded). That's why /owner/name/setup route is needed to finalise setting up repository before redirecting to builds view.

src/common/containers/with-repository.js
bartaz: @katiefenn You mean passed to the wrapped component?

In dev env React will likely complain on receiving prop that is not defined in PropTypes.
In prod probably nothing.

I disabled the check for this line as @earnubs suggested.

PR #95, opened 2017-01-18
58: Refactor configuration

katiefenn:
- nconf removed due to unavoidable dependency on the Node fs API
- Rewritten config helper exposing an immutable config object
- Added "baked in" config to JS bundle, using WebPack.DefinePlugin to
  replace process.env
- Added environment variables to Makefile
- Removed runtime config from HTML helper component

### DIFF
--- a/mocks/service/index.js
+++ b/mocks/service/index.js
@@ -1,3 +1,4 @@
 require('babel-polyfill');
 require('babel-register');
+require('babel-polyfill');
 require('./server');
--- a/mocks/service/index.js
+++ b/mocks/service/index.js
@@ -1,4 +1,3 @@
 require('babel-polyfill');
 require('babel-register');
-require('babel-polyfill');
 require('./server');
--- a/package.json
+++ b/package.json
@@ -41,12 +41,11 @@
     "express-winston": "^2.0.0",
     "helmet": "^3.1.0",
     "images-require-hook": "^1.0.3",
+    "immutable": "^3.8.1",
     "isomorphic-fetch": "^2.2.1",
     "js-yaml": "^3.7.0",
     "memcached": "^2.2.2",
     "moment": "^2.17.1",
-    "nconf": "^0.8.4",
-    "nconf-yaml": "^1.0.2",
     "normalize.css": "^5.0.0",
     "oauth": "^0.9.14",
     "openid": "^2.0.6",
@@ -64,7 +63,8 @@
     "redux-thunk": "^2.1.0",
     "request": "^2.79.0",
     "winston": "^2.3.0",
-    "winston-daily-rotate-file": "^1.4.2"
+    "winston-daily-rotate-file": "^1.4.2",
+    "yargs": "^6.4.0"
   },
   "devDependencies": {
     "assets-webpack-plugin": "^3.5.0",
--- a/src/common/helpers/config.js
+++ b/src/common/helpers/config.js
@@ -1,40 +1,12 @@
-export class Config {
-  constructor(config) {
-    // https://github.com/purposeindustries/window-or-global/blob/master/lib/index.js
-    // Establish the root object, `window` in the browser, or `global` on the server.
-    let root = typeof self == 'object' && self.Object == Object && self;
-    // root is window, we expect it to have config
-    if (root && !root.__CONFIG__) {
-      throw new ReferenceError('Client config: no config defined.');
-    }
-    // root is not window; test for node global and assign to avoid ref error
-    // we don't expect config here, this is client side only, but we can set it
-    // for unit tests
-    root = root || (typeof global == 'object' && global.Object == Object && global);
+import { Map } from 'immutable';
 
-    this._store = config || root.__CONFIG__;
-  }
+export const conf = Map(getGlobal().__CONFIG__);
 
-  get(key) {
-    let target = this._store;
-    const path = this._path(key);
-
-    while (path.length > 0) {
-      key = path.shift();
-
-      if (target && target.hasOwnProperty(key)) {
-        target = target[key];
-        continue;
-      }
-      return undefined;
-    }
-
-    return target;
-  }
-
-  _path(key) {
-    return (key == null) ? [] : key.split(':');
-  }
+function getGlobal() {
+  // window-or-global helper:
+  // returns global in Node.JS, returns window in browser
+  // https://github.com/purposeindustries/window-or-global/blob/master/lib/index.js
+  return (typeof self === 'object' && self.self === self && self) ||
+    (typeof global === 'object' && global.global === global && global) ||
+    this;
 }
-
-export default new Config();
--- a/src/config/defaults.js
+++ b/src/config/defaults.js
@@ -1,0 +1,19 @@
+export default {
+  HOST: '0.0.0.0',
+  PORT: '8000',
+  BASE_URL: 'http://localhost:8000',
+  WEBPACK_DEV_URL: 'http://localhost:8001',
+  NODE_ENV: 'development',
+  UBUNTU_SSO_URL: 'https://login.ubuntu.com',
+  OPENID_VERIFY_URL: 'http://localhost:8000/login/verify',
+  LP_API_URL: 'https://api.launchpad.net',
+  STORE_API_URL: 'https://myapps.developer.ubuntu.com/dev/api',
+  STORE_DEVPORTAL_URL: 'https://myapps.developer.ubuntu.com/dev',
+  GITHUB_API_ENDPOINT: 'http://localhost:4000/github',
+  GITHUB_AUTH_LOGIN_URL: 'http://localhost:4000/github/login/oauth/authorize',
+  GITHUB_AUTH_VERIFY_URL: 'http://localhost:4000/github/login/oauth/access_token',
+  GITHUB_AUTH_REDIRECT_URL: 'http://localhost:8000/auth/verify',
+  GITHUB_AUTH_CLIENT_ID: '389a7b4c2ade662c0bf4',
+  GITHUB_AUTH_CLIENT_SECRET: 'ea482b47a4830d38a838f48be43ab28002c33187',
+  GITHUB_WEBHOOK_SECRET: 'dummy-webhook-secret'
+};
--- a/src/server/helpers/config.js
+++ b/src/server/helpers/config.js
@@ -1,12 +1,13 @@
-import path from 'path';
-import nconf from 'nconf';
+import { Map } from 'immutable';
+import { argv } from 'yargs';
 import dotenv from 'dotenv';
 
-/**
- * Whitelist config items that are safe to send
- * to the client.
- * DO NOT ADD SENSITIVE DATA TO THIS LIST
- */
+// Whitelist for client-side config
+// Add settings to this list to use them in
+// the client.
+// ANY SETTINGS ADDED TO THIS LIST WILL BE
+// VISIBLE TO THE WORLD.
+//
 const CLIENT_SIDE_WHITELIST = [
   'NODE_ENV',
   'BASE_URL',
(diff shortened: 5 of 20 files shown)


Makefile
```diff
@@ -31,6 +31,19 @@ DIST = $(DISTDIR)/.done
...
+NODE_ENV = production
+GITHUB_API_ENDPOINT = https://api.github.com
+STORE_DEVPORTAL_URL = https://myapps.developer.ubuntu.com/dev
+
+target staging:
+	BASE_URL = https://build.staging.snapcraft.io
+
+target production:
+	BASE_URL = https://build.snapcraft.io
+
```
katiefenn: ☝️ This is my random, uninformed stab at setting the environment variables we need in Jenkins when we build the JS bundle. Some thoughts:

1) Is there a better place for these settings to "live"?
2) Have I used the correct syntax for setting variables and switching targets?
3) Do we already have staging and production targets set up on Jenkins?

cjwatson: This is quite unfortunate; it means that we won't be using exactly the same bundle that we tested on staging when we promote it to production.  Is there any way to avoid this while preserving the other benefits you're trying to gain?  I'd really strongly prefer the bundles to be identical, rather than testing one build and then hoping that nothing mysterious went wrong with the other similar build before promoting it to production; in other words, I view separate staging and production builds as an antifeature.

As for syntax, this doesn't look right; the variables would need to be exported, and the syntax for setting a target-specific variable looks more like `staging: export BASE_URL = https://build.staging.snapcraft.io`.  But of course this is all contingent on the above comment.

src/config/defaults.js
```diff
@@ -0,0 +1,19 @@
...
+  OPENID_VERIFY_URL: "http://localhost:8000/login/verify",
+  LP_API_URL: "https://api.launchpad.net",
+  STORE_API_URL: "https://myapps.developer.ubuntu.com/dev/api",
+  STORE_DEVPORTAL_URL: "https://myapps.developer.ubuntu.com/dev",
+  GITHUB_API_ENDPOINT: "http://localhost:4000/github",
+  GITHUB_AUTH_LOGIN_URL: "http://localhost:4000/github/login/oauth/authorize",
+  GITHUB_AUTH_VERIFY_URL: "http://localhost:4000/github/login/oauth/access_token",
+  GITHUB_AUTH_REDIRECT_URL: "http://localhost:8000/auth/verify",
+  GITHUB_AUTH_CLIENT_ID: "389a7b4c2ade662c0bf4",
+  GITHUB_AUTH_CLIENT_SECRET: "ea482b47a4830d38a838f48be43ab28002c33187",
```
vilagithub: GITHUB_AUTH_CLIENT_ID and GITHUB_AUTH_CLIENT_SECRET look like secrets to me ? Which have therefore leaked ?

package.json
```diff
@@ -64,7 +65,8 @@
     "redux-thunk": "^2.1.0",
     "request": "^2.79.0",
     "winston": "^2.3.0",
-    "winston-daily-rotate-file": "^1.4.2"
+    "winston-daily-rotate-file": "^1.4.2",
+    "yargs": "https://registry.npmjs.org/yargs/-/yargs-3.32.0.tgz"
```
earnubs: semver, not url (npm, grr)

src/config/defaults.js
katiefenn: No. They are placeholder values that I created on my own account and immediately revoked. They have not been used by either the staging or the production sites.

I will be changing them to be obvious placeholder values.

mocks/service/index.js
```diff
@@ -1,3 +1,4 @@
 require('babel-polyfill');
 require('babel-register');
+require('babel-polyfill');
```
katiefenn: The mocks won't start without this line :-(

package.json
```diff
@@ -41,12 +41,11 @@
     "express-winston": "^2.0.0",
     "helmet": "^3.1.0",
     "images-require-hook": "^1.0.3",
+    "immutable": "^3.8.1",
```
katiefenn: Using Immutable for the config storage. There are a few reasons for this:

- It has `get` and `set` methods like nconf, so we won't need to change calls to`conf`
- It is universal-friendly
- I think what we need is not a config management tool, but a generic data structure tool
- We wanted an opportunity to try Immutable out

We could just code our own implementation, but it

package.json
```diff
@@ -41,12 +41,11 @@
...
     "isomorphic-fetch": "^2.2.1",
     "js-yaml": "^3.7.0",
     "memcached": "^2.2.2",
     "moment": "^2.17.1",
-    "nconf": "^0.8.4",
```
katiefenn: I decided to drop nconf primarily because of the unavoidable dependency on the Node fs module. Even if we were to retain it for use in server-only code, we would need to code our own implementation for universal code. I think there are benefits for making the implementation as similar as possible throughout the codebase.

package.json
```diff
@@ -64,7 +63,8 @@
     "redux-thunk": "^2.1.0",
     "request": "^2.79.0",
     "winston": "^2.3.0",
-    "winston-daily-rotate-file": "^1.4.2"
+    "winston-daily-rotate-file": "^1.4.2",
+    "yargs": "^6.4.0"
```
katiefenn: Added yargs as an explicit dependency to consume configuration settings from command-line arguments. It's already used by WebPack, but was not previously in package.json or in the production bundle.

We should be able to remove it from production dependencies when we start configuring deployed applications with environment variables only (I've dropped this objective for this pull request because there is a lot more work needed).

src/common/helpers/config.js
```diff
@@ -1,40 +1,12 @@
...
-    }
-    // root is not window; test for node global and assign to avoid ref error
-    // we don't expect config here, this is client side only, but we can set it
-    // for unit tests
-    root = root || (typeof global == 'object' && global.Object == Object && global);
+import { Map } from 'immutable';
 
-    this._store = config || root.__CONFIG__;
-  }
+export const conf = Map(getGlobal().__CONFIG__);
```
katiefenn: Config data is created as a singleton. Testing this is a bit tricky; testing this module and modules that depend on it requires proxyquire. It does have the benefit of reducing boilerplate in application code.

src/server/helpers/config.js
```diff
@@ -1,51 +1,42 @@
...
+  process.env,                               // environment variables, including envfile
+  argv                                       // CLI arguments
+);
 
-export const getClientConfig = () => {
-  // Return memoized config if it has been built
-  if (configForClient) {
-    return configForClient;
-  }
+export const conf = Map(global.__CONFIG__);
```
katiefenn: Config data is created as a singleton. Testing this is a bit tricky; testing this module and modules that depend on it requires proxyquire. It does have the benefit of reducing boilerplate in application code.

src/server/helpers/config.js
```diff
@@ -1,51 +1,42 @@
...
-
-export const conf = nconf;
+// Compile config from the following sources
+// in order of precedence
+global.__CONFIG__ = Object.assign(
+  {},
+  require('../../config/defaults').default,  // development site defaults
+  process.env,                               // environment variables, including envfile
+  argv                                       // CLI arguments
+);
```
katiefenn: The config helper consumes settings from these sources in descending order of precedence:

- The defaults file
- Environment variables
- Environment variables (injected by dotenv from an env file)
- Command-line arguments

src/common/helpers/config.js
```diff
@@ -1,40 +1,12 @@
...
-      return undefined;
-    }
-
-    return target;
-  }
-
-  _path(key) {
-    return (key == null) ? [] : key.split(':');
-  }
+function getGlobal() {
```
katiefenn: Re-used the window-or-global helper. It's used to take configuration for universal code from either window.__CONFIG__ when run in the browser, or global.__CONFIG__ when run on the server.

test/unit/src/server/handlers/t_login.js
```diff
@@ -1,9 +1,15 @@
 import expect from 'expect';
```
katiefenn: Changes in the `src` directory are fairly clean. Not so with the `test` directory, which screws around with config in order to test stuff.

The changes in this directory are mostly the replacement of `conf.stores` with proxyquire, and tidying up test files after the additional boilerplate had been added.

katiefenn: It might be helpful to use whitespace-insensitive diffs from here on. https://github.com/canonical-ols/build.snapcraft.io/pull/95/files?w=1

mocks/service/index.js
bartaz: We need to require `babel-polyfill` twice?

Or is it a merge issue - I remember adding this line too when implementing GH repo listing API, so I guess we both did it and it ended up twice in the file.

src/common/helpers/config.js
bartaz: Now as it seems it's the only export from this file, can it be default?

Or maybe (as I see later here) it's just to make it consistent with `server/helpers/config.js` - which makes sense.
It's a bit confusing to have both files, but I guess there is no way around it. But I assume that common/config is used in common files only, and server/config only in server side only files, right?

test/unit/src/server/handlers/t_login.js
```diff
@@ -23,31 +29,19 @@ describe('processVerifiedAssertion', () => {
...
-    after(() => {
-      const overrides = conf.stores['test-overrides'];
-      overrides.clear('OPENID_TEAMS');
+  context('when OPENID_TEAMS are configured', () => {
+    let processVerifiedAssertion;
+
+    beforeEach(() => {
+      processVerifiedAssertion = proxyquire(
+        '../../../../../src/server/handlers/login',
+        {
```
bartaz: It looks cumbersome a bit, but fortunately it's just in tests.

Maybe (in undefined future) we could add some 'override' layer to the config? Just as a test affordance, so we could do something like:

```js
before(() => {
    conf.setOverrides({ UBUNTU_SSO_URL, OPENID_VERIFY_URL, OPENID_TEAMS, GITHUB_API_ENDPOINT });
});

after(() => {
    conf.clearOverrides();
});
```

src/server/helpers/config.js
```diff
@@ -1,51 +1,42 @@
-import path from 'path';
-import nconf from 'nconf';
+import { Map } from 'immutable';
+import { argv } from 'yargs';
 import dotenv from 'dotenv';
```
earnubs: dotenv also depends on 'fs', why is it not a problem but nconf is?

mocks/service/index.js
katiefenn: Er - nice spot, I think it is indeed a merge issue 😵

src/common/helpers/config.js
katiefenn: @bartaz: Yes. It makes importing the helper cleaner (because of the shorter path) and it allows us to cleanly separate the implementation. I don't think there's a major reason not to export conf as default, but I guess I prefer helper files with multiple helpers to export them all in the same way ¯\_(ツ)_/¯

src/server/helpers/config.js
katiefenn: We only need to use `dotenv` in `src/server`. We won't ever need to use it in `src/common`. If we kept `nconf`, we would need to re-implement the `get`, `set` and `stores` features in universal code, because these are the features we currently depend on.

I think it's worth having a config API that is mirrored in implementation in server-only and universal code.

test/unit/src/server/handlers/t_login.js
katiefenn: @bartaz I'm not sure about giving the config helper a method just to help with testing. I'd prefer to replace it with a stub using proxyquire if we can.

Having said that, I'm pretty sure I can reduce the boilerplate here. I'll get back to you.

test/unit/src/server/handlers/t_login.js
```diff
@@ -63,15 +65,11 @@ describe('processVerifiedAssertion', () => {
...
-      processVerifiedAssertion = proxyquire(
-        '../../../../../src/server/handlers/login',
-        {
-          '../helpers/config': {
-            conf: Map({ UBUNTU_SSO_URL, OPENID_VERIFY_URL, GITHUB_API_ENDPOINT }),
-            '@noCallThru': true
-          }
-        }
-      ).processVerifiedAssertion;
+      ({ processVerifiedAssertion } = requireWithMockConfig({
```
bartaz: Why the parents `()` around whole thing?

I also see a bit different pattern here and in test above: `processVerifiedAssertion = ...processVerifiedAssertion` vs `{ processVerifiedAssertion } = ...`. Would be good to use one of them consistently.

PR #98, opened 2017-01-24
Repository listing api

bartaz:
Simple internal API route for authenticated requests to get list of user GitHub repositories.

### DIFF
--- a/src/server/handlers/github-auth.js
+++ b/src/server/handlers/github-auth.js
@@ -3,7 +3,7 @@ import logging from '../logging';
 
 import { conf } from '../helpers/config';
 
-const AUTH_SCOPE = 'admin:repo_hook';
+const AUTH_SCOPE = 'admin:repo_hook'; // another scope (like 'public_repo') may be needed to list organisations' repositories
 const GITHUB_AUTH_LOGIN_URL = conf.get('GITHUB_AUTH_LOGIN_URL');
 const GITHUB_AUTH_VERIFY_URL = conf.get('GITHUB_AUTH_VERIFY_URL');
 const GITHUB_AUTH_CLIENT_ID = conf.get('GITHUB_AUTH_CLIENT_ID');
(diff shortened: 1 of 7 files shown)


src/server/handlers/github-auth.js
```diff
@@ -3,7 +3,7 @@ import logging from '../logging';
 
 import { conf } from '../helpers/config';
 
-const AUTH_SCOPE = 'admin:repo_hook';
+const AUTH_SCOPE = 'admin:repo_hook'; // another scope (like 'public_repo') may be needed to list organisations' repositories
```
bartaz: It seems that current OAuth scope we use allows only access to repositories owned by the user (not by organisations). If we want to list all repositories user has access to we need to add another scope to authorisation (probably `public_repo` should be enough [[docs](https://developer.github.com/v3/oauth/#scopes)]).

Another issue is that it doesn't limit repositories to those that user is admin of - and we require admin privileges to create hooks. So we may get some repositories (from organisations or collaboration) that user has access to, but is not admin of and can't create hooks.

I couldn't find any way through API to list only the repositories that user is admin of, so we will probably need to filter it separately or simply check at the time when user wants to build given repo (to avoid API calls for each repo from the list separately).

src/server/routes/github.js
```diff
@@ -1,11 +1,14 @@
...
-import { createWebhook } from '../handlers/github';
+import { createWebhook, listRepositories } from '../handlers/github';
 
 const router = Router();
 
 router.use('/github/webhook', json());
 router.post('/github/webhook', createWebhook);
 
+router.use('/github/repos', json());
+router.get('/github/repos', listRepositories);
```
cjwatson: Please document new routes in `docs/routes.md` (I know not all the existing ones are documented, but we should at least keep up with new ones).

src/server/handlers/github-auth.js
bartaz: Right. I just checked what Travis asks for, and according to my settings page they use:

```
 Read org and team membership
 Access commit status
 Access deployment status
 Access user email addresses (read-only)
 Write repository hooks
```

Not sure how exactly it translates to scope values, but we have a list to compare to when we want to change these.

PR #99, opened 2017-01-26
User repository listing

bartaz:
Added listing of user owned repositories in place of repository input.

screen shot 2017-01-27 at 14 34 02


src/common/components/repositories-home/index.js
```diff
@@ -0,0 +1,68 @@
+import React, { Component, PropTypes } from 'react';
+import { connect } from 'react-redux';
+
+import { fetchUserRepositories } from '../../actions/repositories';
+
```
katiefenn: Picky: Can the first-party dependency imports be grouped together without these line breaks?

test/unit/src/common/actions/t_repositories.js
```diff
@@ -0,0 +1,131 @@
...
+      api = nock(conf.get('BASE_URL'));
+    });
+
+    afterEach(() => {
+      nock.cleanAll();
+    });
+
+    it('should store repositories on fetch success', () => {
+
+      api.get('/api/github/repos')
```
katiefenn: Needs a `beforeEach` block.

test/unit/src/common/actions/t_repositories.js
```diff
@@ -0,0 +1,131 @@
...
+          api.done();
+          expect(store.getActions()).toHaveActionOfType(
+            ActionTypes.SET_REPOSITORIES
+          );
+        });
+    });
+
+    it('should store error on Launchpad request failure', () => {
+
+      api.get('/api/github/repos')
```
katiefenn: Needs a `beforeEach` block.

src/common/components/repositories-home/index.js
bartaz: Sure, I kind of wanted actions to be separate from components, but I guess in most cases it may mean new line under every import. So yep - let's group them all together.

test/unit/src/common/actions/t_repositories.js
bartaz: @katiefenn There is a `beforeEach` block that sets up `nock`, but each test case has different response (one is success, other is failure), so there is not much to move to `beforeEach`.

Unless we wrap each test case in its own `context` and `beforeEach`, but I'm not sure if that is needed nesting for a single test case.

katiefenn: Ah, okay. Is there another way of removing the boilerplate from the `it` block? Can the API mock be extracted to a helper function?

src/common/actions/repositories.js
```diff
@@ -0,0 +1,68 @@
+import 'isomorphic-fetch';
+
+import conf from '../helpers/config';
+
+const BASE_URL = conf.get('BASE_URL');
+
+export const FETCH_REPOSITORIES = 'FETCH_REPOSITORIES';
+export const FETCH_REPOSITORIES_ERROR = 'FETCH_REPOSITORIES_ERROR';
```
earnubs: My reading of FSA's is that you only need the  FETCH_REPOSITORIES action, FETCH_REPOSITORIES_ERROR is replaced by the optional error property of FETCH_REPOSITORIES

src/common/components/repositories-list/index.js
```diff
@@ -0,0 +1,109 @@
...
+
+class RepositoriesList extends Component {
+
+
+  getErrorMessage() {
+    const input = this.props.repositoryInput;
+
+    if (input.error) {
+      const payload = input.error.json.payload;
+      if (payload.code === 'snap-name-not-registered') {
```
earnubs: maybe move to a constant somewhere?

src/common/components/repositories-list/index.js
```diff
@@ -0,0 +1,109 @@
...
+            Please <a href={registerNameUrl}>register it</a> before trying
+            again.
+          </span>
+        );
+      } else {
+        return input.error.message;
+      }
+    }
+
+    return 'Unexpected error. Please make sure you are entering a valid GitHub repository and try again.';
```
earnubs: same, maybe move to string constants somewhere?

src/common/components/repositories-list/index.js
```diff
@@ -0,0 +1,109 @@
...
+      const payload = input.error.json.payload;
+      if (payload.code === 'snap-name-not-registered') {
+        const snapName = payload.snap_name;
+        const devportalUrl = conf.get('STORE_DEVPORTAL_URL');
+        const registerNameUrl = `${devportalUrl}/click-apps/register-name/` +
+                                `?name=${encodeURIComponent(snapName)}`;
+        return (
+          <span>
+            The name provided in the snapcraft.yaml file ({snapName}) is not
+            registered in the store.
```
earnubs: "the store" is probably too vague, linking the term with the url would put it beyond doubt.

test/unit/src/common/actions/t_repositories.js
```diff
@@ -0,0 +1,131 @@
...
+  };
+
+  let store;
+  let action;
+
+  beforeEach(() => {
+    store = mockStore(initialState);
+  });
+
+  context('setRepositories', () => {
```
earnubs: fwiw, arrow syntax in mocha tests is probably an antipattern (and one I might have introduced!) because you can't avail yourself of skip, retries, or other methods

src/common/components/repositories-list/index.js
bartaz: It's just a quick workaround error message introduced earlier to link to name registering page. We should get a new copy with the redesign, so let's wait for that.

src/common/components/repositories-list/index.js
bartaz: Yep.. actually this error message doesn't make so much sense now when there is no repository input.

test/unit/src/common/actions/t_repositories.js
bartaz: Maybe, but all they share now is `api.get('/api/github/repos')`, the response part is different, so there is not so much boilerplate there...

I don't think it's such a big deal here, but for sure if we had more expectations around the same response it would make sense to create a separate context and beforeEach block.

src/common/actions/repositories.js
bartaz: We use the same pattern in other actions/reducers, so full refactoring would be needed.
We have FSA card in backlog, so let's do it separately.

src/common/reducers/repositories-status.js
```diff
@@ -0,0 +1,34 @@
+import * as ActionTypes from '../actions/repository-input';
```
katiefenn: Why does the repository status data exist separately from the repository data? Could this have not been part of the `repositories` reducer?

bartaz: In theory it could. At first it seemed easier to just change existing repository-input reducer to accept status of multiple repos.
Also it makes both reducers smaller and responding to less actions rather than one big one doing lot's of different things.

But that's exactly what I was thinking and what I'm about to mention as possible improvement  in an 'pre-holiday' email I'm about to send ;)

PR #103, opened 2017-01-31
sort builds in build list by id

earnubs:


earnubs:
[Sorting in builds list](https://trello.com/c/jfXluS5g/70-sorting-in-builds-list)

src/common/components/build-history/index.js
```diff
@@ -1,20 +1,23 @@
...
 
 import BuildRow from '../build-row';
 import { Message } from '../forms';
 
 const BuildHistory = (props) => {
-  const { repository, success } = props;
+  const { repository, success, builds } = props;
 
-  const builds = props.builds.map((build) => (
+  sortBy(builds, ['buildId']);
```
cjwatson: The docs say that `sortBy` returns the sorted array, not that it mutates it in place, so shouldn't this be `const sortedBuilds = sortBy(builds, ['buildId'])` or something like that?  (If it did mutate it in place, I'd worry slightly about whether this was mutating `props`.)

src/common/components/build-history/index.js
```diff
@@ -4,21 +4,26 @@ import { connect } from 'react-redux';
...
+  if (!success) {
+    return null;
+  }
+
+  if (!hasBuilds) {
+    return <Message status='info'>This snap has not been built yet.</Message>;
+  }
+
+  const buildRows = builds
+    .sort((a,b) => (+a.buildId < +b.buildId))
```
cjwatson: This will produce wrong results depending on what pairs of elements the comparison function happens to be called with (because if `a.buildId` is greater than `b.buildId`, then this will incorrectly return 0 rather than something negative).  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort suggests using subtraction as the comparison function for integers, so maybe `(+b.buildId) - (+a.BuildId)` instead?

PR #105, opened 2017-02-02
67: Repo pagination

katiefenn:
- New PageLinks component
- Added feature to fetchRepositories handler to transform and pass on
  page links from GitHub API
- Split FETCH_REPOSITORIES_SUCCESS from SET_REPOSITORIES to clean up a
  side-effect
- Repositories are now cleared from store when loading new pages
- New and updated unit tests

### DIFF
--- a/src/common/actions/repositories.js
+++ b/src/common/actions/repositories.js
@@ -8,7 +8,7 @@ export const FETCH_REPOSITORIES = 'FETCH_REPOSITORIES';
 export const FETCH_REPOSITORIES_SUCCESS = 'FETCH_REPOSITORIES_SUCCESS';
 export const FETCH_REPOSITORIES_ERROR = 'FETCH_REPOSITORIES_ERROR';
 export const SET_REPOSITORIES = 'SET_REPOSITORIES';
-export const SET_PAGE_LINKS = 'SET_REPOSITORY_PAGE_LINKS';
+export const SET_REPOSITORY_PAGE_LINKS = 'SET_REPOSITORY_PAGE_LINKS';
 
 export function setRepositories(repos) {
   return {
@@ -89,7 +89,7 @@ export function fetchRepositoriesError(error) {
 
 export function setPageLinks(pageLinks) {
   return {
-    type: SET_PAGE_LINKS,
+    type: SET_REPOSITORY_PAGE_LINKS,
     payload: pageLinks
   };
 }
--- a/src/common/actions/repositories.js
+++ b/src/common/actions/repositories.js
@@ -5,8 +5,10 @@ import { conf } from '../helpers/config';
 const BASE_URL = conf.get('BASE_URL');
 
 export const FETCH_REPOSITORIES = 'FETCH_REPOSITORIES';
+export const FETCH_REPOSITORIES_SUCCESS = 'FETCH_REPOSITORIES_SUCCESS';
 export const FETCH_REPOSITORIES_ERROR = 'FETCH_REPOSITORIES_ERROR';
 export const SET_REPOSITORIES = 'SET_REPOSITORIES';
+export const SET_PAGE_LINKS = 'SET_REPOSITORY_PAGE_LINKS';
 
 export function setRepositories(repos) {
   return {
@@ -35,14 +37,21 @@ function checkStatus(response) {
   }
 }
 
-export function fetchUserRepositories() {
+export function fetchUserRepositories(pageNumber) {
   return (dispatch) => {
+    let urlParts = [BASE_URL, 'api/github/repos'];
+
+    if (pageNumber) {
+      urlParts.push('page/' + pageNumber);
+    }
 
     dispatch({
       type: FETCH_REPOSITORIES
     });
 
-    return fetch(`${BASE_URL}/api/github/repos`, {
+    dispatch(setRepositories([]));
+
+    return fetch(urlParts.join('/'), {
       headers: { 'Content-Type': 'application/json' },
       credentials: 'same-origin'
     })
@@ -53,9 +62,20 @@ export function fetchUserRepositories() {
             throw getError(response, result);
           }
           dispatch(setRepositories(result.payload.repos));
+          dispatch(fetchRepositoriesSuccess());
+
+          if (result.pageLinks) {
+            dispatch(setPageLinks(result.pageLinks));
+          }
         });
       })
       .catch(error => dispatch(fetchRepositoriesError(error)));
+  };
+}
+
+export function fetchRepositoriesSuccess() {
+  return {
+    type: FETCH_REPOSITORIES_SUCCESS
   };
 }
 
@@ -64,5 +84,12 @@ export function fetchRepositoriesError(error) {
     type: FETCH_REPOSITORIES_ERROR,
     payload: error,
     error: true
+  };
+}
+
+export function setPageLinks(pageLinks) {
+  return {
+    type: SET_PAGE_LINKS,
+    payload: pageLinks
   };
 }
--- a/src/common/components/page-links/index.js
+++ b/src/common/components/page-links/index.js
@@ -1,0 +1,63 @@
+import React, { Component, PropTypes } from 'react';
+
+import styles from './styles.css';
+
+const LINK_LABEL_OPTIONS = ['first', 'prev', 'next', 'last'];
+
+export default class PageLinks extends Component {
+
+  render() {
+    const renderLink = this.renderLink.bind(this);
+
+    let links = LINK_LABEL_OPTIONS.map((item) => {
+      return renderLink(
+        this.props[item],
+        item
+      );
+    });
+
+    return (
+      <ul className={ styles.container }>
+        { links }
+      </ul>
+    );
+  }
+
+  renderLink(item, label) {
+    let url;
+
+    if (typeof item == 'string') {
+      url = item;
+    }
+
+    if (typeof item == 'undefined') {
+      return (
+        <li key={ label } className={ styles.link }>
+          <div>
+            { label }
+          </div>
+        </li>
+      );
+    }
+
+    return (
+      <li key={ label } className={ styles.link }>
+        <a href={ url } onClick={ this.onClick.bind(this, item) }>
+          { label }
+        </a>
+      </li>
+    );
+  }
+
+  onClick(item) {
+    this.props.onClick(item);
+  }
+}
+
+PageLinks.propTypes = {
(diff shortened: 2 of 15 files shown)


src/common/actions/repositories.js
```diff
@@ -5,8 +5,10 @@ import { conf } from '../helpers/config';
 const BASE_URL = conf.get('BASE_URL');
 
 export const FETCH_REPOSITORIES = 'FETCH_REPOSITORIES';
+export const FETCH_REPOSITORIES_SUCCESS = 'FETCH_REPOSITORIES_SUCCESS';
```
katiefenn: I've split `FETCH_REPOSITORIES_SUCCESS` out from `SET_REPOSITORIES` so that we can clear repositories when the page is changed without setting the success status too early. Setting success with the repositories is a side-effect - setting them doesn't necessarily mean data has been loaded from service.

src/common/actions/repositories.js
```diff
@@ -35,14 +37,21 @@ function checkStatus(response) {
...
+    if (pageNumber) {
+      urlParts.push('page/' + pageNumber);
+    }
 
     dispatch({
       type: FETCH_REPOSITORIES
     });
 
-    return fetch(`${BASE_URL}/api/github/repos`, {
+    dispatch(setRepositories([]));
```
katiefenn: Clear repositories when page is changed. This allows the spinner to be shown without making the repo list jump down.

src/common/actions/repositories.js
```diff
@@ -53,16 +62,31 @@ export function fetchUserRepositories() {
             throw getError(response, result);
           }
           dispatch(setRepositories(result.payload.repos));
+          dispatch(fetchRepositoriesSuccess());
+          dispatch(setPageLinks(result.pageLinks));
```
katiefenn: Page links are supplied by the service (which are in-turn supplied from the GitHub API) to make stateful pagination in the client unnecessary.

src/common/components/page-links/index.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  renderLink(item, label) {
+    let url;
+
+    if (typeof item == 'string') {
+      url = item;
+    }
+
+    if (typeof item == 'undefined') {
```
katiefenn: Just show the label if a page doesn't exist. This happens when we are viewing the first or last page.

src/common/components/page-links/index.js
```diff
@@ -0,0 +1,63 @@
...
+      <ul className={ styles.container }>
+        { links }
+      </ul>
+    );
+  }
+
+  renderLink(item, label) {
+    let url;
+
+    if (typeof item == 'string') {
```
katiefenn: Only add an href attribute to links if item is a string. This is to support default link behaviour as well as event-handled links.

src/common/components/page-links/index.js
```diff
@@ -0,0 +1,63 @@
...
+          <div>
+            { label }
+          </div>
+        </li>
+      );
+    }
+
+    return (
+      <li key={ label } className={ styles.link }>
+        <a href={ url } onClick={ this.onClick.bind(this, item) }>
```
katiefenn: Call the event handler with the item (either a string URL or a number page number) on click instead of the event object.

src/server/handlers/github.js
```diff
@@ -137,3 +144,35 @@ const getRequest = (owner, name, token, secret) => {
...
+    }
+    let number = parseInt(section[0]
+                  .replace(/per_page=(\d+)/, '') // Make sure per_page does not match
+                  .match(/page=(\d+)/)[1]);
+    let name = section[1].replace(/rel="(.*)"/, '$1').trim();
+    links[name] = number;
+  }
+
+  return links;
+};
```
katiefenn: The fact that this is not easier makes me 😢

src/common/actions/repositories.js
```diff
@@ -5,8 +5,10 @@ import { conf } from '../helpers/config';
 const BASE_URL = conf.get('BASE_URL');
 
 export const FETCH_REPOSITORIES = 'FETCH_REPOSITORIES';
+export const FETCH_REPOSITORIES_SUCCESS = 'FETCH_REPOSITORIES_SUCCESS';
 export const FETCH_REPOSITORIES_ERROR = 'FETCH_REPOSITORIES_ERROR';
 export const SET_REPOSITORIES = 'SET_REPOSITORIES';
+export const SET_PAGE_LINKS = 'SET_REPOSITORY_PAGE_LINKS';
```
cjwatson: Maybe make these match for ease of grepping later?

src/common/components/page-links/index.js
cjwatson: This sort of comment should probably be a comment in the source code itself rather than being buried in a PR and hard to find in future.  (Ditto other similar comments.)

What happens if the item is something other than a string or undefined?  That looks like it'll render an `<a>` element with `href` set to undefined; I don't know off the top of my head what that will do.

src/common/components/repositories-list/index.js
```diff
@@ -59,17 +62,34 @@ class RepositoriesList extends Component {
...
+
   render() {
     const isLoading = this.props.repositories.isFetching;
+    const pageLinks = this.props.repositories.pageLinks;
 
     return (
       <div>
+        { this.props.repositories.success && pageLinks &&
+          <div className={ styles['page-links-container'] }>
+            Pages: <PageLinks { ...pageLinks } onClick={ this.onPageLinkClick.bind(this) } />
```
cjwatson: The "Pages:" text seems a bit unnecessary, here and below; I'd say the links should probably stand by themselves.

src/server/routes/github.js
```diff
@@ -10,5 +10,6 @@ router.post('/github/webhook', createWebhook);
 
 router.use('/github/repos', json());
 router.get('/github/repos', listRepositories);
+router.get('/github/repos/page/:page', listRepositories);
```
cjwatson: Not a blocker, but to me the page feels more like a query string parameter than a route parameter.  (That would also match GitHub.)

Whichever way you go, this could also do with corresponding changes to the route documentation in `docs/routes.md`.

src/server/handlers/github.js
```diff
@@ -46,12 +47,17 @@ const RESPONSE_CREATED = {
...
 
   if (!req.session || !req.session.token) {
     return res.status(401).send(RESPONSE_AUTHENTICATION_FAILED);
   }
 
+  if (req.params.page) {
+    params.push('page=' + req.params.page);
+  }
+
+  const uri = REPOSITORY_ENDPOINT + '?' + params.join('&');
```
cjwatson: Consider using `qs.stringify` instead.  Doesn't matter much here, but might save us from quoting errors in future.

src/server/handlers/github.js
```diff
@@ -64,13 +70,19 @@ export const listRepositories = (req, res) => {
...
         payload: {
           code: 'github-list-repositories',
           repos: response.body
         }
-      });
+      };
+
+      if (response.headers.link) {
+        body.pageLinks = parseLinkHeader(response.headers.link);
+      }
```
cjwatson: This could do with corresponding changes to the route documentation in `docs/routes.md`.

src/server/handlers/github.js
```diff
@@ -137,3 +149,35 @@ const getRequest = (owner, name, token, secret) => {
...
   };
 };
+
+/*
+ * parse_link_header()
+ *
+ * Parse the Github Link HTTP header used for pageination
+ * http://developer.github.com/v3/#pagination
+ *
+ * Modified by kfenn to return page numbers instead of urls
```
cjwatson: I feel that if we're working out page numbers anyway then we might as well present those rather than first/prev/next/last, but that can be in a later PR.

src/server/handlers/github.js
```diff
@@ -137,3 +149,35 @@ const getRequest = (owner, name, token, secret) => {
...
+  let links = {};
+  // Parse each part into a named link
+  for (let index=0; index<parts.length; index++) {
+    let section = parts[index].split(';');
+    if (section.length != 2) {
+      throw new Error('section could not be split on ";"');
+    }
+    let number = parseInt(section[0]
+                  .replace(/per_page=(\d+)/, '') // Make sure per_page does not match
+                  .match(/page=(\d+)/)[1]);
```
cjwatson: Something involving `url.parse(..., true)` would be neater and would relieve you from fiddling with `per_page`.

src/server/handlers/github.js
cjwatson: There are a few existing modules that do this that I found by searching NPM for "link header", e.g. [parse-link-header](https://www.npmjs.com/package/parse-link-header).  I haven't looked at the code for any of them.

src/common/components/page-links/index.js
katiefenn: I add comments to my PRs purely to make reviewing easier, but if you think a comment is needed in the source code I can add one 👍

src/server/handlers/github.js
katiefenn: @cjwatson: Yeah, extracting page numbers from the GitHub API response still saves us work, which I am trying to do at every opportunity.

src/server/handlers/github.js
katiefenn: @cjwatson: True. For something this small, I'd prefer to copy an example from the community and add it to our own source code rather than installing a module. It's simple enough that we can maintain it ourselves while avoiding the risk of something happening upstream, `left-pad` style. I will probably replace this with a helper we can re-use in `src/server/helpers` in time.

src/common/components/page-links/index.js
cjwatson: The line I tend to draw is that if the comment is purely about the change then it can go in the PR, but if it's about the new state of the code then it's better to put it in the code because it might equally be useful for people trying to understand the code in future.

src/server/handlers/github.js
```diff
@@ -46,12 +50,19 @@ const RESPONSE_CREATED = {
 };
 
 export const listRepositories = (req, res) => {
-  const uri = '/user/repos?affiliation=owner';
+  let params = {
```
cjwatson: Can be `const`.

src/server/handlers/github.js
```diff
@@ -64,13 +75,19 @@ export const listRepositories = (req, res) => {
         });
       }
 
-      return res.status(response.statusCode).send({
+      let body = {
```
cjwatson: Can be `const`.

PR #107, opened 2017-02-06
100: Batched repo list

katiefenn:
- Added new selected-repositories-form reducer
    - Removed "Page" label from pagelinks
    - Added new SelectRepositoryList and SelectRepositoryRow component.
      These stand alone for now to make merging easier


src/common/components/select-repositories-list/index.js
```diff
@@ -0,0 +1,116 @@
+import React, { Component, PropTypes } from 'react';
```
katiefenn: I've split out this file from `repositories-list` instead of modifying the existing file. It means I can continue working on this feature without breaking staging, and it should make integrating work a bit easier.

We are also coming to the point where we have two distinct repository lists:

- One for selecting repositories to be built
- One for reviewing repositories that are being built

src/common/actions/repositories.js
```diff
@@ -3,6 +3,7 @@ import 'isomorphic-fetch';
 import { conf } from '../helpers/config';
 
 const BASE_URL = conf.get('BASE_URL');
+const log = console;
```
katiefenn: Get around console statement linting.

Correct me if I'm wrong, can we not use winston here?

src/common/actions/repositories.js
```diff
@@ -67,9 +68,15 @@ export function fetchUserRepositories(pageNumber) {
...
             dispatch(setPageLinks(result.pageLinks));
           }
+        })
+        .catch(error => {
+          return Promise.reject(error);
         });
       })
-      .catch(error => dispatch(fetchRepositoriesError(error)));
+      .catch(error => {
+        log.warn(error);
```
katiefenn: I've reworked the error handling here because errors thrown by actions inside the promise were being swallowed. The errors were actually being thrown inside a `src/common/components/select-repository-list/index.js`, so why this would be swallowed by this promise is not known.

src/common/components/repositories-list/index.js
```diff
@@ -87,7 +87,7 @@ class RepositoriesList extends Component {
         }
         { this.props.repositories.success && pageLinks &&
           <div className={ styles['page-links-container'] }>
-            Pages: <PageLinks { ...pageLinks } onClick={ this.onPageLinkClick.bind(this) } />
+            <PageLinks { ...pageLinks } onClick={ this.onPageLinkClick.bind(this) } />
```
katiefenn: Removed a "Pages" label that had been forgotten from previous PR.

src/common/components/select-repository-list/index.js
```diff
@@ -0,0 +1,135 @@
+import React, { Component, PropTypes } from 'react';
```
katiefenn: Most of this component has been copied from `src/common/components/repository-list.js`. I've separated it out into its own page and its own route to make merging with work @bartaz is doing easier.

src/common/components/select-repository-row/index.js
```diff
@@ -0,0 +1,49 @@
+import React, { Component, PropTypes } from 'react';
```
katiefenn: Mostly copied from `src/common/components/repository-row/index.js`.

src/common/actions/select-repositories-form.js
```diff
@@ -0,0 +1,8 @@
+export const TOGGLE_REPOSITORY = 'SELECT_REPOSITORY_FORM_TOGGLE_REPOSITORY';
```
katiefenn: The reason the value is not the same as the key is that I expect "TOGGLE_REPOSITORY" to one day be used in more than one place, so I am namespacing the action value here so that it doesn't clash if we do add similar action elsewhere.

src/common/actions/repositories.js
bartaz: What kind of errors is it catching? Why do we want to console.log them (in production)?

src/common/routes.js
```diff
@@ -14,5 +15,6 @@ export default (
     <Route path="/:owner/:name/builds" component={Builds}/>
     <Route path="/:owner/:name/builds/:buildId" component={BuildDetails}/>
     <Route path="/login/failed" component={LoginFailed}/>
+    <Route path="/dashboard/select-repositories" component={SelectRepositories}/>
```
katiefenn: Opinions on what the route should be?

- `/dashboard/katiefenn/select-repositories`
- `/dashboard/katiefenn/add-repositories`
- `/dashboard/katiefenn/add`
- `/dashboard`?

katiefenn: I think the dashboard definitely needs a route of its own.

- It allows deep-linking of the dashboard and the "select-repositories" page
- It allows users to view dashboards when logged out
- We can get the benefits of hooking into the browser history API

bartaz: Suggested URL for dashboard in UX doc is `/repos`.

I'm not sure if at this point we want not signed in users to see dashboards. What we are building is an 'admin' interface (adding repos, triggering builds, registering names) - these cannot be visible without logging in. We would need to have separate read-only view to handle it.

Travis simply has 'dashboard' view on route based on github user name: `https://travis-ci.org/canonical-ols/` - it looks nice, but is quite restrictive (meaning we can't have any other first level route (`/dashboard` `/settings`, etc) without possible collisions with GH user names. For example `repos` is a real user name https://github.com/repos.

...

It seems from the doc that the URL is still being discussed https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.or9ei8m7j2f0

So I suppose we need to keep it as something (I guess /dashboard and /dashboard/add would be fine for start) and we can update it later.

src/common/components/select-repositories-page/index.js
```diff
@@ -0,0 +1,41 @@
...
+
+    if (authenticated) {
+      this.props.dispatch(fetchUserRepositories());
+    }
+  }
+
+  render() {
+    return (
+      <div>
+        <h2>You’ve successfully imported all your repos from GitHub!</h2>
```
bartaz: This heading looks bit scary.
Listing user repositories doesn't mean we did anything with them, especially we didn't 'import' them or anything.

src/common/components/select-repository-list/index.js
```diff
@@ -0,0 +1,135 @@
...
+
+  render() {
+    const isLoading = this.props.repositories.isFetching;
+    const pageLinks = this.props.repositories.pageLinks;
+
+    return (
+      <div>
+        { this.props.repositories.success && pageLinks &&
+          <div className={ styles['page-links-container'] }>
+            <PageLinks { ...pageLinks } onClick={ this.onPageLinkClick.bind(this) } />
```
bartaz: It may be an issue in PageLinks, not related specifically to this PR, but it seems it shows unclickable 'First | Prev | Next | Last' even if there is only one page and no navigation is needed.

src/common/routes.js
katiefenn: @bartaz: We don't necessarily have to have separate routes for "logged in" and "logged out" versions of the dashboard. If we go with `dashboard/add`, `katiefenn/dashboard/add` or `katiefenn`, we could write an Express handler to redirect requests to `/` in the absence of a GitHub auth key.

This gives us the ability to railroad users into the landing page as well as giving us URL state.

bartaz: Yes, but if the dashboard is also meant to be a public profile page (as suggested for future in UX docs) there would need to be 2 views of dashboard (edit and read only).

But either way - it doesn't affect us now.

src/common/actions/repositories.js
katiefenn: @bartaz:
screen shot 2017-02-08 at 09 42 04

If I don't log the error out, this fails silently. It's also another story for another day about why **this** error is being thrown **here**, but I have limited time to investigate this right now.

src/common/components/select-repository-list/index.js
katiefenn: @bartaz This is an assumption on my behalf, because we don't have a wireframe for just one page of results. It will be easy to change if design direction goes that way.

katiefenn:
Initial styling for the "select repositories" page:

screen shot 2017-02-08 at 13 22 27

src/common/components/vanilla/button/button-positive.css
```diff
@@ -0,0 +1,29 @@
+$button-color: $success;
```
katiefenn: I've created the new directory `src/common/components/vanilla` for components I am implementing with the latest Vanilla styles. In the past, we created components in `/src/common/component` (e.g. `/src/common/button`). I hoped this would mark the start of an effort to make a modular version of Vanilla that we can publish in npm.

We should move the existing `<Button>` component in here, but there seems to be no equivalent to the branded orange buttons currently in use.

src/common/components/vanilla/button/button-positive.js
```diff
@@ -0,0 +1,15 @@
...
+  return (
+    <button className={ styles['button-positive'] } onClick={ props.onClick }>
+      { props.children }
+    </button>
+  );
+}
+
+ButtonPositive.propTypes = {
+  children: PropTypes.node,
+  onClick: PropTypes.func
```
bartaz: It would be better to add vanilla styles to current `Button` implementation I guess.
We already cover more cases and props there (disabled, type for forms, href for anchor).

src/common/components/vanilla/button/button.css
```diff
@@ -0,0 +1,3 @@
+.button {
+  font-family: "Ubuntu", -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
```
bartaz: Vanilla defined separate font family for buttons (or other components)?

This seems like a good candidate for moving to variables.

src/common/components/vanilla/button/button-positive.js
earnubs: -1 to vanilla folder

src/common/actions/repositories.js
earnubs: don't "get around" linting, just put in a comment to disable linting, either inline or doc wide, then it's clear. :)
```
// eslint-disable-line no-console
```

src/common/actions/repositories.js
earnubs: How can styles not be defined in prod?

src/common/actions/select-repositories-form.js
earnubs: Premature optimisation :)

src/common/components/select-repository-list/index.js
earnubs: I would expect a good design to not show those links if paging wasn't needed.

src/common/components/vanilla/button/button-positive.css
earnubs: -1 to separate Vanilla folder, the components should all be Vanilla.

src/common/helpers/repositories.js
```diff
@@ -0,0 +1,5 @@
+export function hasRepository(arr, repository) {
+  return arr.filter((item) => {
+    return item.fullName === repository.fullName;
```
earnubs: return (item.fullName === repository.fullName);

src/common/reducers/select-repositories-form.js
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export function selectRepositoriesForm(state = initialState, action) {
+  switch(action.type) {
+    case ActionTypes.TOGGLE_REPOSITORY:
+      if (hasRepository(state.selectedRepos, action.payload)) {
+        return {
+          ...state,
+          selectedRepos: state.selectedRepos.filter(
+            (repo) => repo.fullName !== action.payload.fullName
```
earnubs: My 2c is that anything more complex than a single expression return should be cuddled, for clarity and the avoidance of mistakes down the line.

test/unit/src/common/components/select-repository-row/t_select-repository-row.js
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  context('when a repository is supplied', () => {
+    beforeEach(() => {
+      props.repository = {
+        fullName: 'katiefenn/parker'
+      };
+    });
+
+    it('should display a label with the name of the repository', () => {
```
earnubs: FWIW I suggest that arrow syntax in mocha tests is an antipattern, as it prevents the correct use of `this` to skip, retry, set slow limits, or any of mochas test methods.

cjwatson: We should possibly redo this later; at the moment we're using arrow syntax everywhere in tests.

earnubs: Yep, and that's my fault. It doesn't matter to me if this is changed here or later, but going forward we should treat it as an antipattern.

src/common/routes.js
cjwatson: FWIW `dashboard` is a reserved GitHub username (it must be, since https://github.com/dashboard has a special meaning), so if we want to use that at our top level it's safe to do so.  `settings` is similarly reserved.

src/common/actions/repositories.js
katiefenn: @earnubs: They won't be. This was a run of the mill development bug I encountered while I was working on a component file. I forgot to define `styles`. The surprising bit that this error was swallowed by a promise you'd expect was unrelated 😩

src/common/components/vanilla/button/button-positive.css
katiefenn: @bartaz: A separate folder makes it marginally easier to extract into a separate folder later.

My feeling with the buttons is to have a different component for every named component in Vanilla. The difference is marginal, it's a personal preference to have numerous smaller elements than fewer switched elements.

bartaz: @katiefenn but if smaller components share most of the code and differ only in style it would be good to have one implementation (with possible switches) and exports of different components with those switches set to different values. What I mean is to have `Button` that can be anything (with some style attribute), but we may also have `ButtonPositive` that sets the style to some specific value.

src/common/actions/repositories.js
```diff
@@ -74,7 +73,10 @@ export function fetchUserRepositories(pageNumber) {
         });
       })
       .catch(error => {
-        log.warn(error);
+        // TODO: Replace with logging helper
```
katiefenn: Replaced "workaround" with linter annotation.

src/common/actions/select-repositories-form.js
```diff
@@ -1,4 +1,4 @@
-export const TOGGLE_REPOSITORY = 'SELECT_REPOSITORY_FORM_TOGGLE_REPOSITORY';
+export const TOGGLE_REPOSITORY = 'TOGGLE_REPOSITORY';
```
katiefenn: Renamed action value.

src/common/components/button/button.css
```diff
@@ -1,69 +0,0 @@
-$primary-button-color: $ubuntu-orange;
```
katiefenn: The existing Button component has been relocated to the vanilla directory, and replaces the new component I was using.

src/common/components/vanilla/button/button.css
```diff
@@ -1,3 +1,89 @@
...
+$secondary-border-color: $mid-grey;
+
+$positive-button-color: $success;
+$positive-button-color-hover: #0a5615;
+$positive-button-text-color: #FFF;
+$positive-border-color: $success;
+
 .button {
-  font-family: "Ubuntu", -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
+  font-family: $base-font-family;
```
katiefenn: Moved font declaration to a variable.

src/common/style/variables.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
   'success':             '#0f8420', // success notifications
   'information':         '#007aa6', // information notifications
 
-  'base-font-family':    'Ubuntu, Arial, "libra sans", sans-serif',
+  'base-font-family':    '"Ubuntu", -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif',
```
katiefenn: Updated the base font to match the latest Vanilla style.

src/common/components/select-repositories-page/index.js
```diff
@@ -0,0 +1,50 @@
...
+    if (authenticated) {
+      this.props.dispatch(fetchUserRepositories());
+    }
+  }
+
+  render() {
+    return (
+      <div>
+        <HeadingTwo>
+          All your repositories have been successfully added to build.snapcraft.io
```
cjwatson: Feels a bit odd, but I guess this is pending copy?

src/common/components/select-repository-list/index.js
```diff
@@ -0,0 +1,145 @@
...
+        }
+        <div className={ styles.footer }>
+          <div className={ styles.left }>
+            <HeadingThree>
+              In order to enable your repositories, you need to sign in to your Ubuntu One account.
+            </HeadingThree>
+          </div>
+          <div className={ styles.right }>
+            <Button onClick={ this.onSubmit.bind(this) } appearance={ 'positive' }>
+              Enable repos
```
cjwatson: I don't know if this is in the copy from design, but personally I'd prefer that we not abbreviate "repositories" to "repos" outside of casual sketches.

src/common/components/select-repository-list/index.js
```diff
@@ -0,0 +1,145 @@
...
+          <div className={ spinnerStyles }><Spinner /></div>
+        }
+        { this.props.repositories.success &&
+          this.props.repositories.repos.map(this.renderRepository.bind(this))
+        }
+        { this.props.repositories.success && pageLinks &&
+          <div className={ styles['page-links-container'] }>
+            <PageLinks { ...pageLinks } onClick={ this.onPageLinkClick.bind(this) } />
+          </div>
+        }
```
cjwatson: Assign this to a variable to avoid duplication with the same links above the repository list.

src/common/routes.js
```diff
@@ -16,5 +17,6 @@ export default (
...
```
cjwatson: Can you move this up a few lines so that it's next to `/dashboard`?

I think I'd prefer a less wordy name; `/dashboard/add` sounds OK.

src/common/components/select-repositories-page/index.js
katiefenn: @cjwatson: It was the most recent copy in the design spec document at the time of writing.

test/unit/src/common/components/select-repository-row/t_select-repository-row.js
katiefenn: It's heavily ingrained in me at the moment - we used them everywhere in my last job too. Agreed that we can fix this later.

src/common/components/select-repository-list/index.js
katiefenn: It's most recent copy from the design specification at the time of writing. I'm anticipating that we'll update copy across the board when it's ready.

PR #110, opened 2017-02-07
Updated footer with legal info and social icons

bartaz:
screen shot 2017-02-07 at 10 37 07

Footer updated to be as on http://samepage.io


src/common/components/footer/footer.css
```diff
@@ -36,5 +36,61 @@ body {
...
 .copyright {
-  margin-bottom: 0.3em;
+  margin-bottom: 1em;
+}
+
+.social {
+  float: right;
+}
+
+.heading {
```
bartaz: Style updates based on https://github.com/ubuntudesign/snapcraft.io/blob/master/_sass/_footer.scss

PR #111, opened 2017-02-07
Separating landing and dashboard routes and containers

bartaz:
Added separate landing container.

With this PR:

- Repository listing is moved to new `/dashboard` route (URL will likely to be changed later).
- `build.snapcraft.io` renders landing container.
- When user signs in, GitHub redirects back to `/dashboard`.

### DIFF
--- a/src/common/containers/dashboard.js
+++ b/src/common/containers/dashboard.js
@@ -4,7 +4,7 @@ import Helmet from 'react-helmet';
 import RepositoriesHome from '../components/repositories-home';
 import styles from './container.css';
 
-class Home extends Component {
+class Dashboard extends Component {
   render() {
     return (
       <div className={ styles.container }>
@@ -17,4 +17,4 @@ class Home extends Component {
   }
 }
 
-export default Home;
+export default Dashboard;
--- a/src/common/containers/landing.js
+++ b/src/common/containers/landing.js
@@ -1,0 +1,43 @@
+import React, { Component, PropTypes } from 'react';
+import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
+
+import { Anchor } from '../components/button';
+
+import styles from './container.css';
+
+class Landing extends Component {
+  componentWillMount() {
+    if (this.props.auth.authenticated) {
+      this.props.router.replace('/repos');
+    }
+  }
+
+  render() {
+    return (
+      <div className={ styles.container }>
+        <h2>Auto-build and publish software for any Linux system.</h2>
+        <p>Just a placeholder landing page...</p>
+        <Anchor href="/auth/authenticate">Set Up in Minutes</Anchor>
+      </div>
+    );
+  }
+}
+
+Landing.propTypes = {
+  children: PropTypes.node,
+  auth: PropTypes.object,
+  router: PropTypes.object
+};
+
+function mapStateToProps(state) {
+  const {
+    auth
+  } = state;
+
+  return {
+    auth
+  };
+}
+
+export default connect(mapStateToProps)(withRouter(Landing));
--- a/src/common/containers/landing.js
+++ b/src/common/containers/landing.js
@@ -1,18 +1,11 @@
 import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
-import { withRouter } from 'react-router';
 
 import { Anchor } from '../components/button';
 
 import styles from './container.css';
 
 class Landing extends Component {
-  componentWillMount() {
-    if (this.props.auth.authenticated) {
-      this.props.router.replace('/repos');
-    }
-  }
-
   render() {
     return (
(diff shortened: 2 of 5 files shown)


src/common/containers/dashboard.js
```diff
@@ -4,7 +4,7 @@ import Helmet from 'react-helmet';
 import RepositoriesHome from '../components/repositories-home';
 import styles from './container.css';
 
-class Home extends Component {
+class Dashboard extends Component {
```
bartaz: Renamed as 'Home' was confusing with separate landing page.

src/common/containers/landing.js
```diff
@@ -0,0 +1,43 @@
...
+import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
+
+import { Anchor } from '../components/button';
+
+import styles from './container.css';
+
+class Landing extends Component {
+  componentWillMount() {
+    if (this.props.auth.authenticated) {
```
bartaz: If landing container detects authenticated user it redirects to dashboard (`/repos`) route.

So authenticated user will not see landing page (they will always get to dashboard), but dashboard still has separate URL.

This probably means that we also need to verify if user is authenticated on `/repos` route, so that not authenticated user would not be able to visit it (redirect back to landing? or just go to GH login directly?).

Other way to solve this would be to keep both landing and dashboard in `/` route (same container), but during render we would detect if user authenticated or not and render different component (landing page contents or user dashboard).

@cjwatson @katiefenn what do you think?

cjwatson: This is among the reasons I'm not a fan of the separate `/repos` route; at best it just seems to move complexity around.  As I mentioned in the call this morning, I'd prefer to have them both in `/` seeing as we need to test whether the user is authenticated anyway.

Matthew said that they were trying to prepare for one user being able to see another user's dashboard in the future, which would necessitate it being on `/:user` or similar.

Also, without a complete URL reorganisation, we can't use `/repos`, since there exists a GitHub user called "repos".

bartaz: Agreed. Seeing some public dashboard of another user is different story and probably would likely have different URL scheme from current user dashboard. This could work for example by having: landing/dashboard on `/` (for current user) and `/:user` would be a 'public' dashboard view. So as a logged in user you would see 'edit' view of your dashboard at `/` and any public 'profile' of any user (including you) at `/:user`.

So I'm for having both landing and dashboard at `/` too.
The only reason I see for having them as separate URLs is for auth user to get back to landing page (for example to see some info on it or links to documentation or something).

cjwatson: We shouldn't be causing users to rely on the landing page for documentation.  IMO the way to get back to the landing page is to sign out.

katiefenn: I don't understand the reason for having `/` redirect, or switch to, the dashboard. I don't think there is anything that qualifies build.snapcraft.io as a SPA, merely that it is a site that happens to be built using React. URLs are the conventional state of the location within an app, and this seemed like an obvious example for separating pages by route.

Take for instance the "select repos" dashboard. If this is also located on `/`, then a user can't use the browser's history buttons to return to the "list repos" dashboard. That's breaking the web.

If the redirect to the `/repos` route stands, I would like to suggest this redirect is done using an HTTP redirect in an Express handler instead. This follows convention and is where I would expect to look to find a landing page redirect.

src/common/routes.js
```diff
@@ -3,13 +3,15 @@ import { Route } from 'react-router';
...
+import Dashboard from './containers/dashboard.js';
+import Landing from './containers/landing.js';
 import LoginFailed from './containers/login-failed.js';
 import RepositorySetup from './containers/repository-setup.js';
 
 export default (
   <Route component={App}>
-    <Route path="/" component={Home}/>
+    <Route path="/" component={Landing}/>
+    <Route path="/repos" component={Dashboard}/>
```
katiefenn: Why not `/dashboard`? If that's what we're using to describe this interface, why shouldn't the user?

src/common/containers/landing.js
bartaz: Yes, redirect in component is a hack.

From the discussion here and in UX doc there seems to be preference for separate URL for dashboard, so it will probably end up like this.

Travis has dashboard on `/` and it works fine. Also doesn't really break the history (all other views have their URLs).

But I also agree with benefits of having access to landing page even if user is logged in. But still, these benefits are rather occasional (if they need to refer to some info there) and users that have account and are logged it are most likely to be interested in their dashboard and not landing page when they visit build.snapcraft.io.

Anyway. The URL decision needs to be taken outside of this PR anyway, but we can start with some URL and interate on that (as discussed in https://github.com/canonical-ols/build.snapcraft.io/pull/107#discussion_r100009669).

For the redirect I agree that it would be best to keep it server side.

PR #113, opened 2017-02-08
Added api route to list snaps by repository owner

bartaz:


src/server/routes/launchpad.js
```diff
@@ -14,6 +15,9 @@ router.use('/launchpad/snaps', json());
 router.post('/launchpad/snaps', newSnap);
 router.get('/launchpad/snaps', findSnap);
 
+router.use('/launchpad/snaps/list', json());
+router.get('/launchpad/snaps/list', findSnaps);
```
bartaz: I wasn't sure about the route here as already existing`GET /launchpad/snaps` seems to fit nice here, but I didn't want to break the compatibility of previous calls for now.

src/server/handlers/launchpad.js
```diff
@@ -340,6 +340,47 @@ export const internalFindSnap = async (repositoryUrl) => {
...
+  });
+};
+
+export const findSnaps = (req, res) => {
+  internalFindSnapsByPrefix(`https://github.com/${req.query.owner}/`)
+    .then((snaps) => {
+      return res.status(200).send({
+        status: 'success',
+        payload: {
+          code: 'snaps-found',
```
bartaz: LP returns success with empty list when nothing is found.

I keep it like this, because I'm not sure if we need 404 here in case of empty list.

src/server/handlers/launchpad.js
```diff
@@ -340,6 +340,47 @@ export const internalFindSnap = async (repositoryUrl) => {
   });
 };
 
+const internalFindSnapsByPrefix = (urlPrefix) => {
+  const username = conf.get('LP_API_USERNAME');
+
+  return getLaunchpad().named_get('/+snaps', 'findByURLPrefix', {
```
bartaz: I guess it might be useful to memcache the responses here (like we do with finding snap by URL)?

src/server/handlers/launchpad.js
```diff
@@ -340,6 +340,47 @@ export const internalFindSnap = async (repositoryUrl) => {
...
+};
+
+export const findSnaps = (req, res) => {
+  internalFindSnapsByPrefix(`https://github.com/${req.query.owner}/`)
+    .then((snaps) => {
+      return res.status(200).send({
+        status: 'success',
+        payload: {
+          code: 'snaps-found',
+          snaps: snaps
```
bartaz: Sending whole snap object from LP to client.
May be more data then we need, but I guess we should let the client decide what they need (for sure GH repo URL, possibly API URLs for builds, etc)

src/server/handlers/launchpad.js
cjwatson: Yes, we need to memcache this, as it's potentially somewhat expensive to compute and we'll be calling it quite a bit.  However, we need to make sure to clear relevant cached values when we make changes.

src/server/handlers/launchpad.js
cjwatson: An empty list seems more correct to me.

src/server/routes/launchpad.js
cjwatson: I'd prefer `GET /launchpad/snaps` in the long-term, but it'd be a bit confusing since they have different return types.  Let's stick with this for now and adjust later; I suspect that `GET /launchpad/snaps` can be decommissioned once the new authz workflow is in place, at which point we'll have an easier time rearranging it.

PR #114, opened 2017-02-09
Initiate the snapcraft.yaml check from the client

cjwatson:
We'll soon need to move the process of getting a package_upload macaroon
to the client side, which means that the client will need to know the
snap name when it's creating a snap.  Add an
`/api/github/snapcraft-yaml/:owner/:name` route to fetch the parsed
file, and change `/api/launchpad/snaps` to take the snap name, series,
and channels as parameters; this lets us control more of the data flow
from the client side.

### DIFF
--- a/src/common/actions/create-snap.js
+++ b/src/common/actions/create-snap.js
@@ -8,7 +8,13 @@ const BASE_URL = conf.get('BASE_URL');
 
 export const SET_GITHUB_REPOSITORY = 'SET_GITHUB_REPOSITORY';
 export const CREATE_SNAP = 'CREATE_SNAP';
+export const CREATE_SNAP_SUCCESS = 'CREATE_SNAP_SUCCESS';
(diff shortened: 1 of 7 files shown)


src/common/actions/create-snap.js
```diff
@@ -8,33 +8,75 @@ const BASE_URL = conf.get('BASE_URL');
...
 export function setGitHubRepository(value) {
   return {
     type: SET_GITHUB_REPOSITORY,
     payload: value
   };
 }
 
+// Just enough to satisfy higher-level code that might receive errors from
+// our internal API or errors thrown directly from here.
+function throwAPICompatibleError(payload) {
```
katiefenn: Have you considered creating a sub-class of Error instead? It might be more idiomatic.

```
class APICompatibleError extends Error {
  constructor(payload) {
    this.message = ...
    this.json = ....
  }
}

throw new APICompatibleError(payload);
```

PR #115, opened 2017-02-09
Hiding secret properties in as non-enumerable

bartaz:
Fix for leaked lp_client secret keys in API responses.

### DIFF
--- a/src/server/launchpad/resources.js
+++ b/src/server/launchpad/resources.js
@@ -9,11 +9,21 @@ import { normalizeURI } from './uri';
 /** The base class for objects retrieved from Launchpad's web service. */
 export class Resource {
   /** Initialize a resource with its representation and URI. */
-  init(client, representation, uri) {
-    this.lp_client = client;
-    this.uri = uri;
-    for (const key of Object.keys(representation)) {
-      this[key] = representation[key];
+  init(client, uri, representation) {
+    // defining internal properties as non-enumerable not to leak any secrets
+    Object.defineProperty(this, 'lp_client', {
+      value: client,
+      enumerable: false
+    });
+    Object.defineProperty(this, 'uri', {
+      value: uri,
+      enumerable: false
+    });
+
+    if (representation) {
+      for (const key of Object.keys(representation)) {
+        this[key] = representation[key];
+      }
     }
   }
 
@@ -32,7 +42,7 @@ export class Resource {
 export class Root extends Resource {
   constructor(client, representation, uri) {
     super();
-    this.init(client, representation, uri);
+    this.init(client, uri, representation);
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/server/launchpad/resources.js
```diff
@@ -9,11 +9,21 @@ import { normalizeURI } from './uri';
...
+    Object.defineProperty(this, 'lp_client', {
+      value: client,
+      enumerable: false
+    });
+    Object.defineProperty(this, 'uri', {
+      value: uri,
+      enumerable: false
+    });
+
+    if (representation) {
```
bartaz: Making `representation` optional, so that collection `Entry` can call it to get `lp_client` and `uri` without setting `representation`.

src/server/launchpad/resources.js
```diff
@@ -9,11 +9,21 @@ import { normalizeURI } from './uri';
...
 export class Resource {
   /** Initialize a resource with its representation and URI. */
-  init(client, representation, uri) {
-    this.lp_client = client;
-    this.uri = uri;
-    for (const key of Object.keys(representation)) {
-      this[key] = representation[key];
+  init(client, uri, representation) {
+    // defining internal properties as non-enumerable not to leak any secrets
+    Object.defineProperty(this, 'lp_client', {
```
bartaz: I made sure in other parts that `lp_client` is always inherited (via `extends Resource`), so shouldn't be enumerable, but as it contains very sensitive data it's better to make it explicitly non-enumerable in case we instantiate `Resource` directly at some point.

src/server/launchpad/resources.js
```diff
@@ -75,10 +85,18 @@ export class Collection extends Resource {
 export class Entry extends Resource {
   constructor(client, representation, uri) {
     super();
-    this.lp_client = client;
-    this.uri = uri;
-    this.lp_attributes = {};
-    this.dirty_attributes = [];
+    this.init(client, uri);
```
bartaz: `lp_client` and `uri` will be inherited from `Resource` so will be non-enumerable.

src/server/launchpad/resources.js
```diff
@@ -75,10 +85,18 @@ export class Collection extends Resource {
...
+
+    // defining internal properties as non-enumerable not to leak any secrets
+    Object.defineProperty(this, 'lp_attributes', {
```
bartaz: Making local internal propeties non-enumerable, so they don't pollute serialised objects.

src/server/launchpad/resources.js
```diff
@@ -32,15 +42,15 @@ export class Resource {
 export class Root extends Resource {
   constructor(client, representation, uri) {
     super();
-    this.init(client, representation, uri);
+    this.init(client, uri, representation);
```
cjwatson: It might be a good idea to switch the order of `uri` and `representation` in these inherited constructors' signatures at some point to match the new `Resource` constructor signature, to reduce the potential for confusion.

PR #117, opened 2017-02-10
Redux actions/reducer to store snaps (enabled repositories)

bartaz:
Not used anywhere yet, but these actions allow to fetch user snaps from API.

### DIFF
--- a/src/common/actions/snaps.js
+++ b/src/common/actions/snaps.js
@@ -1,0 +1,55 @@
+import 'isomorphic-fetch';
+
+import { checkStatus, getError } from '../helpers/api';
+import { conf } from '../helpers/config';
+
+const BASE_URL = conf.get('BASE_URL');
+
+export const FETCH_SNAPS = 'FETCH_REPOSITORIES';
+export const FETCH_SNAPS_SUCCESS = 'FETCH_SNAPS_SUCCESS';
+export const FETCH_SNAPS_ERROR = 'FETCH_SNAPS_ERROR';
+
+export function fetchUserSnaps(owner) {
+  return (dispatch) => {
+    const url = `${BASE_URL}/api/launchpad/snaps/list?owner=${owner}`;
+
+    dispatch({
+      type: FETCH_SNAPS
+    });
+
+    return fetch(url, {
+      headers: { 'Content-Type': 'application/json' },
+      credentials: 'same-origin'
+    })
+      .then(checkStatus)
+      .then(response => {
+        return response.json().then(result => {
+          if (result.status !== 'success') {
+            throw getError(response, result);
+          }
+
+          dispatch(fetchSnapsSuccess(result.payload.snaps));
+
+        })
+        .catch(error => {
+          return Promise.reject(error);
+        });
+      })
+      .catch(error => dispatch(fetchSnapsError(error)));
+  };
+}
+
+export function fetchSnapsSuccess(snaps) {
+  return {
+    type: FETCH_SNAPS_SUCCESS,
+    payload: snaps
+  };
+}
+
+export function fetchSnapsError(error) {
+  return {
+    type: FETCH_SNAPS_ERROR,
+    payload: error,
+    error: true
+  };
+}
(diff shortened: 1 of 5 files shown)


src/common/reducers/snaps.js
```diff
@@ -0,0 +1,41 @@
+import * as ActionTypes from '../actions/snaps';
+import { parseGitHubRepoUrl } from '../helpers/github-url';
+
+export function snaps(state = {
+  isFetching: false,
+  success: false,
+  error: null,
+  snaps: null,
```
bartaz: Wasn't sure if we should be naming it `snaps` or maybe `enabledRepositories`.
Technically they are snaps with some repo info on them, and we will need more info about them as snaps (builds status, name registering, snapcraft.yaml), so I guess calling them snaps is not that confusing.

src/common/reducers/snaps.js
```diff
@@ -0,0 +1,41 @@
...
+      };
+    case ActionTypes.FETCH_SNAPS_SUCCESS:
+      return {
+        ...state,
+        isFetching: false,
+        success: true,
+        snaps: action.payload.map((snap) => {
+          return {
+            // parse repository info to keep consistent data format
+            repo: parseGitHubRepoUrl(snap.git_repository_url),
```
bartaz: Every snap on the list contains two pieces of information now:
```js
{
  repo: { fullName, owner, name, url } // full repo object as used elsewhere in the app
  snap: { git_repository_url, ... } // full snap data as returned from API
}
```

Maybe we don't need whole snap object here, but we will likely need some API URLs from there (to fetch build status, etc).

src/common/actions/snaps.js
```diff
@@ -0,0 +1,55 @@
...
+
+const BASE_URL = conf.get('BASE_URL');
+
+export const FETCH_SNAPS = 'FETCH_REPOSITORIES';
+export const FETCH_SNAPS_SUCCESS = 'FETCH_SNAPS_SUCCESS';
+export const FETCH_SNAPS_ERROR = 'FETCH_SNAPS_ERROR';
+
+export function fetchUserSnaps(owner) {
+  return (dispatch) => {
+    const url = `${BASE_URL}/api/launchpad/snaps/list?owner=${owner}`;
```
cjwatson: Use `encodeURIComponent(owner)`, or `qs.stringify`.

src/common/actions/snaps.js
```diff
@@ -0,0 +1,55 @@
...
+      type: FETCH_SNAPS
+    });
+
+    return fetch(url, {
+      headers: { 'Content-Type': 'application/json' },
+      credentials: 'same-origin'
+    })
+      .then(checkStatus)
+      .then(response => {
+        return response.json().then(result => {
```
cjwatson: I think we're standardising on parenthesising the signature part of arrow functions taking just one parameter.

src/common/reducers/snaps.js
cjwatson: I think `snaps` is fine.

src/common/reducers/snaps.js
cjwatson: Maybe a bit redundant since `repo` can be derived trivially from `snap.git_repository_url`?  You could call `parseGitHubRepoUrl` from a `mapStateToProps` function.  Then `snaps` really would just contain snap objects.

bartaz: I thought it would be worth to keep repo info parsed already not to do it in UI components too much, but it's likely that we will only have to do it once, so maybe it's a good point that we don't need it here.

I'll remove it and if at any point we see we do this parsing in many places we can always add it back.

src/common/actions/snaps.js
bartaz: Good catch, thanks.

src/common/actions/snaps.js
bartaz: Yes... and I remember me being the one mentioning it ;)

PR #119, opened 2017-02-10
100: Batched repository selection

katiefenn:
- Added initial support for batched snap creation to createSnap action
  creators
- Changed redirect on successful login to "select repositories"
  dashboard view

### DIFF
--- a/src/common/actions/create-snap.js
+++ b/src/common/actions/create-snap.js
@@ -2,7 +2,6 @@ import 'isomorphic-fetch';
 
 import { checkStatus, getError } from '../helpers/api';
 import { conf } from '../helpers/config';
-import { parseGitHubRepoUrl } from '../helpers/github-url';
 
 const BASE_URL = conf.get('BASE_URL');
 
@@ -54,10 +53,11 @@ function getSnapName(owner, name) {
     }));
 }
 
-export function createSnap(repositoryUrl, location) { // location for tests
+export function createSnap(repository) {
   return (dispatch) => {
+    const repositoryUrl = repository.url;
     if (repositoryUrl) {
-      const { owner, name, fullName } = parseGitHubRepoUrl(repositoryUrl);
+      const { owner, name, fullName } = repository;
 
       dispatch({
         type: CREATE_SNAP,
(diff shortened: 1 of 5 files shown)


src/common/actions/create-snap.js
```diff
@@ -54,10 +53,11 @@ function getSnapName(owner, name) {
     }));
 }
 
-export function createSnap(repositoryUrl, location) { // location for tests
+export function createSnap(repository) { // location for tests
```
katiefenn: I discovered that all code calling createSnap already had a parsed repository object. Instead of parsing it twice I decided to just pass the repository object.

src/common/actions/create-snap.js
```diff
@@ -81,30 +81,45 @@ export function createSnap(repositoryUrl, location) { // location for tests
         .then(checkStatus)
         .then(response => {
           return response.json().then(result => {
-            if (result.status !== 'success' ||
-                result.payload.code !== 'snap-created') {
-              throw getError(response, result);
+            if (result.status !== 'success' || result.payload.code !== 'snap-created') {
```
katiefenn: Unrelated codestyle change.

src/common/actions/create-snap.js
```diff
@@ -81,30 +81,45 @@ export function createSnap(repositoryUrl, location) { // location for tests
...
+              return Promise.reject(getError(response, result));
```
katiefenn: `createSnap` is now called by `createSnaps`, so instead of handling the success and failure scenarios here they are now handled by the `createSnaps` promise.

src/common/actions/create-snap.js
```diff
@@ -81,30 +81,45 @@ export function createSnap(repositoryUrl, location) { // location for tests
...
-            (location || window.location).href = `${BASE_URL}/${fullName}/builds`;
-          } else {
-            dispatch(createSnapError(fullName, error));
-          }
         });
     }
   };
 }
 
+export function createSnaps(repositories, location) {
```
katiefenn: This action will eventually support batch creation of snaps. This is currently blocked in the LP auth API.

Until then, I've prepared this code to accept multiple repositories, but it actually only creates a snap for the first selected repository. This allows us work on the UI without breaking the staging website.

src/server/handlers/github-auth.js
```diff
@@ -66,7 +66,7 @@ export const verify = (req, res, next) => {
     logger.info('User successfully authenticated');
 
     // Redirect to logged in URL
-    res.redirect('/dashboard');
+    res.redirect('/dashboard/select-repositories');
```
katiefenn: After logging in, redirect to the dashboard select-repositories view as if logging in for the first time.

katiefenn: We don't technically need this change. Eventually, the user will only be redirected to the select-repositories view if they have no enabled repositories.

The reason I made this change now is that it will allow us to validate the "initial login" path before the smarter redirect is in-place. Our designers can see a more complete picture of the site before the "enabled-repositories" dashboard view is finished.

It's a small, meaningless change but I think it speaks a lot.

src/common/actions/create-snap.js
cjwatson: The "location for tests" comment is now misplaced - it should be moved to `createSnaps`, since that's where the `location` parameter now lives.

src/common/components/select-repository-list/index.js
```diff
@@ -70,7 +70,7 @@ class SelectRepositoryList extends Component {
     const { selectedRepos } = this.props.selectRepositoriesForm;
     if (selectedRepos.length) {
       // TODO: Switch to batched repository action
```
cjwatson: This TODO comment is obsolete now, isn't it?  That is, we still need changes inside `createSnaps` but not here.

src/common/actions/create-snap.js
katiefenn: Arg, nice spot 😅

src/server/handlers/github-auth.js
bartaz: OK. 

We probably need full workflow (with usable dashboard and enabled repos list and 'smarter' redirect) before it's meaningful to discuss the complete picture, but it doesn't hurt to have this for now.

PR #120, opened 2017-02-10
Getting GitHub user info to the app

bartaz:
API route and redux actions/reducer to fetch current user data from GH.

GH OAuth verify only sends token, not any other user info, so we need to fetch it from API.
It's not perfect when we do it from client side, because it means we know user is authenticated (from session) but still have to wait for user data API call.

I guess other option would be to do the API call on server side (in verify handler), and keep user data in session once we get it from GH (and only then redirect back to app, with user data injected into state).

### DIFF
--- a/src/common/containers/app.js
+++ b/src/common/containers/app.js
@@ -5,7 +5,23 @@ import Helmet from 'react-helmet';
 import Header from '../components/header';
 import Footer from '../components/footer';
 
+import { fetchUser } from '../actions/user';
+
 export class App extends Component {
+  fetchData(props) {
+    if (props.auth.authenticated && !props.user.user && !props.user.isFetching) {
+      this.props.dispatch(fetchUser());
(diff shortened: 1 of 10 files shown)


src/common/containers/app.js
```diff
@@ -5,7 +5,23 @@ import Helmet from 'react-helmet';
 import Header from '../components/header';
 import Footer from '../components/footer';
 
+import { fetchUser } from '../actions/user';
+
 export class App extends Component {
+  fetchData(props) {
+    if (props.auth.authenticated && !props.user.user && !props.user.isFetching) {
```
bartaz: App container fetches user data when user is authenticated.

Not an ideal solution as we need to wait for request to finish.
It's for example now visible in header, that we render page with Logout and Dashboard links, but user name shows up some time later (after we get data from GH).

PR #121, opened 2017-02-12
Add routes to retrieve SSO discharge from session

cjwatson:
Nothing stores a discharge macaroon in the session yet, but this will be
useful once we've reworked our authorisation workflow to use a
package_upload_request macaroon.

I suspect the new actions won't actually work in browsers yet since
macaroons.js needs a little bit of polyfilling, but that'll be easier to
test and fix up once a few more pieces are in place.  It's harmless for
now since the new actions aren't used yet.


src/server/handlers/universal.js
```diff
@@ -52,6 +52,12 @@ export const handleMatch = (req, res, error, redirectLocation, renderProps) => {
       return;
     }
 
+    if (req.session.ssoDischarge) {
+      // Tell the client that it can pick up a discharge macaroon from the
+      // session and move it to local storage.
+      initialState.authStore = { hasDischarge: true };
```
bartaz: It seems that on the client we will be fetching ssoDischarge (via an action).
Can't we just set ssoDischarge here for client in initialState, so that it doesn't have to be fetched at all?
Maybe even, once sent to client in initialState it can be just immediately deleted from session (because client will already have it)?

cjwatson: I thought about this, but it doesn't quite work because we need to prove that the client has had a chance to save it to local storage before deleting it.

We could perhaps push it into the initial state, it's true, but I have a slight preference for secrets not showing up in view-source, which is perhaps irrational in this case ...

PR #127, opened 2017-02-13
Use package_upload_request macaroons for authz

cjwatson:
Creating snaps now requires being "signed into the store", which equates
to having a discharged package_upload_request macaroon in local storage.
For each snap to be created, we request a specific non-expiring
package_upload macaroon for it and tell the backend to store that in
Launchpad.  This allows batch-creation of snaps, and avoids a round-trip
through SSO per snap.

This leaves some unused endpoints around, but this PR is already large
enough; they can be cleaned up later.

The UI is pretty weak in places, but this should be enough to unblock other development (in particular, the `package_upload_request` macaroon can be reused for registering a snap name, after lifting up the `authStore` state to some suitable common ancestor component, and it should now be possible to put the rest of the dashboard workflow in place).

This cannot be landed until SCA >= r3802 is deployed to production.

### DIFF
--- a/src/common/actions/create-snap.js
+++ b/src/common/actions/create-snap.js
@@ -1,4 +1,6 @@
 import 'isomorphic-fetch';
+import localforage from 'localforage';
+import { MacaroonsBuilder } from 'macaroons.js';
 
 import { checkStatus, getError } from '../helpers/api';
 import { conf } from '../helpers/config';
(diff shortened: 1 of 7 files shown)


bartaz:
@katiefenn as this one touches repositories list (and shows how to use new auth actions) you should probably have a look at this for your dashboard work.

While r3802 is not deployed yet in SCA and we shouldn't merge it, you can at least be prepared for these updates.

PR #128, opened 2017-02-13
User in session

bartaz:
Follow up to #120 to actually store user data in session and avoid waiting for GH API response in client.


src/common/containers/app.js
```diff
@@ -35,7 +19,7 @@ export class App extends Component {
         />
         <Header
           authenticated={this.props.auth.authenticated}
-          user={this.props.user.user}
+          user={this.props.user}
```
bartaz: `user` data is not nested in reducer anymore (as we don't need any `isFetching` etc).

src/server/handlers/github-auth.js
```diff
@@ -60,13 +61,24 @@ export const verify = (req, res, next) => {
...
+        if (response.statusCode !== 200) {
+          return next(new Error(`Authentication failed. ${response.body.message}`));
+        }
+
+        logger.info('User info successfully fetched');
+
+        // Save auth token and user info to session
+        req.session.githubAuthenticated = true;
+        req.session.token = body.access_token;
+        req.session.user = response.body;
```
bartaz: Both token and user data are stored in session.

PR #131, opened 2017-02-14
Added mock to list user snaps

bartaz:


mocks/service/launchpad/index.js
```diff
@@ -15,6 +15,34 @@ router.post('/+snaps', (req, res) => {
   }
 });
 
+// we use * instead of + because router treats + as regexp pattern
+// and doesn't actually match +snaps
+router.get('/devel/*snaps', (req, res) => {
```
bartaz: It turns out router treats `+` in path as regexp pattern, so using `/devel/+snaps` was not properly matching.
Using `*` solves this issue as it simply allows any characters (which may lead to some issues if we have other paths ending with `snaps`, but it's not the case right now).

Now it makes me wonder if any of the other paths in this mock actually work, as they all have `+snaps`.
But I suppose they are outdated anyway, so if we want to go any further with LP mocks (auth, creating snaps), these will need to be updated too.

Issue #133, opened 2017-02-14
/login/failed is an inaccessible universal route


earnubs:
`/login/failed` is a universal route, but the user won't be able to reach it because the universal route handler redirects anyone without a session and team to login authenticate (https://github.com/canonical-ols/build.snapcraft.io/blob/master/src/server/handlers/universal.js#L33), which looks from the user point of view like nothing happened on sign in (user is quickly redirected from sso > bsi > sso)

cjwatson:
The teams bit isn't quite as you describe it.  The reduced conditional is like this:

    const teams = JSON.parse(conf.get('OPENID_TEAMS') || 'null');
    ...
    if (!req.session && teams && teams.length) {
      res.redirect(302, '/login/authenticate');
    }

So it's not that the user is without a team, but that team membership information is expected from SSO.  It might be phrased as "if we need team membership information, and we don't have a session, then redirect".

I'm not sure how this could happen from SSO auth without the session being totally broken in some way, because it only redirects if there is a session (https://github.com/canonical-ols/build.snapcraft.io/blob/master/src/server/handlers/login.js#L130-L132).

It's probably not vital that we track this down, since the team-membership front-door is only temporary and will be removed next week anyway.

earnubs:
Sorry, yes, I misread.  #135, was the real issue.

PR #135, opened 2017-02-14
allow setting secure cookies to false

earnubs:
I got stuck in a sso loop, I've added a bug for part of it (can't see login/failed #133), but this I think fixes the loop bit (cookies not being set in http, therefore you are simply redirected to sso for all eternity) ... and now I've pushed I see why no-one else sees this, because they don't have `COOKIE_SECURE` in the env file _at all_ whereas I have it as `false`

### DIFF
--- a/src/server/helpers/session.js
+++ b/src/server/helpers/session.js
@@ -9,6 +9,7 @@ const SESSION_DEFAULTS = {
   name : 'sessionId',
   cookie: {
     maxAge: 2 * 60 * 60 * 1000, // 2 hours, macaroon life span
+    secure: false
   },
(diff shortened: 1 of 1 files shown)


src/server/helpers/session.js
```diff
@@ -9,6 +9,7 @@ const SESSION_DEFAULTS = {
   name : 'sessionId',
   cookie: {
     maxAge: 2 * 60 * 60 * 1000, // 2 hours, macaroon life span
+    secure: false
```
earnubs: `false` is the default, but it doesn't hurt to be explicit...

PR #137, opened 2017-02-15
4: Enabled repos list

katiefenn:
- Changed RepositoriesList to show snaps instead of repositories
- Removed owner parameter from fetchUserSnaps action creator
- Added new Table component based on Vanilla patterns
- Added new "neutral" button style based on latest Vanilla patterns


mocks/service/launchpad/index.js
```diff
@@ -29,12 +29,18 @@ router.get('/devel/*snaps', (req, res) => {
...
         {
           resource_type_link: `${base}/#snap`,
           git_repository_url: `${prefix}mock-repo-2`,
-          self_link: `${base}/${owner}/+snap/mock-snap-1`
+          self_link: `${base}/${owner}/+snap/mock-snap-2`,
+          name: 'mock-snap-2'
+        },
+        {
+          resource_type_link: `${base}/#snap`,
+          git_repository_url: `${prefix}mock-repo-3`
```
katiefenn: Not sure what the response data looks like for a snap that has not been registered. I've made a guess here. Needs clarification.

(cc: @cjwatson)

src/common/actions/snaps.js
```diff
@@ -9,16 +9,15 @@ export const FETCH_SNAPS = 'FETCH_REPOSITORIES';
 export const FETCH_SNAPS_SUCCESS = 'FETCH_SNAPS_SUCCESS';
 export const FETCH_SNAPS_ERROR = 'FETCH_SNAPS_ERROR';
 
-export function fetchUserSnaps(owner) {
+export function fetchUserSnaps() {
```
katiefenn: I've removed the `owner` parameter because this can be added to the request in the service layer. I have changed the Express handler for this API endpoint to add the owner from the session if it is not explicitly sent in the request.

src/common/components/repositories-home/index.js
```diff
@@ -1,65 +1,50 @@
...
-      <Step number="1" complete={ authenticated }>
-        <Anchor href="/auth/authenticate">Log in with GitHub</Anchor>
-      </Step>
-    );
-  }
-
-  step2() {
-    return (
-      <Step number="2">
-        Choose one of your repositories
```
katiefenn: All this can go now that the older Step 1 ... Step 2 layout for the Cape Town demo has been dispensed with.

mocks/service/launchpad/index.js
cjwatson: Launchpad neither knows nor cares about snap name registration (that's all in SCA), so whether the name has been registered makes no difference to the response here.

src/common/actions/snaps.js
cjwatson: As I said on IRC yesterday, I think this is a mistake.  While currently it's true that the owner here is always identical to the logged-in user, this will need to stop being the case as soon as we implement organisation support.

src/common/components/repositories-list/index.js
```diff
@@ -5,9 +5,9 @@ import { conf } from '../../helpers/config';
 import { createSnap } from '../../actions/create-snap';
```
katiefenn: This component needs a new name, because it doesn't show repositories. It shows snaps.

How does everyone feel about `SnapList`? (cc: @cjwatson @bartaz @earnubs)

src/common/components/repositories-list/index.js
```diff
@@ -40,18 +48,12 @@ class RepositoriesList extends Component {
...
 
-  renderRepository(repo) {
-    const { fullName } = repo;
-    const status = this.props.repositoriesStatus[fullName] || {};
-
+  renderRow(snap) {
+    const { fullName } = parseGitHubRepoUrl(snap.git_repository_url);
     return (
       <RepositoryRow
-        key={ `repo_${repo.fullName}` }
```
katiefenn: Ditto `RepositoryRow`.

src/common/components/repositories-list/index.js
```diff
@@ -62,57 +64,51 @@ class RepositoriesList extends Component {
...
-    const pageLinks = this.props.repositories.pageLinks;
+    const isLoading = this.props.snaps.isFetching;
 
     return (
       <div>
-        { this.props.repositories.success && pageLinks &&
-          <div className={ styles['page-links-container'] }>
-            <PageLinks { ...pageLinks } onClick={ this.onPageLinkClick.bind(this) } />
-          </div>
-        }
```
katiefenn: I've removed pagelinks because they were built for pagination of results from the GitHub API, not the Launchpad API. Pagination of these results was agreed as a stretch goal.

src/common/components/repositories-list/index.js
```diff
@@ -62,57 +64,51 @@ class RepositoriesList extends Component {
...
+              <TH>Registered for publishing</TH>
+              <TH>Latest build</TH>
+            </TR>
+          </THead>
+          <TBody>
+            { this.props.snaps.success &&
+              this.props.snaps.snaps.map(this.renderRow.bind(this))
+            }
+          </TBody>
+        </Table>
```
katiefenn: This is my interpretation of a Vanilla table. I saw no point in giving components names that diverged from the existing table elements because tables are defined by their structure as well as their content.

Opinions welcome.

src/common/components/repository-row/index.js
```diff
@@ -1,46 +1,46 @@
...
-    onButtonClick
+    snap
   } = props;
 
+  const { fullName } = parseGitHubRepoUrl(snap.git_repository_url);
+
+  return (
+    <TR>
+      <TD><a href={ `/${fullName}/builds` }>{ fullName }</a></TD>
+      <TD></TD>
```
katiefenn: This column is for the "configured" status (the existance of a valid snapcraft.yml file). I require clarification of how to obtain this information for one or more repositories at once in order to populate this status for all repos in this table.

src/common/components/repository-row/index.js
```diff
@@ -1,46 +1,46 @@
...
+      <TD>{ renderSnapNameLabel(snap) }</TD>
+      <TD></TD>
```
katiefenn: This column is for the "Latest build" status. I require clarification of how to obtain this information for one or more repositories at once in order to populate this status for all snaps in this table. (cc: @cjwatson)

src/common/components/vanilla/button/button.css
```diff
@@ -12,6 +12,11 @@ $positive-button-color-hover: #0a5615;
 $positive-button-text-color: #FFF;
 $positive-border-color: $success;
 
+$neutral-button-color: #FFF;
```
katiefenn: New "neutral" button style to match wireframes.

src/common/components/repository-row/index.js
cjwatson: There's an `/api/github/snapcraft-yaml/:owner/:name` route that you should use for this.

To the best of my knowledge, the GitHub API only lets us query this for a single repository, so we'll have to do this with one request per repository, perhaps in batches with `Promise.all`.  We should make the backend route cache positive responses in memcached (I'll try to do that today if I have time, but no promises).

src/common/components/repository-row/index.js
cjwatson: This would currently require calling the `/api/launchpad/builds` route once per repository and picking out the most recent one.

At present, this will be very inefficient.  I propose that we omit this column for MWC until we have a chance to work out how to get this information efficiently (perhaps by augmenting the Launchpad `Snap` class with an attribute representing the latest build date which we could populate in bulk from the query used via `/api/launchpad/snaps/list`).

src/common/actions/snaps.js
katiefenn: Should this be the responsibility of the client or the API?

If we add organisation support to the application, should the API not supply all snaps the user has access to? It's difficult for me to judge without knowing what organisation support will look like.

src/common/components/repositories-list/index.js
cjwatson: I think it shows repositories that have had snap building enabled, not snaps as such.

src/common/actions/snaps.js
cjwatson: I'm not sure yet; that's why I'd propose leaving this be until we have a chance to work that out, rather than having code churn that may turn out to be unnecessary.

katiefenn: Okay, will fix.

src/common/components/repositories-list/index.js
katiefenn: @cjwatson: Good point. Let's leave it as is.

mocks/service/launchpad/index.js
katiefenn: Okay. Is there a way we can pull snap names from SCA in batch, or would this be subject to the same efficiency issues as the other calls?

cjwatson: No.  They'll have a `name` field, but that's just a unique URL segment which in this case is generated by BSI, and doesn't relate to the name in `snapcraft.yaml`.

We're explicitly intending to create the snap object in Launchpad before a snap name is necessarily known, so you should expect that repositories in this response may well not have a registered snap name as part of the normal workflow.

src/common/components/repositories-list/index.js
bartaz: My worry with using HTML table here we will get to the point where it will be hard to do some styling we need because of limitations of how tables are rendered.

It may show up really soon when we get to the point where we need to expand the row (for name registering and snapcraft.yaml info).

The suggested pattern is using MAAS tables (http://ubuntudesign.github.io/maas-gui-vanilla-theme/index.html#patterns) and if you go to 'Table - Interactive' section you will see that they actually use `div` for these highly interactive tables.

src/common/components/repositories-list/index.js
bartaz: It's a fullName of repository rather then a snap, but - whatever, it's internal React key anyway.

mocks/service/launchpad/index.js
katiefenn: Okay, perhaps it would be better for me to leave the status fields blank until integrations are complete?

cjwatson: The Launchpad snaps route is used regardless, so I think it's still sensible to have a working mock for it.  It just has nothing particular to do with registered snap names, since it's at the wrong layer.

cjwatson: Well.  I should clarify.  A snap where BSI has worked out the snap name will have the `store_name` field set; that's up to BSI to fill in.  From Launchpad's point of view there's nothing to say that it must be registered, although I expect once we complete the integration we'll be able to guarantee that it will be.

From the mock point of view, though, filling in as much as we have so far is just fine.

katiefenn: Okay, acknowledged keeping the mock.

What about removing the "configured", "registered for publishing" and "latest build" columns from the enabled repos table until we have integrations?

bartaz: Build status can be a plain link to repo's builds page for now.

I guess it would be good to have placeholders for the others too already. We need them anyway, and we need them soon.

src/common/components/repositories-list/index.js
katiefenn: Thanks for pointing this out. I'll re-write this based on the interactive table from MAAS. I'm surprised this isn't in Vanilla.

src/common/actions/snaps.js
```diff
@@ -9,15 +9,16 @@ export const FETCH_SNAPS = 'FETCH_REPOSITORIES';
 export const FETCH_SNAPS_SUCCESS = 'FETCH_SNAPS_SUCCESS';
 export const FETCH_SNAPS_ERROR = 'FETCH_SNAPS_ERROR';
 
-export function fetchUserSnaps() {
+export function fetchUserSnaps(owner) {
```
katiefenn: Restored owner parameter to `fetchUserSnaps` action creator.

src/common/components/repositories-list/index.js
```diff
@@ -15,15 +15,6 @@ import { spinner as spinnerStyles } from '../../containers/container.css';
...
 class RepositoriesList extends Component {
-
-  componentDidMount() {
-    const { authenticated } = this.props.auth;
-
-    if (authenticated) {
-      this.props.dispatch(fetchUserSnaps());
-    }
-  }
-
```
katiefenn: `fetchUserSnaps` is also dispatched in component `RepositoriesHome`, so I removed the duplicate.

src/common/components/repositories-list/index.js
```diff
@@ -52,7 +43,7 @@ class RepositoriesList extends Component {
     const { fullName } = parseGitHubRepoUrl(snap.git_repository_url);
     return (
       <RepositoryRow
-        key={ `snap_${fullName}` }
+        key={ `repo_${fullName}` }
```
katiefenn: Renamed this on feedback from @cjwatson and @bartaz.

src/common/components/repositories-list/index.js
```diff
@@ -75,19 +66,19 @@ class RepositoriesList extends Component {
...
-          <TBody>
+          <Head>
+            <Row>
+              <Header col="30">Name</Header>
+              <Header col="20">Configured</Header>
+              <Header col="20">Registered for publishing</Header>
+              <Header col="30">Latest build</Header>
+            </Row>
+          </Head>
+          <Body>
```
katiefenn: I've replaced the Vanilla-style table with a MAAS-style interactive table. I've adjusted the names of the components to match the naming scheme of the classes used by the MAAS theme.

src/common/components/vanilla/table-interactive/body.js
```diff
@@ -0,0 +1,14 @@
+import React, { PropTypes } from 'react';
+import styles from './table.css';
+
+export default function Body(props) {
+  return (
+    <main className={ styles.body }>
```
earnubs: Why not use table elements?

bartaz: GH is swallowing comments on outdated diffs, but we discussed it here:
https://github.com/canonical-ols/build.snapcraft.io/pull/137#discussion_r101288714

Styling table elements may get tricky, especially that we want them to expand to allow showing additional info and register name form.

src/common/components/repositories-list/index.js
bartaz: From code readability point of view I would prefer it to be `width="30%"` to be more explicit.
Only because these numbers add up to 100 I realised they are probably meant to be percent.

But it's not something I would block this PR on.

PR #138, opened 2017-02-15
Remove code related to old authz workflow

cjwatson:
These endpoints are no longer used with the new approach using a
package_upload_request macaroon.


mocks/service/launchpad/index.js
```diff
@@ -44,10 +44,7 @@ router.get('/devel/*snaps', (req, res) => {
 });
 
 router.post('/~:owner/+snap/:name', (req, res) => {
```
bartaz: Does this work for you (I mean using `+` in path)?
I had issues yesterday with trying to mock `/devel/+snaps` because router treasts `+` as  in regexp. (See: https://github.com/cjwatson/build.snapcraft.io/blob/9d4087eaf03a71c89f7c5ff59079c9e2d6457987/mocks/service/launchpad/index.js#L18-L20)

cjwatson: Good point.  I've fixed all the routes in this file (the need for the doubled backslash is documented in https://www.npmjs.com/package/path-to-regexp#custom-match-parameters).

bartaz: Ah, double backslash... I didn't dig that deep into the docs.
Thanks for fixing it!

PR #140, opened 2017-02-15
Make the authError state be consistently an object

cjwatson:
Otherwise we get this when navigating directly to `/login/failed`
(though not when being sent there due to an error, since `handleMatch`
sets `authError` to an object in the initial state):

    Warning: Failed prop type: Invalid prop `authError` of type `array`
    supplied to `LoginFailed`, expected `object`.
        in LoginFailed (created by Connect(LoginFailed))
        in Connect(LoginFailed) (created by RouterContext)
        in div (created by App)
        in App (created by Connect(App))
        in Connect(App) (created by RouterContext)
        in RouterContext
        in Provider (created by Html)

### DIFF
--- a/src/common/reducers/auth-error.js
+++ b/src/common/reducers/auth-error.js
@@ -1,9 +1,12 @@
 import { AUTH_ERROR } from '../actions/auth-error';
 
-export function authError(state = [], action) {
+export function authError(state = {}, action) {
   switch (action.type) {
     case AUTH_ERROR:
-      return [...state, { message: action.message }];
+      return {
+        ...state,
+        message: action.message
+      };
     default:
       return state;
   }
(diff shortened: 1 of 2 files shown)


src/common/reducers/auth-error.js
```diff
@@ -1,9 +1,12 @@
 import { AUTH_ERROR } from '../actions/auth-error';
 
-export function authError(state = [], action) {
+export function authError(state = {}, action) {
```
bartaz: I guess the array initially came from MU (https://github.com/canonical-ols/my.ubuntu.com/blob/master/src/common/reducers/oyez.js#L3) - where we possibly show multiple status messages rendered by server. So the idea here was to support showing multiple error messages.

But as long as it fixed the issue we have now I guess it's good to consistently treat `authError` as an object.

PR #145, opened 2017-02-15
Clear macaroon from local storage when signing out

### LINKED ISSUE #125: Explicitly clear local storage when signing out
The macaroon in local storage will expire after a couple of hours anyway, but signing out should explicitly clear it.  I'm not sure where best to do this - the "Log out" link is currently to `/auth/logout` which is server-rendered and therefore can't do this, so it will presumably be a bit involved.

cjwatson:
Fixes #125.


src/common/actions/auth-store.js
```diff
@@ -231,3 +233,23 @@ function checkSignedIntoStoreError(error) {
     error: true
   };
 }
+
+export function signOutOfStore(location) { // location for tests
```
bartaz: The `signOutOfStore` name may be a bit misleading, as it not only 'signs out of the store' (by clearing macaroon from localStorage), but it also redirects to logout route which clears session (signing out user from BSI completely by clearing GH token).

So I guess simply `signOut` would be more true.

src/common/reducers/auth-store.js
```diff
@@ -3,6 +3,7 @@ import * as ActionTypes from '../actions/auth-store';
 export function authStore(state = {
   isFetching: false,
   hasDischarge: false,
+  signingOut: false,
```
bartaz: Do we need that information?
It doesn't seem that we use it anywhere, so I don't think it's necessary it add.

PR #146, opened 2017-02-16
Storing builds for multiple repositories in redux store

bartaz:
Not as elegant as I hoped, but this extends `snapBuilds` reducer to allow storing builds for many repositories (keyed by repo full name).

Updated relevant views (repository setup, builds list and build details).
Currently not connected to new dashboard.

### DIFF
--- a/src/common/actions/snap-builds.js
+++ b/src/common/actions/snap-builds.js
@@ -3,6 +3,8 @@ import 'isomorphic-fetch';
 import { checkStatus } from '../helpers/api';
 import { conf } from '../helpers/config';
 
+import { parseGitHubRepoUrl } from '../helpers/github-url';
+
 const BASE_URL = conf.get('BASE_URL');
 
 export const FETCH_BUILDS = 'FETCH_BUILDS';
@@ -18,27 +20,38 @@ const REQUEST_POST_OPTIONS = {
   credentials: 'same-origin'
 };
 
-export function fetchBuildsSuccess(builds) {
+export function fetchBuildsSuccess(id, builds) {
   return {
     type: FETCH_BUILDS_SUCCESS,
-    payload: builds
+    payload: {
+      id,
+      builds
+    }
   };
 }
 
-export function fetchBuildsError(error) {
+export function fetchBuildsError(id, error) {
   return {
     type: FETCH_BUILDS_ERROR,
-    payload: error,
+    payload: {
+      id,
+      error
+    },
     error: true
   };
 }
 
 // Fetch snap info (self_link) for given repository
 export function fetchSnap(repositoryUrl) {
+  const { fullName } = parseGitHubRepoUrl(repositoryUrl);
+
   return (dispatch) => {
     if (repositoryUrl) {
       dispatch({
-        type: FETCH_BUILDS
+        type: FETCH_BUILDS,
+        payload: {
+          id: fullName
+        }
       });
 
       repositoryUrl = encodeURIComponent(repositoryUrl);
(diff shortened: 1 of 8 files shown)


src/common/actions/snap-builds.js
```diff
@@ -50,39 +63,52 @@ export function fetchSnap(repositoryUrl) {
...
         })
-        .catch((error) => dispatch(fetchBuildsError(error)));
+        .catch((error) => dispatch(fetchBuildsError(fullName, error)));
     }
   };
 }
 
 // Fetch builds list for given snap
-export function fetchBuilds(snapLink) {
+export function fetchBuilds(repositoryUrl, snapLink) {
```
bartaz: We need both `repositoryUrl` and `snapLink` now to fetch builds, because we need to store them per-repo.

In theory `repositoryUrl` should be enough, as we should already have snapLink for it in the store, but I'm not sure if we want actions to read the store themselves.

src/common/components/build-history/index.js
```diff
@@ -29,15 +30,19 @@ export const BuildHistory = (props) => {
...
 };
 
-function mapStateToProps(state) {
-  const builds = state.snapBuilds.builds;
-  const success = state.snapBuilds.success;
+function mapStateToProps(state, ownProps) {
+  const { fullName } = ownProps.repository;
+
+  // get builds for given repo from the store or set default empty values
+  const repoBuilds = state.snapBuilds[fullName] || snapBuildsInitialStatus;
```
bartaz: Because we store builds per repository the 'initial state' per repo is undefined (until anything is stored in the store for given repo), so when reading data from the store, we need to provide some defaults.

src/common/reducers/snap-builds.js
```diff
@@ -1,39 +1,58 @@
 import * as ActionTypes from '../actions/snap-builds';
 import { snapBuildFromAPI } from '../helpers/snap-builds';
 
-export function snapBuilds(state = {
+export const snapBuildsInitialStatus = {
```
bartaz: Exporting initial status for individual repository builds, because they will be undefined in the store until any action creates entry for given repo. So now we can use `snapBuildsInitialStatus` as defaults where we need to.

src/common/actions/snap-builds.js
```diff
@@ -18,27 +20,38 @@ const REQUEST_POST_OPTIONS = {
...
-    payload: builds
+    payload: {
+      id,
+      builds
+    }
   };
 }
 
-export function fetchBuildsError(error) {
+export function fetchBuildsError(id, error) {
```
katiefenn: Is there a need to store information about fetching a specific build?

We are fetching build data separately because of an implementation detail: we only have an API endpoint for fetching data on builds singularly, not in batch. This means that anything can go wrong with any of the API requests, but is there a feature in our application that relies on the success or failure of individual requests?

bartaz: For now it's the simplest way to go.

We will likely fetch this in parallel for the enabled repositories, so it's easier to just keep whole status separately for each repo, so they don't affect each other. For example error fetching build status of one repo doesn't mean we can't have successfully fetched build status for other repo and show it in UI.

I'm fine with discussing how we further go with storing such data, but what we need right now this seems enough.

katiefenn: Like I said: does it actually matter if we have a problem with just one request? It's more likely that we will have an issue for all or none of them, and we can save a bit of complexity here.

bartaz: Of course, if LP goes down we will likely get same error for all.

Keeping the statuses separate in the store at least makes sure they don't affect each other.
For parallel requests it makes sure at least that error from one repo will not affect success from the other. Or maybe more importantly - error from one repo is not overridden by success response from the other (that comes back later).

I'd prefer to stick to this working solution now and discuss it further when we have more time for that (and any refactoring that may be needed).

katiefenn: Ack 👍

src/common/actions/snap-builds.js
```diff
@@ -18,27 +20,38 @@ const REQUEST_POST_OPTIONS = {
...
+  const { fullName } = parseGitHubRepoUrl(repositoryUrl);
+
   return (dispatch) => {
     if (repositoryUrl) {
       dispatch({
-        type: FETCH_BUILDS
+        type: FETCH_BUILDS,
+        payload: {
+          id: fullName
+        }
```
katiefenn: If I understand this correctly, `FETCH_BUILDS` is called to clear any builds in the store before fetching others?

Why not have an action called `CLEAR_BUILDS` that spells out what is happening?

bartaz: It doesn't clear the builds. It only changes `isFetching` to true.

```js
     case ActionTypes.FETCH_BUILDS:
        return {
          ...state,
         [payload.id]: {
           ...initialStatus,
           ...state[payload.id], // take previous status (keeping stored builds)
           isFetching: true
         }
        };
```

We don't clear builds before fetching, because on builds listing we fetch every 15 seconds, so we don't want them to disappear during loading.

src/common/components/build-history/index.js
katiefenn: We could address this by binding to the store in the parent container (`containers/builds`) and reducing builds to the builds for one selected snap. This prop could then be passed in JSX instead of using connect.

src/common/actions/snap-builds.js
katiefenn: Okay 👍

PR #147, opened 2017-02-16
Include snap name in /api/github/repos response

cjwatson:
Positive responses are cached to avoid an unreasonable number of
requests, and the webhook clears the cache so that we notice changes
more quickly at least for enabled repositories.

I haven't tested this in a browser since this was mostly done in spare minutes at the hospital; somebody not on leave might want to give this a quick run to make sure it doesn't explode or anything.


bartaz:
@cjwatson Just an afterthought - wouldn't it make more sense to include snap name in /api/launchpad/snaps/list rather than /api/github/repos?

Right now we check for snapcraft.yaml for every single user repository when we don't really need that. We only need snap name (or snapcraft.yaml status) for repositories that user enabled.
It seem that right now when showing listing of all repos response from github is unnecessary delayed with checks for snapcraft.yaml.

cjwatson:
Yes, I think you're probably right here.  Let's sort this out early next
week.

PR #152, opened 2017-02-17
Expand local development instructions

nottrobin:
- Update the README to include accurate local development instructions, including how to connect to GitHub itself.

QA
==

View [the README on my branch](https://github.com/nottrobin/build.snapcraft.io/blob/readme-linting-updates/README.md), check it reads okay.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,30 @@
 # build.snapcraft.io
+
 Snapcraft Build site.
 
-## Service layer mock
-A mock service is bundled with this project to replace the GitHub API when running on development platforms. To start the mock service, run the following command:
+## Local development
+
+First, make sure all dependencies are installed:
+
+``` bash
+npm install
+```
+
+### Start the dev server
+
+To start the development server:
+
+``` bash
+npm start
+```
+
+After about 20 seconds, the web application should be available at <http://127.0.0.1:8000>.
+
+### Service layer mock
+
+To actually test out the service, the app needs to connect to a GitHub-like API. By default, the development server is set up to look for the mock API service at <http://localhost:4000>.
+
+You can start the mock service as follows:
 
 ```
 npm run start-mock-service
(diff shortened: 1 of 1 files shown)


bartaz:
We are using ESLint for linting the code.

So to avoid keeping and maintaining setup for 2 different linters maybe we can tell jshint to ignore error in this project?

nottrobin:
If you're already using a linter, I'm not suggesting you change. It's just that I am using the jshint plugin for Atom, and I was getting a large number of errors in my editor - as my editor has no way to know which ES flavour this is. I was hoping the addition of a 3-line hidden file, to accommodate this local development setup might be okay.

If you disagree and think it pollutes the repo too much, I'm happy to remove it.

nottrobin:
@bartaz ah yes I see. We shouldn't maintain specific linting rules in two places. I'll update the config to ignore all errors.

bartaz:
@nottrobin I don't mind the file if it helps. But from you description it seems that just adding `esversion: 6` only reduces number of error (from 1500 to 400) - so it seems it doesn't fully solve the issue and some more config changes would be needed to make jshint fully happy.

I'm just suggesting that maybe it's possible to tell jshint to ignore the errors in the project.

nottrobin:
@bartaz yes sorry I see what you mean now. I also found this option: https://github.com/AtomLinter/linter-jshint/pull/51. I'll just remove the file.

nottrobin:
@bartaz removed.

bartaz:
@nottrobin would adding `.jshintignore` file (https://github.com/jshint/jshint/blob/master/examples/.jshintignore) help?

nottrobin:
@bartaz it's okay, it turns out I can just tell me editor to not lint if there's no JSHint config found, which works for me. It's worth bearing in mind the `.jshintignore` solution, but since I don't actually need it now, we probably shouldn't add it until it's actually needed.

nottrobin:
@katiefenn can this be merged then?

PR #156, opened 2017-02-17
Properly mock API request in createSnap test

bartaz:


test/unit/src/common/actions/t_create-snap.js
```diff
@@ -89,6 +89,9 @@ describe('repository input actions', () => {
     });
 
     it('stores a CREATE_SNAPS_START action', () => {
+      scope.get('/api/github/snapcraft-yaml/foo/bar')
```
bartaz: `createSnaps` will try to call GH API for snapcraft.yaml, so we need to mock it properly to avoid test timeout.

PR #157, opened 2017-02-17
Help install snap

earnubs:
Help 'install snap' component.


src/common/components/help/install-snap/index.js
```diff
@@ -0,0 +1,30 @@
+import React, { Component, PropTypes } from 'react';
+
+import { HeadingSix } from '../../vanilla/heading/';
```
earnubs: This could be a little less declarative?

src/common/containers/build-details.js
```diff
@@ -60,6 +61,7 @@ class BuildDetails extends Component {
             <BuildLog logUrl={build.buildLogUrl} />
           </div>
         }
+        <HelpInstallSnap name='foo' revision='1' />
```
earnubs: Stubbed out for now, blocked on need to get snap name and revision into redux.

src/common/style/base.css
```diff
@@ -49,8 +49,8 @@ q:before, q:after {
 }
 
 a {
-  color: #dd4814;
-  text-decoration: none;
+  color: $cool-grey;
```
earnubs: Global change, but seems to be what the layouts have.

src/common/style/base.css
```diff
@@ -64,47 +64,3 @@ a:focus {
 body {
   background-color: #FFF;
 }
-
-h1,
```
earnubs: These just get in the way of we're using react components for h1-6, also they're not in line with current Vanilla afaics.

src/common/components/help/install-snap/index.js
katiefenn: They could be. When they're as small and as simple as this, the difference seems marginal.

earnubs: Yeah, but there's a lot of room for error in all this boilerplate.

src/common/style/base.css
bartaz: @earnubs Layout are grayscale wireframes and Vanilla has coloured links, so I'm not so sure about that...

PR #158, opened 2017-02-17
Latest build status component

bartaz:
screen shot 2017-02-17 at 14 46 10

Showing styles latest build with build date in repository listing.

### DIFF
--- a/src/common/components/build-status/index.js
+++ b/src/common/components/build-status/index.js
@@ -1,5 +1,6 @@
 import React, { PropTypes } from 'react';
 import { Link } from 'react-router';
+import moment from 'moment';
 
 import { BuildStatusConstants } from '../../helpers/snap-builds';
 
@@ -10,7 +11,8 @@ const BuildStatus = (props) => {
   const {
     link,
     status,
-    statusMessage
+    statusMessage,
+    dateStarted
   } = props;
 
   const statusStyle = {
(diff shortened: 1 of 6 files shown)


src/common/components/build-status/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+  const statusStyle = {
+    [BuildStatusConstants.SUCCESS]: styles.success,
+    [BuildStatusConstants.ERROR]: styles.error,
+    [BuildStatusConstants.PENDING]: styles.pending
+  };
+
+  let humanDateStarted;
+
+  if (dateStarted) {
```
bartaz: Build status and date handling mostly copied from current `BuildRow` component.
`BuildRow` need to be redesigned as MAAS table anyway, so it can now reuse `BuildStatus` as part of it.

katiefenn: This could be extracted into a function.

src/common/components/repository-row/index.js
```diff
@@ -19,9 +21,12 @@ const RepositoryRow = (props) => {
             and also show 'Never built' when no builds available
         */}
         { latestBuild &&
-          <Link to={ `/${fullName}/builds/${latestBuild.buildId}`}>
-            { latestBuild.statusMessage }
-          </Link>
+          <BuildStatus
```
katiefenn: Suggestion: This could use some unit tests, time permitting.

bartaz: It would be great to have more react components tests - but as repos list doesn't have any yet, I don't want to spend time creating them from scratch now.

PR #162, opened 2017-02-17
154: Added snap name to enabled repos list

katiefenn:


src/common/components/repository-row/index.js
```diff
@@ -33,6 +39,14 @@ const RepositoryRow = (props) => {
   );
 };
 
+const renderSnapName = (snapName) => {
+  if (!snapName) {
+    return (<a>Not registered</a>);
```
katiefenn: Registration integration to be completed later.

src/server/handlers/github.js
```diff
@@ -100,9 +101,8 @@ export const getSnapName = (repositoryUrl, token) => {
         .then((snapcraftYaml) => {
           const snapName = snapcraftYaml.name;
           if (snapName) {
-            return getMemcached().set(cacheId, snapName, 3600, () => {
-              return resolve(snapName);
-            });
+            getMemcached().set(cacheId, snapName, 3600);
+            return resolve(snapName);
```
katiefenn: This was causing no snapname to be returned even if one existed when the application is not run with Memcached. There is no reason that the absence of Memcached should cause this feature to fail.

bartaz: Memcached is stubbed when not configured (https://github.com/canonical-ols/build.snapcraft.io/blob/master/src/server/helpers/memcached.js#L15) so it should always call the callback.

Maybe something else was causing an issue?

src/common/components/repository-row/index.js
bartaz: Why `<a>` if no href?

Also - if `snapName` is null - does it mean name is not registered or only that it's not defined in snapcraft.yaml?

src/server/handlers/launchpad.js
```diff
@@ -381,13 +382,24 @@ export const findSnaps = (req, res) => {
...
-          code: 'snaps-found',
-          snaps: snaps
-        }
-      });
+      return Promise.all(snaps.map((snap) => {
+        return getSnapName(snap.git_repository_url, req.session.token)
+          .then((snapName) => {
+            return {
+              ...snap,
+              snap_info: { name: snapName }
```
bartaz: Whole `snap` object is "snap info", so while it doesn't matter much how we name it maybe some other name would be more appropriate? `yaml` or `snapcraft_yaml`? - we also should make sure we don't collide with snap API props names (not to override something by mistake).

src/common/components/repository-row/index.js
```diff
@@ -14,7 +14,13 @@ const RepositoryRow = (props) => {
     <Row>
       <Data col="30"><Link to={ `/${fullName}/builds` }>{ fullName }</Link></Data>
       <Data col="20"> </Data>
```
bartaz: This is the column that checks snapcraft.yaml status.
If `snapName` is defined, we know that snapcraft.yaml is valid, so we can show it here.

katiefenn: @bartaz @cjwatson: I think this is risky. `snapName` being defined infers snapcraft.yaml is valid, but how do we tell snapcraft.yaml is invalid or missing? I have an [open question about this in Trello](https://trello.com/c/tCfftHzV/79-snapcraft-yaml-integration-1d). I'm reluctant to start this feature until this has been clarified.

mocks/service/github/responses.js
```diff
@@ -79,3 +79,14 @@ export function okayBadSharedSecret(req, res) {
...
+export function okaySnapcraftYamlFound(req, res) {
+  res.status(200).send(`name: ${req.params.name}\n`);
+}
+
+export function errorSnapcraftYamlNotFound(req, res) {
+  res.status(404).send({
+    "message": "Not Found",
+    "documentation_url": "https://developer.github.com/v3"
+  });
+}
```
katiefenn: Added mock for missing snapcraft.yml file

src/server/handlers/github.js
```diff
@@ -99,13 +100,18 @@ export const getSnapName = (repositoryUrl, token) => {
...
-          } else {
-            return resolve(null);
+          const snapcraftData = {};
+          for (let index in snapcraftYaml) {
+            if (snapcraftYaml.hasOwnProperty(index)) {
+              if (SNAPCRAFT_INFO_WHITELIST.indexOf(index) >= 0) {
+                snapcraftData[index] = snapcraftYaml[index];
+              }
+            }
           }
```
katiefenn: I've replaced storage of the snap name in Memcached with an object of whitelisted items from snapcraft.yml. This felt like the best way to cache both the snap name and the existence of the snapcraft.yml file.

src/server/handlers/github.js
```diff
@@ -138,31 +144,19 @@ export const listRepositories = (req, res) => {
         });
       }
 
-      return Promise.all(response.body.map((repo) => {
```
katiefenn: Removed the `snap_info` feature from `listRepositories`. The snap name is not needed in the "select repositories" list, and stripping this out should improve performance.

src/server/handlers/github.js
```diff
@@ -99,13 +100,18 @@ export const getSnapName = (repositoryUrl, token) => {
...
         .then((snapcraftYaml) => {
-          const snapName = snapcraftYaml.name;
-          if (snapName) {
-            getMemcached().set(cacheId, snapName, 3600);
-            return resolve(snapName);
-          } else {
-            return resolve(null);
+          const snapcraftData = {};
+          for (let index in snapcraftYaml) {
+            if (snapcraftYaml.hasOwnProperty(index)) {
```
cjwatson: `for (const index of Object.keys(snapcraftYaml))` might be a bit more compact.

test/routes/src/server/routes/t_github.js
```diff
@@ -296,9 +273,9 @@ describe('The GitHub API endpoint', () => {
             .get('/github/repos')
             .end((err, res) => {
               expect(res.body.payload.repos).toEqual([
-                { ...repos[0], snap_info: { name: 'snap1' } },
-                { ...repos[1], snap_info: { name: null } },
-                { ...repos[2], snap_info: { name: null } }
+                { ...repos[0] },
+                { ...repos[1] },
+                { ...repos[2] }
```
cjwatson: Can possibly just be `expect(res.body.payload.repos).toEqual(repos)` now?

PR #163, opened 2017-02-17
proptype should be number or string

earnubs:


earnubs:
@cjwatson I was going with "liberal in what you accept ..." given me own error and what this actually does, but I'm happy to make the change.

PR #164, opened 2017-02-20
match specs from invision doc

earnubs:
invision doc does in one case mention specific vanilla patterns to use for these elements, adding them in this PR


earnubs:
[[dashboard: repo detail] add install snapd + install snap copy (0.5d)](https://trello.com/c/Csljntb8/120-dashboard-repo-detail-add-install-snapd-install-snap-copy-0-5d)

PR #165, opened 2017-02-20
Login redirect to dashboard

bartaz:
Login with GH should properly redirect to /dashboard and only if user doesn't have enabled repos show add repositories view (go to/dashboard/select-repositories).

### DIFF
--- a/src/common/components/repositories-home/index.js
+++ b/src/common/components/repositories-home/index.js
@@ -66,8 +66,9 @@ class RepositoriesHome extends Component {
   }
 
   render() {
+    const { snaps } = this.props;
     // show spinner until we know if user has any enabled repos
-    return this.props.snaps.success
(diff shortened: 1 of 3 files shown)


src/common/components/repositories-home/index.js
```diff
@@ -46,12 +56,29 @@ class RepositoriesHome extends Component {
...
+    return (
+      <div className={ spinnerStyles }>
+        <Spinner />
+      </div>
+    );
+  }
+
+  render() {
+    // show spinner until we know if user has any enabled repos
+    return this.props.snaps.success
```
bartaz: I decided to show spinner until we know if user has any enabled repos, so the empty UI of repositories list doesn't pop up for a second.

cjwatson: I think we need to handle the error case here rather than spinning forever.

PR #166, opened 2017-02-20
Text fixes

matthewpaulthomas:
Text changes in the Dashboard:
- Removes spurious sign-in prompt from Add dialog
- Changes 'Enable repos' to 'Add'
- Removes incorrect statement about all your repos being added (will be replaced by the first-run flow later)
- Shortens button label 'Add new repository' to 'Add repos...'.

### DIFF
--- a/src/common/components/repositories-home/index.js
+++ b/src/common/components/repositories-home/index.js
@@ -49,7 +49,7 @@ class RepositoriesHome extends Component {
         <h2>Repos to build and publish</h2>
         <div className={ styles['button-container'] }>
           <Anchor appearance="neutral" href="dashboard/select-repositories/">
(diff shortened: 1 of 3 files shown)


bartaz:
Thanks @matthewpaulthomas.

Your changes fail on CI because removing headings left some unused variables that fail in lint check.

To fix those you would need to:

in `src/common/components/select-repository-list/index.js` remove whole line 17:
`import { HeadingThree } from '../vanilla/heading';` (because HeadingThree is not used anymore)

in `src/common/components/select-repositories-page/index.js` line 6 remove `HeadingTwo` from import, because only HeadingTree is needed in this component.

This should fix lint errors and allow us to merge your branch, thanks!

PR #167, opened 2017-02-20
Added table dropdown component with dummy usage in repository row

bartaz:
screen shot 2017-02-20 at 16 10 17

screen shot 2017-02-20 at 16 10 29

Simple dropdown component for interactive table with basic usage in RepositoryRow.

### DIFF
--- a/src/common/components/repositories-list/repositoriesList.css
+++ b/src/common/components/repositories-list/repositoriesList.css
@@ -1,0 +1,7 @@
+.repositoriesList {
+  background-color: $light-grey;
+  padding: 20px;
+  width: 100%;
+  clear: both;
+  position: relative;
+}
--- a/src/common/components/repository-row/index.js
+++ b/src/common/components/repository-row/index.js
@@ -1,37 +1,75 @@
-import React, { PropTypes } from 'react';
+import React, { Component, PropTypes } from 'react';
 import { Link } from 'react-router';
 
-import { Row, Data } from '../vanilla/table-interactive';
+import { Row, Data, Dropdown } from '../vanilla/table-interactive';
 import BuildStatus from '../build-status';
 
 import { parseGitHubRepoUrl } from '../../helpers/github-url';
 
-const RepositoryRow = (props) => {
-  const { snap, latestBuild } = props;
-  const { fullName } = parseGitHubRepoUrl(snap.git_repository_url);
+class RepositoryRow extends Component {
 
-  return (
-    <Row>
-      <Data col="30"><Link to={ `/${fullName}/builds` }>{ fullName }</Link></Data>
-      <Data col="20"> </Data>
-      <Data col="20"> </Data>
-      <Data col="30">
-        {/*
-          TODO: show 'Loading' when waiting for status?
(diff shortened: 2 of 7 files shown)


src/common/components/repositories-list/repositoriesList.css
```diff
@@ -0,0 +1,7 @@
+.repositoriesList {
```
bartaz: Grey background under repos table.

It's probably a third place we use 'light strip' pattern without actually creating a component for it... so maybe we will at some point find it useful when refactoring.

src/common/components/repository-row/index.js
```diff
@@ -1,37 +1,75 @@
...
+    super(props);
+
+    this.state = {
+      notConfiguredDropdownExpanded: false
+    };
+  }
+
+  onConfiguredClick() {
+    this.setState({
+      notConfiguredDropdownExpanded: !this.state.notConfiguredDropdownExpanded
```
bartaz: With interactive table we need some place to store the state of the component (if dropdowns were expanded or not). Some rows may even have 2 dropdowns (in Configured and Registered columns), so there will be 2 dropdowns where one will close the other.

Ideally it would be great to keep all this state in redux, but it creates a lot of boilerplate code to create actions and reducers for it now, so I suppose it will be easier to keep this state just in component instance for now and see if later we want to extract it to redux.

src/common/components/repository-row/index.js
```diff
@@ -1,37 +1,75 @@
...
+        </Row>
+      </Dropdown>
+    );
+  }
+
+  render() {
+    const { snap, latestBuild } = this.props;
+    const { fullName } = parseGitHubRepoUrl(snap.git_repository_url);
+
+    const notConfigured = true;
```
bartaz: Hardcoded just for the example.

The actual state of given repository should be deducted from the data passed to the component (snap, snapcraft.yaml, etc) - it is covered in separate tickets.

src/common/components/repository-row/index.js
```diff
@@ -1,37 +1,75 @@
...
+    const showNotConfiguredDropdown = notConfigured && this.state.notConfiguredDropdownExpanded;
+
+    const isActive = showNotConfiguredDropdown; // TODO (or any other dropdown)
+    return (
```
bartaz: This diff is much bigger than it should be because whole thing in indented more now.

src/common/components/repository-row/index.js
```diff
@@ -1,37 +1,75 @@
...
-    </Row>
-  );
-};
+class RepositoryRow extends Component {
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      notConfiguredDropdownExpanded: false
```
cjwatson: Pretty confusing variable name.  Maybe `unconfiguredDropdownExpanded` so that it doesn't sound like "not" goes with "expanded"?

bartaz: True. It came from 'Not configured' snap state, but I agree that in the longer variable name it's confusing what 'not' is negating actually ;)

PR #168, opened 2017-02-20
Add /api/store/register-name endpoint

cjwatson:
Simple endpoint to proxy register-name requests from the client in
advance of cross-origin support in the SCA register-name endpoint.

### DIFF
--- a/src/server/handlers/store.js
+++ b/src/server/handlers/store.js
@@ -1,0 +1,22 @@
+import 'isomorphic-fetch';
+
+import { conf } from '../helpers/config';
+
+export const registerName = (req, res) => {
+  const snapName = req.body.snap_name;
+  const macaroon = req.body.macaroon;
+
+  return fetch(`${conf.get('STORE_API_URL')}/register-name/`, {
+    method: 'POST',
+    headers: {
+      'Authorization': `Macaroon root="${macaroon}"`,
+      'Content-Type': 'application/json',
+      'Accept': 'application/json'
+    },
+    body: JSON.stringify({ snap_name: snapName })
+  }).then((response) => {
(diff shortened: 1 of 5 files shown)


src/server/handlers/store.js
```diff
@@ -0,0 +1,22 @@
...
+    method: 'POST',
+    headers: {
+      'Authorization': `Macaroon root="${macaroon}"`,
+      'Content-Type': 'application/json',
+      'Accept': 'application/json'
+    },
+    body: JSON.stringify({ snap_name: snapName })
+  }).then((response) => {
+    return response.json().then((json) => {
+      return res.status(response.status).send(json);
```
bartaz: For sake of consistency it may be worth using same response pattern we use elsewhere: `{ status: 'success', payload: 'response data'}`.

I also assume we don't treat errors in any special way, we just pass them to client as they are?

test/routes/src/server/routes/t_store.js
```diff
@@ -0,0 +1,38 @@
...
+describe('The store API endpoint', () => {
+  const app = Express();
+  app.use(store);
+
+  describe('register-name route', () => {
+    afterEach(() => {
+      nock.cleanAll();
+    });
+
+    it('passes request through to store', (done) => {
```
bartaz: Could be worth having test for error response.

Especially if we end up wrapping it in `{ success, payload }` scheme. Right now I agree that it may be not necessary as it's just a pass-through.

src/server/handlers/store.js
cjwatson: I'm deliberately not touching the response so that it's easy to switch to making the request to SCA directly once it's ready.  Yes, errors are unmodified.

PR #170, opened 2017-02-21
Add actions for snap name registration

cjwatson:
These aren't hooked up to anything yet.

### DIFF
--- a/src/common/actions/register-name.js
+++ b/src/common/actions/register-name.js
@@ -15,7 +15,7 @@ export function registerName(fullName, snapName) {
   return (dispatch) => {
     dispatch({
       type: REGISTER_NAME,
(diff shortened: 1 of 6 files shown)


src/common/actions/register-name.js
```diff
@@ -0,0 +1,89 @@
...
+      })
+      .then(() => dispatch(registerNameSuccess(fullName)))
+      .catch((error) => dispatch(registerNameError(fullName, error)));
+  };
+}
+
+export function registerNameSuccess(id) {
+  return {
+    type: REGISTER_NAME_SUCCESS,
+    payload: { id }
```
bartaz: Would it be worth to storing the snapName that was registered?

PR #175, opened 2017-02-21
Removes GitHub profile link from the header

matthewpaulthomas:
Also changes 'Log in' to 'Sign in' and 'Log out' to 'Sign out' per spec.

### DIFF
--- a/src/common/components/header/index.js
+++ b/src/common/components/header/index.js
@@ -18,7 +18,7 @@ class Header extends Component {
           { authenticated
             ?
               <div className={ styles.sideNav }>
-                { user && Hi, {user.name} }
+                { user && `Hi, {user.name}` }
                 <Link to="/dashboard" className={ styles.link }>Dashboard</Link>
                 <a
                   className={ styles.link }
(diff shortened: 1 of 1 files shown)


matthewpaulthomas:
[[header] user name shouldn't link to github, and we should have "Hi" before user name](https://trello.com/c/l0UDc3SH/162-header-user-name-shouldn-t-link-to-github-and-we-should-have-hi-before-user-name)

src/common/components/header/index.js
```diff
@@ -18,18 +18,18 @@ class Header extends Component {
           { authenticated
             ?
               <div className={ styles.sideNav }>
-                { user && <a href={user.html_url} className={ styles.link }>{user.name}</a> }
+                { user && Hi, {user.name} }
```
bartaz: This wouldn't work this way - we still need some wrapping element to properly render and style it. 

The simplest way to fix it would be to make it:
`{ user && <span className={ styles.link } >Hi, {user.name}</span> }`

This will simply apply the same style to it as to links (in terms of positioning in header) - it will not make it clickable or change on hover.

PR #177, opened 2017-02-21
add workflow text and images + snap image

pmahnke:
## Done

On the landing page: 

* Added images and text for the workflow page based on the [copy doc](https://docs.google.com/document/d/1ijyI7EMPu0BoyVbB-mPB7gyKnPyloC0Rvx1Oq-yEqKk/edit#)
* Added snap block to the snap section
* Reduced the size of the top illustration per Jamie

## QA

1. go to the landing page
2. see that the top illustration is 920px
3. see that the workflow section is correct compared to the copy doc
4. see there is a snap image in the snap section


src/common/containers/landing.css
```diff
@@ -35,9 +35,20 @@
...
 .workflowImage {
-  display: block;
-  max-width: 100%;
-  margin-bottom: 60px;
+  display: inline;
+  justify-content: center;
+  margin-bottom: 20px;
+}
+
+.workflowItemOneThrid {
```
earnubs: typo on 'Thrid', should be Third?

pmahnke: good catch - fixed, but had to update the class as well.

PR #189, opened 2017-02-21
Added links to headings in build pages

bartaz:


src/common/components/vanilla/heading/heading.css
```diff
@@ -40,3 +40,7 @@
   font-size: 1.063rem;
   line-height: 1.412;
 }
+
+.heading a {
+  color: inherit;
```
bartaz: Making link in headers in text colour so we don't end up with huge orange titles.

PR #196, opened 2017-02-22
Fix dashboard redirect loop

cjwatson:
On the first time through for a new user, /dashboard will fetch snaps
and find an empty list, then send the user to
/dashboard/select-repositories.  If the user then creates a snap and
returns to /dashboard, the store may already have state that looks like
`{ snaps: { success: true, snaps: [] } }`, which will cause it to
immediately redirect back to /dashboard/select-repositories, entering a
redirect loop.

Clearing previous success/error state in response to the `FETCH_SNAPS`
action avoids this problem.

### DIFF
--- a/src/common/reducers/snaps.js
+++ b/src/common/reducers/snaps.js
@@ -10,7 +10,9 @@ export function snaps(state = {
     case ActionTypes.FETCH_SNAPS:
       return {
         ...state,
-        isFetching: true
+        isFetching: true,
+        success: false,
+        error: null
       };
     case ActionTypes.FETCH_SNAPS_SUCCESS:
       return {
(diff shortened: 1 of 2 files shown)


src/common/reducers/snaps.js
```diff
@@ -10,7 +10,9 @@ export function snaps(state = {
     case ActionTypes.FETCH_SNAPS:
       return {
         ...state,
-        isFetching: true
+        isFetching: true,
+        success: false,
+        error: null
```
bartaz: I wonder if we don't want to clear `snaps` too (if we clear the success flag).

But on the other hand, we probably don't want to clear data that we already have, so maybe it's not an issue.

cjwatson: Yeah, that was exactly my train of thought. :-)  I think it's best to keep the existing data for now, but we may decide differently in the future.

PR #197, opened 2017-02-22
Tweaks snap install text

matthewpaulthomas:
Reflecting that (a) we’re building on only two architectures and (b) the build page doesn't yet know whether this build succeeded.


matthewpaulthomas:
["Install the latest successful build" in HelpInstallSnap ... match spec, remove --revision](https://trello.com/c/ySkO0QKO/196-install-the-latest-successful-build-in-helpinstallsnap-match-spec-remove-revision)

PR #198, opened 2017-02-22
Move snap authorization after name registration

cjwatson:
Enabling a repository now creates a bare snap object in Launchpad with
no store configuration, and all the store configuration is filled in
later once the name is registered.  This lets us remove all dependencies
on being signed into the store from `/dashboard/select-repositories`.

### DIFF
--- a/src/common/components/repository-row/index.js
+++ b/src/common/components/repository-row/index.js
@@ -8,6 +8,7 @@ import BuildStatus from '../build-status';
 
 import { signIntoStore } from '../../actions/auth-store';
 import { registerName } from '../../actions/register-name';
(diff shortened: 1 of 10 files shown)


src/common/components/repository-row/index.js
```diff
@@ -82,7 +84,7 @@ class RepositoryRow extends Component {
 
     let caption;
     if (registerNameStatus.error) {
-      caption = <div>{ registerNameStatus.error.json.detail }</div>;
+      caption = <div>{ registerNameStatus.error.message }</div>;
```
bartaz: It would be nice to give it some error look, but it's not something we need to do in this PR.

PR #199, opened 2017-02-22
Show feedback on successful snap name registration

cjwatson:
As per spec: we show a checkmark and the name in the table column if the
name is registered, and after successful registration we show a
checkmark and "Registered successfully" for two seconds before closing
the dropdown panel.

### DIFF
--- a/src/common/components/repository-row/index.js
+++ b/src/common/components/repository-row/index.js
@@ -13,6 +13,13 @@ import { parseGitHubRepoUrl } from '../../helpers/github-url';
 
 import styles from './repositoryRow.css';
 
+const tickIcon = (
+  <img
+    src='http://assets.ubuntu.com/v1/6c395e6d-green-tick.svg'
+    width={ 22 } height={ 16 }
+  />
+);
+
 class RepositoryRow extends Component {
 
   constructor(props) {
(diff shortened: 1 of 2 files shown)


src/common/components/repository-row/index.js
```diff
@@ -13,6 +13,13 @@ import { parseGitHubRepoUrl } from '../../helpers/github-url';
 
 import styles from './repositoryRow.css';
 
+const tickIcon = (
+  <img
+    src='http://assets.ubuntu.com/v1/6c395e6d-green-tick.svg'
+    width={ 22 } height={ 16 }
```
bartaz: It would be better to move these to CSS (repositoryRow.css).

```css
.icon {
  width: 22px;
  height: 16px;
}
```

And use `className={styles.icon}` on the img.

src/common/components/repository-row/index.js
```diff
@@ -149,15 +179,24 @@ class RepositoryRow extends Component {
...
+    const showUnregisteredDropdown = this.state.unregisteredDropdownExpanded;
     const showRegisterNameInput = (
       showUnregisteredDropdown && authStore.authenticated
     );
+    const registeredName = (
+      registerNameStatus.success ?
+      registerNameStatus.snapName : snap.store_name
+    );
 
     const isActive = showUnconfiguredDropdown; // TODO (or any other dropdown)
```
bartaz: `isActive` should be set true also when `showUnregisteredDropdown` is true, this will make the row white.

PR #202, opened 2017-02-22
Trigger builds if possible once name is registered

cjwatson:
Once snapcraft.yaml exists and the name is registered, we can trigger
builds.  One path for this (already implemented) is that if the name is
registered first then we may be told of a new snapcraft.yaml by way of a
webhook push event.  The other possibility is implemented here: if
snapcraft.yaml already exists when the name is registered, we should
trigger builds at that point.

### DIFF
--- a/src/common/actions/register-name.js
+++ b/src/common/actions/register-name.js
@@ -5,6 +5,7 @@ import { MacaroonsBuilder } from 'macaroons.js';
 import { APICompatibleError, checkStatus, getError } from '../helpers/api';
 import { conf } from '../helpers/config';
 import { checkPackageUploadRequest } from './auth-store';
+import { requestBuilds } from './snap-builds';
 
 const BASE_URL = conf.get('BASE_URL');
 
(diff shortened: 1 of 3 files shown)


src/common/actions/register-name.js
```diff
@@ -139,6 +140,10 @@ export function registerName(repository, snapName) {
       .then(checkStatus)
       .then(() => dispatch(registerNameSuccess(fullName)))
       .catch((error) => dispatch(registerNameError(fullName, error)));
+    if (triggerBuilds) {
+      promise = promise.then(() => requestBuilds(repository.url)(dispatch));
```
bartaz: This should probably be `dispatch(requestBuilds(repository.url))`.

test/unit/src/common/actions/t_register-name.js
```diff
@@ -112,7 +116,7 @@ describe('register name actions', () => {
         type: ActionTypes.REGISTER_NAME,
         payload: { id: 'foo/bar', snapName: 'test-snap' }
       };
-      return store.dispatch(registerName(repository, 'test-snap'))
+      return store.dispatch(registerName(repository, 'test-snap', false))
```
bartaz: Can't it be just `false` by default?

It seems to actually be already as we only check `if (triggerBuilds)` so it would not be called for `undefined`.

PR #203, opened 2017-02-22
redirect unauthed users to /

earnubs:
Redirect /dashboard to / for any user who isn't authenticated.


earnubs:
[[dashboard] handle unauthenticated user](https://trello.com/c/zyKA2tAP/159-dashboard-handle-unauthenticated-user)

PR #205, opened 2017-02-22
203: UI fixes

katiefenn:
- Changed default traffic light colour
- Demoted a heading to make it smaller
- Repositioned repository list heading next to button
- Changed the avatar from a square to a circle
- Changed colour of "login" heading to dark grey
- Re-aligned "Add" button to the right on the "select repos" list
- Added cancel button to select repos list

### DIFF
--- a/src/common/components/first-time-heading/index.js
+++ b/src/common/components/first-time-heading/index.js
@@ -1,7 +1,7 @@
 import React, { Component, PropTypes } from 'react';
 
 import { parseGitHubRepoUrl } from '../../helpers/github-url';
-import { HeadingOne } from '../vanilla/heading';
+import { HeadingThree } from '../vanilla/heading';
 import TrafficLights, { SIGNALS } from '../traffic-lights';
 
 import styles from './firstTimeHeading.css';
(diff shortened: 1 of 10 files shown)


src/common/components/first-time-heading/index.js
```diff
@@ -59,11 +59,12 @@ class FirstTimeHeading extends Component {
 
   // TODO: bartaz display state properly (not based on message)
   renderProgress(progress) {
+    progress = [SIGNALS.DONE, SIGNALS.ACTIVE, SIGNALS.DEFAULT];
```
earnubs: is this a test line? it means not being able to set `progress` with `renderProgress`

src/common/components/select-repository-list/index.js
```diff
@@ -99,10 +101,20 @@ class SelectRepositoryList extends Component {
...
-              Add
-            </Button>
+          <HeadingThree>
+            { selectedRepos.length } selected
+          </HeadingThree>
+          <div className={ styles['footer-right'] }>
+            <div className={ styles['button-wrapper'] }>
+              <LinkButton appearance="neutral" to="/dashboard">
+                Cancel
+              </LinkButton>
```
cjwatson: We should probably hide this button in the case where you have no snaps, since otherwise `/dashboard` is going to redirect right back to `/dashboard/select-repositories`.

PR #206, opened 2017-02-22
Create webhook before creating snap

cjwatson:
We used to create a webhook during the setup workflow, but that fell out
during the major reorganisations of the last couple of weeks.  This
restores it.  We can safely create the webhook as the very first thing
we do, since the webhook handler will fail cleanly if the snap doesn't
exist yet, and error handling is easier this way round.

This duplicates part of `src/common/actions/webhook.js`, but that can
just be removed after this is in place; we don't need to have separate
actions and state for this.

### DIFF
--- a/src/common/actions/create-snap.js
+++ b/src/common/actions/create-snap.js
@@ -31,7 +31,7 @@ export function createWebhook(repository) {
         return response;
       } else {
         return response.json().then((json) => {
-          if (json.payload && json.payload.code == 'github-already-created') {
+          if (json.payload && json.payload.code === 'github-already-created') {
             return response;
           }
           throw getError(response, json);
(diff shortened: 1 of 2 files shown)


src/common/actions/create-snap.js
```diff
@@ -18,6 +18,28 @@ export function setGitHubRepository(value) {
...
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify({ owner, name }),
+    credentials: 'same-origin'
+  })
+    .then((response) => {
+      if (response.status >= 200 && response.status < 300) {
+        return response;
+      } else {
+        return response.json().then((json) => {
+          if (json.payload && json.payload.code == 'github-already-created') {
```
earnubs: style nit, `===`

PR #209, opened 2017-02-22
Landing page tweaks

pmahnke:
## DONE

* Added testimonials to the landing page
* Added left, center, right align property on Heading module - sorry

## QA

1. go to the landing page
2. look at the testimonials... see that the look like the [design spec](https://www.dropbox.com/s/6je201c3jp2as2z/1%20-%20testimonials%20-%20no%20avatar.jpg?dl=0) and [trello card instructions](https://trello.com/c/bPbPoJ1u)
3. read them and see that they conform to the [copy doc](https://docs.google.com/document/d/1ijyI7EMPu0BoyVbB-mPB7gyKnPyloC0Rvx1Oq-yEqKk/edit#)


src/common/components/vanilla/heading/index.js
```diff
@@ -4,15 +4,17 @@ import styles from './heading.css';
 
 const Heading = (props) => {
   const H = props.heading || 'h1';
+  const A = props.align || 'left';
```
bartaz: I'd rather call it `align`.

`H` is used above only because it's used later as an `<H>` tag (and I didn't want invent another name as `Heading` couldn't be used because it's name of this component already), but for other variables it would use more readable names.

src/common/components/vanilla/heading/index.js
```diff
@@ -4,15 +4,17 @@ import styles from './heading.css';
...
-    <H className={ styles[H] }>
+    <H className={ `${styles[H]} ${styles[A]} `}>
       { props.children }
     </H>
   );
 };
 
 Heading.propTypes = {
   children: PropTypes.node,
+  align: React.PropTypes.oneOf(['left', 'right', 'center']),
```
bartaz: Maybe it doesn't matter so much ATM, but there are no such variants of headings in vanilla.

In theory, if we want this heading component to be equivalent of vanilla it should only have styles of vanilla headings.

src/common/containers/landing.js
```diff
@@ -65,67 +67,33 @@ class Landing extends Component {
         </section>
 
         <section className={ styles.section }>
-          <HeadingSix>
+          <HeadingSix align='centered'>
```
bartaz: This should be `center`.

This heading is centered even with this typo because HeadingSix is always centered according to its styles (`.h6` class).

`.h6` seems to have very specific (non vanilla) styles ATM, so maybe it all should be some variant of Heading component.

src/common/containers/landing.js
```diff
@@ -134,11 +102,42 @@ class Landing extends Component {
             </div>
 
             <div className={styles.rowItemOneThird}>
-              <img src='https://assets.ubuntu.com/v1/cbd43363-snap-icon-default.svg' width='150' height='150'/></div>
+              <img src='https://assets.ubuntu.com/v1/7af63a6d-workflow-icon04.svg' width='150' height='150'/></div>
           </div>
         </section>
 
+
         {/* TODO testimonials */}
```
bartaz: TODO can be removed if it's done

src/common/components/vanilla/heading/index.js
pmahnke: I agree... we don't really ever do anything other than left align, but jujucharms added center... we should look at this again.

pmahnke:
@bartaz ready for re-review

src/common/components/vanilla/testimonial/index.js
```diff
@@ -0,0 +1,22 @@
+import React, { PropTypes } from 'react';
```
katiefenn: Is Testamonial a Vanilla pattern? I couldn't find anything matching this description in the Vanilla docs or in the MAAS docs.

If this isn't Vanilla, it should live in `src/common/components`.

pmahnke: good point.

the intention is that it will become a vanilla component... but it isn't yet.

I can move it.

pmahnke:
FYI, I can't merge it.

PR #210, opened 2017-02-22
Fetching snapcraft.yaml from all 3 possible locations

bartaz:
Fetching snapcraft.yaml from all 3 possible locations (`/snap/snapcraft.yaml`, `snapcraft.yaml`, `.snapcraft.yaml`) one by one.

If one succeeds we return it and don't fetch the others.

### DIFF
--- a/test/routes/src/server/routes/t_github.js
+++ b/test/routes/src/server/routes/t_github.js
@@ -19,6 +19,157 @@ describe('The GitHub API endpoint', () => {
   });
   app.use(github);
 
+  describe('get snapcraft.yaml route', () => {
+    const fullName = 'anowner/aname';
+
+    context('when snap/snapcraft.yaml is valid', () => {
+
+      beforeEach(() => {
+        scope = nock(conf.get('GITHUB_API_ENDPOINT'))
+        .get(`/repos/${fullName}/contents/snap/snapcraft.yaml`)
+        .reply(200, 'name: snap-name');
+      });
+
+      afterEach(() => {
+        nock.cleanAll();
+      });
+
+      it('should successfully return', (done) => {
+        supertest(app)
+        .get('/github/snapcraft-yaml/anowner/aname')
+        .expect(200, done);
+      });
+
+      it('should return a "success" status', (done) => {
+        supertest(app)
+        .get('/github/snapcraft-yaml/anowner/aname')
+        .expect(hasStatus('success'))
+        .end(done);
+      });
+
+      it('should return a body with a "snapcraft-yaml-found" message', (done) => {
+        supertest(app)
+        .get('/github/snapcraft-yaml/anowner/aname')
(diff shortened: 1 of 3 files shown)


test/routes/src/server/routes/t_github.js
```diff
@@ -19,6 +19,157 @@ describe('The GitHub API endpoint', () => {
   });
   app.use(github);
 
+  describe('get snapcraft.yaml route', () => {
```
bartaz: I guess we don't use out `/github/snapcraft-yaml/:owner/:name` route on client side anymore, but it was quite a nice way to test `internalGetSnapcraftYaml` in separation from more complicated functionality that uses it.

test/routes/src/server/routes/t_github.js
```diff
@@ -503,7 +654,7 @@ const hasMessage = (expected) => {
     if (typeof actual.body.payload === 'undefined'
         || typeof actual.body.payload.code === 'undefined'
         || actual.body.payload.code !== expected) {
-      throw new Error('Response does not have payload with code ' + expected);
+      throw new Error(`Response does not have payload with code ${expected} instead of ${actual.body.payload.code}`);
```
bartaz: It was annoying and very not helpful that it only showed expected value and not actual.

Possibly the same code `hasStatus` helper is in other test files too, so would be worth to update them all (or move to helper or extend `expect`).

test/routes/src/server/routes/t_webhook.js
```diff
@@ -198,13 +198,23 @@ describe('The WebHook API endpoint', () => {
...
+            const error = { message: 'Not Found' };
+
             getSnapcraftYaml = nock(conf.get('GITHUB_API_ENDPOINT'))
-              .get('/repos/anowner/aname/contents/snapcraft.yaml')
-              .query({
-                client_id: conf.get('GITHUB_AUTH_CLIENT_ID'),
-                client_secret: conf.get('GITHUB_AUTH_CLIENT_SECRET')
-              })
-              .reply(404, { message: 'Not Found' });
+              .get(`${contentsPath}/snap/snapcraft.yaml`)
```
bartaz: If there is no snapcraft.yaml in repo, we need to mock all 3 possible locations...

bartaz:
@cjwatson I tried your approach (http://paste.ubuntu.com/24047252/) with paths array and loop building up promise chain, but even though it seemed to look ok it didn't seem to work correctly and was throwing unhandled promise rejections for no particular reason.

So I ended simply 'manually' nesting promises (if one snapcraft.yaml fails, I try to fetch another).
With some helper functions it ended up without too much boilerplate code and not terribly nested.

bartaz:
As I'll be likely to be off most (or whole) Thursday due to travel feel free to fix if needed and merge this branch once approved.

PR #219, opened 2017-02-23
134: Build page styles

katiefenn:
screen shot 2017-02-23 at 16 53 13

screen shot 2017-02-23 at 16 58 37


bartaz:
It would be good to confirm with Jamiey, but I suppose he would like to make table on grey background for consistency with enabled repos.

bartaz:
I remember Jamiey also wanted to have more spacing on these views (margin between the table and components below it).

Right now I see that build log is quite far away from 'Build log' heading - which doesn't look good.

src/common/components/build-row/index.js
```diff
@@ -44,18 +37,20 @@ const BuildRow = (props) => {
...
+      <Data col="20">
         {humanDuration}
-      </div>
-      <div className={ styles.item }>
-        {humanDateStarted}
-      </div>
-    </div>
+      </Data>
+      <Data col="40">
+        <BuildStatus link="" status={status} statusMessage={statusMessage} dateStarted={humanDateStarted} />
```
bartaz: `link` is optional, so if it's empty could be omitted.

Also BuildStatus formats `dateStarted` on its own, so you can just pass `dateStarted` to it and remove the part that formats it (it's already done in BuildStatus).

src/common/containers/build-details.js
```diff
@@ -60,9 +61,23 @@ class BuildDetails extends Component {
...
+                  <Header col="20">Architecture</Header>
+                  <Header col="20">Duration</Header>
+                  <Header col="40">Result</Header>
+                </Row>
+              </Head>
+              <Body>
+                <BuildRow repository={repository} {...build} />
+              </Body>
+            </Table>
+            <h3>Build log</h3>
```
bartaz: Should be a `HeadingThree` to keep consistent with vanilla headings.

Also right now it seems that `styles.strip` give build log quite a large top/bottom margin, so maybe it would be good to move heading inside of the strip, so it's closer to build log.

matthewpaulthomas:
From the screenshots this looks a big improvement overall so I’d be happy to see it land.

Particularly better:
- Table rows are now consistent with the Dashboard
- Clearer gap between the build list/log and the installation instructions.

Worse:
- The “Build log” heading is now closer to the table above than to the log itself.

katiefenn:
Updated build page styles:

[x] Moved "Build log" heading to the same container as the build log
[x] Removed redundant dateStarted processing and link stuff
[x] Replaced h3 with HeadingThree

screen shot 2017-02-24 at 12 18 48

PR #221, opened 2017-02-23
Snap name registration tidying

earnubs:
Asking for review early, even though this doesn't cover all error states and keeping the dropdown open through SSO dance.

## Missing, to include in follow up:

* We're not handling the case when the SSO session expires (or maybe the macaroon?) in the background, the user is left with no clear way of dealing with that (the button does not change to 'Sign in').

## To QA:

* Click "Not registered" in repos to build and publish list.
* Check register button is green and on RHS appropriately spaced.
* Check cancel button is plain text.
* Enter "-ubuntu" in text input, check error is "Sorry the name can't start or end with a hyphen."
* Enter "ubuntu-" ... check as above.
* Enter "ubuntu" in text input, check error is "Sorry, that name is already taken. Try a different name.
If you think you should have sole rights to the name, you can file a claim."

### DIFF
--- a/src/common/components/forms/message.js
+++ b/src/common/components/forms/message.js
@@ -6,7 +6,7 @@ export default function Message({ children, status, text }) {
 
   return <div>
     { (children || text) &&
-      <p className={ style[status] }>{ children || text }</p>
+      <div className={ style[status] }>{ children || text }</div>
     }
   </div>;
 }
--- a/src/common/components/repositories-list/index.js
+++ b/src/common/components/repositories-list/index.js
@@ -95,8 +95,8 @@ class RepositoriesList extends Component {
           <Head>
             <Row>
               <Header col="30">Name</Header>
-              <Header col="20">Configured</Header>
-              <Header col="20">Registered for publishing</Header>
+              <Header col="15">Configured</Header>
+              <Header col="25">Registered for publishing</Header>
               <Header col="30">Latest build</Header>
             </Row>
           </Head>
--- a/src/common/components/repository-row/index.js
+++ b/src/common/components/repository-row/index.js
@@ -155,6 +155,8 @@ class RepositoryRow extends Component {
     if (registerNameStatus.success) {
       caption = <div>{ tickIcon } Registered successfully</div>;
     } else if ( registerNameStatus.error
+      && registerNameStatus.error.json
+      && registerNameStatus.error.json.payload
       && registerNameStatus.error.json.payload.code === 'already_registered') {
       caption = (
         <Message status='error'>
--- a/src/common/components/repository-row/index.js
+++ b/src/common/components/repository-row/index.js
@@ -15,7 +15,6 @@ import { parseGitHubRepoUrl } from '../../helpers/github-url';
 
 import styles from './repositoryRow.css';
 
-const MINIMUM_SNAP_NAME_LENGTH = 1;
 const FILE_NAME_CLAIM_URL = 'https://myapps.developer.ubuntu.com/dev/click-apps/register-name/';
 
 const LEARN_THE_BASICS_LINK = 'https://snapcraft.io/docs/build-snaps/your-first-snap';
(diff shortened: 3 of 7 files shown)


src/common/components/repository-row/repositoryRow.css
```diff
@@ -1,5 +1,6 @@
 .snapName {
   width: 100%;
+  height: 24px;
```
earnubs: the input height is breaking the layout, a better fix might be to make the rows taller, but I think that's a design change

src/common/components/forms/message.js
```diff
@@ -6,7 +6,7 @@ export default function Message({ children, status, text }) {
 
   return <div>
     { (children || text) &&
-      <p className={ style[status] }>{ children || text }</p>
+      <div className={ style[status] }>{ children || text }</div>
```
earnubs: A <p> can't have <p> children.

src/common/components/repositories-list/index.js
```diff
@@ -95,8 +95,8 @@ class RepositoriesList extends Component {
           <Head>
             <Row>
               <Header col="30">Name</Header>
-              <Header col="20">Configured</Header>
-              <Header col="20">Registered for publishing</Header>
+              <Header col="15">Configured</Header>
+              <Header col="25">Registered for publishing</Header>
```
earnubs: Give name a little more room.

src/common/components/repository-row/index.js
```diff
@@ -73,11 +75,28 @@ class RepositoryRow extends Component {
...
     const triggerBuilds = (!!snap.snapcraft_data);
-    this.props.dispatch(registerName(repository, snapName, triggerBuilds));
+    let clientValidationError = false;
+
+    if (/^-|-$/.test(snapName)) {
+      clientValidationError = {
+        message: 'Sorry the name can\'t start or end with a hyphen.'
+      };
+    } else if (!/[a-z0-9-]/.test(snapName)) {
+      // XXX we can't get here on any browser that supports pattern validation
```
earnubs: Safari will allow you to submit even with a pattern attr ... but I can't test this just yet because of #222

src/common/components/repository-row/index.js
```diff
@@ -135,6 +154,17 @@ class RepositoryRow extends Component {
     let caption;
     if (registerNameStatus.success) {
       caption = <div>{ tickIcon } Registered successfully</div>;
+    } else if ( registerNameStatus.error
+      && registerNameStatus.error.json.payload.code === 'already_registered') {
```
earnubs: @bartaz I need to look for more error codes from SCA this morning, one I've seen that is unhandled is 'unauthorized' when we're signed out of SSO (expired during session)

src/common/components/repository-row/index.js
```diff
@@ -73,11 +75,28 @@ class RepositoryRow extends Component {
...
   onRegisterClick(repositoryUrl) {
-    const { snap } = this.props;
+    const { snap, dispatch, fullName } = this.props;
     const repository = parseGitHubRepoUrl(repositoryUrl);
     const { snapName } = this.state;
     const triggerBuilds = (!!snap.snapcraft_data);
-    this.props.dispatch(registerName(repository, snapName, triggerBuilds));
+    let clientValidationError = false;
+
+    if (/^-|-$/.test(snapName)) {
```
bartaz: Not sure if this check works properly, I was just able to enter name with hyphens in start and end and they passed validation (API call refused to register that name)

bartaz: And I haven't seen error message "Sorry name can't start or end with a hyphen"

src/common/components/repository-row/index.js
```diff
@@ -10,11 +10,13 @@ import { Message } from '../forms';
 import templateYaml from './template-yaml.js';
 
 import { signIntoStore } from '../../actions/auth-store';
-import { registerName } from '../../actions/register-name';
+import { registerName, registerNameError } from '../../actions/register-name';
 import { parseGitHubRepoUrl } from '../../helpers/github-url';
 
 import styles from './repositoryRow.css';
 
+const MINIMUM_SNAP_NAME_LENGTH = 1;
```
bartaz: It's just one?

earnubs: Yeah, I actually added this thinking it was more than 1, but I checked with @fgallina and it's just 1 ... I could remove this but it's maybe useful for the future.

src/common/components/repository-row/index.js
earnubs: Sorry, I asked for review without properly checking I had everything done :(

src/common/components/repository-row/index.js
earnubs: Removed.

earnubs:
[Snap name registration tidying](https://trello.com/c/gK9QEboE/201-snap-name-registration-tidying)

src/common/components/repository-row/index.js
```diff
@@ -135,6 +147,19 @@ class RepositoryRow extends Component {
     let caption;
     if (registerNameStatus.success) {
       caption = <div>{ tickIcon } Registered successfully</div>;
+    } else if ( registerNameStatus.error
+      && registerNameStatus.error.json
+      && registerNameStatus.error.json.payload
+      && registerNameStatus.error.json.payload.code === 'already_registered') {
```
katiefenn: Non-blocking comment: We really need to review all error objects in the codebase so that this isn't necessary. If `registerNameStatus.error` is defined, there really should be a standard property with a unique status identifier (such as 'already_registered') on the error object.

PR #227, opened 2017-02-24
Updating store_name in snap when name is registered

bartaz:
Snaps with registered name should have `store_name` property set to a registered name (which is done on LP side)

Right now after registering the name from BSI we are not updating snaps list, so on client side they don't have this property set yet, so for example first time heading does not properly move to next step (because it sees snaps still as unregistered).

We could possibly reload snaps list (fetch them again) when name is registered, but this will be an unnecessary request and UI update.

In this PR I update the `store_name` directly in snaps data in redux store in response REGISTER_NAME_SUCCESS action.

### DIFF
--- a/src/common/reducers/snaps.js
+++ b/src/common/reducers/snaps.js
@@ -1,4 +1,31 @@
 import * as ActionTypes from '../actions/snaps';
+import * as RegisterNameActionTypes from '../actions/register-name';
+import { getGitHubRepoUrl } from '../helpers/github-url';
+
+function findSnapByFullName(snaps, fullName) {
+  const snap = snaps.filter((snap) => {
+    return snap.git_repository_url === getGitHubRepoUrl(fullName);
+  })[0];
+
+  return snap;
+}
+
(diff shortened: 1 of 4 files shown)


src/common/reducers/snaps.js
```diff
@@ -31,6 +58,11 @@ export function snaps(state = {
         success: false,
         error: action.payload
       };
+    case RegisterNameActionTypes.REGISTER_NAME_SUCCESS:
+      return {
+        ...state,
+        snaps: updateRegisteredName(state.snaps, action.payload.id, action.payload.snapName)
```
katiefenn: Non-blocking comment: We should look at replacing stores with a proper data structure lib. It would prevent us having to write functions to traverse and update data structures everywhere it is needed.

bartaz: @katiefenn do you know any for redux based apps - because we are likely not getting rid of stores/reducers if we keep redux, but it would be useful to have some 'data structure' abstraction on top of it if anyone has done one already

PR #229, opened 2017-02-24
Saving repository row state in browser store

bartaz:
State of repository row component (state of open dropdowns) is now stored in browser storage, so when user navigates to SSO and returns back it's preserved.

Stored state is cleared when component is unmounted (for example when user navigates within the app itself).

### DIFF
--- a/src/common/components/repository-row/index.js
+++ b/src/common/components/repository-row/index.js
@@ -2,6 +2,7 @@ import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
 import { Link } from 'react-router';
 import url from 'url';
+import localforage from 'localforage';
 
 import Button from '../vanilla/button';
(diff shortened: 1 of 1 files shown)


src/common/components/repository-row/index.js
```diff
@@ -44,6 +45,32 @@ class RepositoryRow extends Component {
     };
   }
 
+  saveState() {
+    localforage.setItem(`repository_row_${this.props.snap.git_repository_url}`, this.state);
+  }
```
katiefenn: Unrelated question:

When do we start paying off technical debt? This task is not a "must have", but I hope we all agree we all want to start using the store as the state of the application where possible. I think it's worth the effort of adding a reducer and action creators for persisting row state if it's not due today.

(cc: @bartaz @earnubs @cjwatson)

evandandrea:
🙌

PR #235, opened 2017-02-27
Show snap name in install help box

bartaz:
screen shot 2017-02-27 at 13 47 47

### DIFF
--- a/docs/routes.md
+++ b/docs/routes.md
@@ -69,8 +69,7 @@ To search for an existing snap:
     Accept: application/json
 
 Successful responses have `status` set to `success` and `code` set to
-`snap-found`; the `message` will be the URL of the snap on the Launchpad
-API.
+`snap-found`; the `snap` will be the snap data from Launchpad API.
 
 To search for snap builds:
 
--- a/src/common/containers/build-details.js
+++ b/src/common/containers/build-details.js
@@ -9,6 +9,7 @@ import BuildLog from '../components/build-log';
 import { Message } from '../components/forms';
 import HelpInstallSnap from '../components/help/install-snap';
 import { HeadingOne, HeadingThree } from '../components/vanilla/heading';
+import Spinner from '../components/spinner';
 
 import withRepository from './with-repository';
 import { fetchBuilds, fetchSnap } from '../actions/snap-builds';
@@ -18,9 +19,9 @@ import styles from './container.css';
 
 class BuildDetails extends Component {
 
-  fetchData({ snapLink, repository }) {
-    if (snapLink) {
-      this.props.dispatch(fetchBuilds(repository.url, snapLink));
+  fetchData({ snap, repository }) {
+    if (snap) {
+      this.props.dispatch(fetchBuilds(repository.url, snap.self_link));
     } else if (repository) {
       this.props.dispatch(fetchSnap(repository.url));
     }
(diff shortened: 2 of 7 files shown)


docs/routes.md
```diff
@@ -69,8 +69,7 @@ To search for an existing snap:
     Accept: application/json
 
 Successful responses have `status` set to `success` and `code` set to
-`snap-found`; the `message` will be the URL of the snap on the Launchpad
-API.
+`snap-found`; the `snap` will be the snap data from Launchpad API.
```
bartaz: I forgot to update docs in previous PR where API was updated.

src/common/containers/builds.js
```diff
@@ -18,9 +18,9 @@ import styles from './container.css';
 class Builds extends Component {
   fetchInterval = null
 
-  fetchData({ snapLink, repository }) {
-    if (snapLink) {
-      this.props.dispatch(fetchBuilds(repository.url, snapLink));
+  fetchData({ snap, repository }) {
```
bartaz: Both `Builds` and `BuildDetails` share exactly the same logic for fetching snap and builds data.

Not sure if that's the right PR to do it, but it may be worth abstracting it.

src/common/reducers/snap-builds.js
```diff
@@ -30,7 +30,7 @@ export function snapBuilds(state = {}, action) {
         [payload.id]: {
           ...state[payload.id],
           isFetching: false,
-          snapLink: payload.snapLink
+          snap: payload.snap
```
bartaz: Thinking about store structure I'm not sure if `snapBuilds` is the right place to store snap info.

While we use this info on build related pages (build listing and build details) it feels like maybe we should have a separate store for snap data.

We already have snaps (enabled repositories) store, but currently it's an array (so doesn't provide direct access to particular snaps in it) and also snap info fetched for build page may not be from enabled snap of current user (build pages are public). So it feels like completely new store would be needed for that.

But refactoring and restructuring the store feels like a big task that we need to discuss and prioritise separately I guess.

src/common/containers/build-details.js
```diff
@@ -80,10 +78,15 @@ class BuildDetails extends Component {
...
-          headline='To test the latest successful build on your PC or cloud instance:'
-          name='foo'
-        />
+        { isFetching &&
+          <div className={styles.spinner}><Spinner /></div>
+        }
+        { snap &&
+          <HelpInstallSnap
+            headline='To test the latest successful build on your PC or cloud instance:'
+            name={ snap.store_name }
```
earnubs: bike shed: `store_name` requires you to know what the store is in this case? so might be better as `registered_name`

src/common/containers/builds.js
earnubs: If you do, can we talk about the shape of the actions and reducers for these and the other fetch actions we have?

src/common/reducers/snap-builds.js
earnubs: https://github.com/paularmstrong/normalizr might be useful here

src/common/containers/build-details.js
cjwatson: It's called `store_name` in LP, and makes sense in that context (also hard to change now); BSI is just consuming the existing API here.

bartaz: We may think about transforming data from API on reducer level - to store only data we need for the app under names we control in BSI.
That would help in such case, and also it will allow us to avoid mixing camelCase that we use as standard with snake_case that we get from LP APIs.

PR #236, opened 2017-02-27
Hide pagination on single page list

earnubs:
* hide pagination component if there is only 1 page
* render pagination once and reuse
* rationalise state a little

### DIFF
--- a/src/common/actions/repositories.js
+++ b/src/common/actions/repositories.js
@@ -21,6 +21,10 @@ export function fetchUserRepositories(pageNumber) {
   return (dispatch) => {
     let urlParts = [BASE_URL, 'api/github/repos'];
 
+    dispatch({
+      type: FETCH_REPOSITORIES
+    });
+
     if (pageNumber) {
       urlParts.push('page/' + pageNumber);
     }
--- a/src/common/actions/repositories.js
+++ b/src/common/actions/repositories.js
@@ -5,8 +5,7 @@ import { conf } from '../helpers/config';
 
 const BASE_URL = conf.get('BASE_URL');
 
-export const FETCH_REPOSITORIES = 'FETCH_REPOSITORIES';
-export const FETCH_REPOSITORIES_SUCCESS = 'FETCH_REPOSITORIES_SUCCESS';
+export const FETCH_REPOSITORIES = 'FETCH_ALL_REPOSITORIES';
 export const FETCH_REPOSITORIES_ERROR = 'FETCH_REPOSITORIES_ERROR';
 export const SET_REPOSITORIES = 'SET_REPOSITORIES';
 export const SET_REPOSITORY_PAGE_LINKS = 'SET_REPOSITORY_PAGE_LINKS';
@@ -26,12 +25,6 @@ export function fetchUserRepositories(pageNumber) {
       urlParts.push('page/' + pageNumber);
     }
 
-    dispatch({
-      type: FETCH_REPOSITORIES
-    });
-
-    dispatch(setRepositories([]));
-
     return fetch(urlParts.join('/'), {
       headers: { 'Content-Type': 'application/json' },
       credentials: 'same-origin'
@@ -42,12 +35,10 @@ export function fetchUserRepositories(pageNumber) {
           if (result.status !== 'success') {
             throw getError(response, result);
           }
-          dispatch(setRepositories(result.payload.repos));
-          dispatch(fetchRepositoriesSuccess());
-
-          if (result.pageLinks) {
-            dispatch(setPageLinks(result.pageLinks));
-          }
+          dispatch(setRepositories({
+            repos: result.payload.repos,
+            links: result.pageLinks
+          }));
         })
         .catch(error => {
           return Promise.reject(error);
@@ -55,17 +46,9 @@ export function fetchUserRepositories(pageNumber) {
       })
       .catch(error => {
         // TODO: Replace with logging helper
-        /* eslint-disable no-console */
-        console.warn(error);
-        /* eslint-enable no-console */
+        console.warn(error); // eslint-disable-line no-console
         dispatch(fetchRepositoriesError(error));
       });
-  };
-}
-
-export function fetchRepositoriesSuccess() {
-  return {
-    type: FETCH_REPOSITORIES_SUCCESS
   };
 }
 
@@ -74,12 +57,5 @@ export function fetchRepositoriesError(error) {
     type: FETCH_REPOSITORIES_ERROR,
     payload: error,
     error: true
-  };
-}
-
-export function setPageLinks(pageLinks) {
-  return {
-    type: SET_REPOSITORY_PAGE_LINKS,
-    payload: pageLinks
   };
 }
--- a/src/common/actions/snaps.js
+++ b/src/common/actions/snaps.js
@@ -5,7 +5,7 @@ import { conf } from '../helpers/config';
 
 const BASE_URL = conf.get('BASE_URL');
 
-export const FETCH_SNAPS = 'FETCH_REPOSITORIES';
+export const FETCH_SNAPS = 'FETCH_SNAP_REPOSITORIES';
 export const FETCH_SNAPS_SUCCESS = 'FETCH_SNAPS_SUCCESS';
 export const FETCH_SNAPS_ERROR = 'FETCH_SNAPS_ERROR';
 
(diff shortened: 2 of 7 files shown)


src/common/actions/repositories.js
```diff
@@ -5,8 +5,7 @@ import { conf } from '../helpers/config';
 
 const BASE_URL = conf.get('BASE_URL');
 
-export const FETCH_REPOSITORIES = 'FETCH_REPOSITORIES';
-export const FETCH_REPOSITORIES_SUCCESS = 'FETCH_REPOSITORIES_SUCCESS';
```
earnubs: Success is the same as set. I like the (fsa)[https://github.com/acdlite/flux-standard-action#example] pattern for actions like these:

```
type: FETCH_ALL_REPOSITORIES,
type: FETCH_ALL_REPOSITORIES, status: 'error', payload: error
type: FETCH_ALL_REPOSITORIES, status: 'success', payload: [ repos ] 
```

And I suggest use ALL here to distinguish actions which get only SNAP repos.

src/common/actions/repositories.js
```diff
@@ -22,15 +21,13 @@ export function fetchUserRepositories(pageNumber) {
...
 
-    if (pageNumber) {
-      urlParts.push('page/' + pageNumber);
-    }
-
     dispatch({
       type: FETCH_REPOSITORIES
     });
 
-    dispatch(setRepositories([]));
```
earnubs: set the default in the reducer instead

src/common/actions/repositories.js
```diff
@@ -42,44 +39,27 @@ export function fetchUserRepositories(pageNumber) {
           if (result.status !== 'success') {
             throw getError(response, result);
           }
-          dispatch(setRepositories(result.payload.repos));
-          dispatch(fetchRepositoriesSuccess());
```
earnubs: I don't think we need a separate reducer for success, it's implicit in getting back results in this case.

src/common/actions/repositories.js
```diff
@@ -42,44 +39,27 @@ export function fetchUserRepositories(pageNumber) {
...
-
-          if (result.pageLinks) {
-            dispatch(setPageLinks(result.pageLinks));
```
earnubs: I think this should happen in the reducer, I've done it very simply for now but a refactor would move all of this to the reducer.

src/common/actions/snaps.js
```diff
@@ -5,7 +5,7 @@ import { conf } from '../helpers/config';
 
 const BASE_URL = conf.get('BASE_URL');
 
-export const FETCH_SNAPS = 'FETCH_REPOSITORIES';
+export const FETCH_SNAPS = 'FETCH_SNAP_REPOSITORIES';
```
earnubs: Since I'm not changing the name of the const only it's value this should have no effect other that to help people who grep and potentially avoid bugs in reducers.

src/common/actions/repositories.js
```diff
@@ -5,8 +5,7 @@ import { conf } from '../helpers/config';
 
 const BASE_URL = conf.get('BASE_URL');
 
-export const FETCH_REPOSITORIES = 'FETCH_REPOSITORIES';
-export const FETCH_REPOSITORIES_SUCCESS = 'FETCH_REPOSITORIES_SUCCESS';
+export const FETCH_REPOSITORIES = 'FETCH_ALL_REPOSITORIES';
```
earnubs: corresponds to FETCH_SNAP_REPOSITORIES later

earnubs:
[[add new repo] don't show pagination when not needed](https://trello.com/c/xFuobidN/173-add-new-repo-don-t-show-pagination-when-not-needed)

src/common/actions/repositories.js
bartaz: Shouldn't `FETCH_REPOSITORIES_ERROR` be removed in such case too? As they all can be handled by one action with status and payload?

One thing that worries me in this approach is that by using one action type for all instead of having one `switch` in reducer we need to put much more logic into every case (to handle different statuses of actions).
But maybe it's not such an overhead in small cases like this.

src/common/components/select-repository-list/index.js
```diff
@@ -90,17 +90,18 @@ class SelectRepositoryList extends Component {
     const isLoading = this.props.repositories.isFetching;
     const { selectedRepos } = this.props.selectRepositoriesForm;
     const { repos } = this.props.repositories;
+    const pageLinks = this.renderPageLinks.call(this);
 
     return (
       <div>
-        { this.renderPageLinks.call(this) }
+        { pageLinks }
```
bartaz: I think wireframes only show pagination on the bottom of the list, so it's as good moment as any other to remove the top one.

src/common/reducers/repositories.js
```diff
@@ -2,10 +2,11 @@ import * as ActionTypes from '../actions/repositories';
 import { parseGitHubRepoUrl } from '../helpers/github-url';
 
 export function repositories(state = {
-  isFetching: false,
+  // has no server side content, so initial state is always 'fetching'
+  isFetching: true,
```
bartaz: Why is that?
It shouldn't be `isFetching` until action is called to start fetching the data.

src/common/actions/repositories.js
earnubs: Sorry, my message is really about the future not this PR, so error is still it's own action here.

I would argue that the logic is similar and not more, but I'm not speaking from experience. For example in place of a case of ERROR and SUCCESS you're testing for properties on the payload, which is the same logic but different.

src/common/reducers/repositories.js
earnubs: Ok, so I went round in my head about this a few times. Here's the argument for what I've done:

The component in it's initial state on mount has no data, so it's 'in need of data' and should show a spinner to let the user know that; calling `fetch()` is just a consequence of the state of being without data, so the property might be better as `needsData` perhaps to break the semantic from `fetch` API to component state.

Does that make sense? And if it makes sense is it valid?

src/common/components/select-repository-list/index.js
earnubs: Ah-ha, will do!

src/common/reducers/repositories.js
bartaz: It seems to make sense, but it also seems to be making things more complicated.

I agree that initial empty state needs we are in `needData: true` state. But it doesn't mean we are in `isFetching: true` state (which means we already triggered a request and are waiting for response).

While it may not be the case in this particular reducer we use the same `isFetching` state everywhere else and in some places there are explicit checks for `isFetching` being true to make sure we don't try to fetch data again while we are waiting for response.
By setting `isFetching` to true in initial state we would break it and data will never load (because components that request the data will assume we are already waiting for it).

I understand the reasoning behind it. Right now our server side rendering doesn't do much - it renders incomplete empty initial state, without spinner even (because actions have not triggered yet), but I don't think using `isFetching` is a good place to fix that.

On the other hand - what is really the benefit of server side rendering the spinner?

katiefenn: Agree with @bartaz.

These three-state situations come up in components quite a lot, and whenever I've tried to reduce it to one binary state it hasn't worked. I think an "initial state", "needs data", "uninitialised" prop would be good, supporting the existing "isFetching" prop.

earnubs: I'm changing this back because it's not relevant to the PR, but I'm not completely convinced by your arguments (neither convinced my own solution) :)

The reason I'm not convinced by the argument is we don't set fetching state when a component is fetching, we set it before a component is fetching, so the true meaning of `isFetching` isn't quite right in either case (some failure could prevent an actual fetch in both cases) ... but from a user's point of view, from the moment that HTML is rendered, that is *before* React has reconciled the tree, that component is in a state of "fetching" ... fetching js, fetching data, whatever, but because there's a fetch API it sounds more like we mean we're engaged in request fetch.

katiefenn: I don't think it matters when it is fetching, we could dispatch this action after the request is initiated if we wanted, and it would still fulfil its purpose: communicating to the user that the application is waiting for data it doesn't have. Other applications use props named "waiting" or "pending".

I think it's still worth having this discussion, because this is a good example of state that will be re-used wherever we have a component that waits for data. Sooner or later we'll need to have the conversation again. Here's my recommendation:

- **pending**: Component is waiting for new data when true
- **initialised**: Component needs new data, but cannot request it because it is rendered on server or waiting for user interaction when false.
- **last-requested**: A timestamp of when the last request started. Used for debouncing API requests.

PR #237, opened 2017-02-27
Extracted fetching snap builds to higher order component

bartaz:
Both builds listing for repo and build details did a bunch of same fetch logic (getting repo from URL, fetching snap and builds for it), so I extracted this functionality to higher order component.

### DIFF
--- a/src/common/containers/builds.js
+++ b/src/common/containers/builds.js
@@ -1,5 +1,4 @@
 import React, { Component, PropTypes } from 'react';
-import { connect } from 'react-redux';
 import Helmet from 'react-helmet';
 import { Link } from 'react-router';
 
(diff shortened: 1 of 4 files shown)


src/common/containers/builds.js
```diff
@@ -89,29 +55,15 @@ Builds.propTypes = {
...
+      self_link: PropTypes.string.isRequired,
+      store_name: PropTypes.string.isRequired
+    }),
+    success: PropTypes.bool,
+    error: PropTypes.object
+  })
 };
 
-export default connect(mapStateToProps)(withRepository(Builds));
+export default withRepository(withSnapBuilds(Builds));
```
bartaz: I moved `connect` out from here to higher order components.

So now this `Builds` component doesn't need to know how data is mapped (from state/params, etc).
Before that this component needed to define `mapStateToProps` to prepare data for both `withRepository` and `withSnapBuilds` wrapping components, but now they do it on their own.

Little drawback here is that now wrapping components have built-in data binding, so it may be harder to reuse them in different context. But I'm not sure if that will be the case - for now they do their job well - they hide repeated functionality and provide data to components that can be purely visual and not care about fetching data.

bartaz:
As this is a refactoring task feel free to discuss implementation details and if you find it readable and understandable.

From code point of view it's a nice abstraction and allows to remove some duplication. Also provides a nice 'data' layer (for logic of connecting to the store and fetching), so that visual components can focus on just rendering.

On the other hand they hide a lot of functionality and may cause issues when debugging (and not knowing where the data comes from, or what exactly is the data they provide).

earnubs:
@bartaz from a 'reading review' point of view this seems to me to be what the HoC pattern is meant for, you've got some cross cutting concerns and now they're tidied up into a single component

src/common/containers/builds.js
katiefenn: I think this is a good trade off. Wrapping / container components like this have a specific role: grouping presentational components together to make a user interface. I do not think we will want to re-use these much.

PR #239, opened 2017-02-28
Handling empty GH user name

bartaz:
screen shot 2017-02-28 at 11 30 11

Showing GH login when full user name is not available.

### DIFF
--- a/src/common/components/header/index.js
+++ b/src/common/components/header/index.js
@@ -1,5 +1,4 @@
 import React, { Component, PropTypes } from 'react';
-import { connect } from 'react-redux';
 import { Link } from 'react-router';
 
 import { signOut } from '../../actions/auth-store';
@@ -7,7 +6,7 @@ import styles from './header.css';
 
 const wordmark = 'https://assets.ubuntu.com/v1/d45097a4-snapcraft.io-logotype.svg';
 
-export class Header extends Component {
+export default class Header extends Component {
   render() {
     const { authenticated, user } = this.props;
 
@@ -56,5 +55,3 @@ Header.propTypes = {
   }),
   dispatch: PropTypes.func.isRequired
 };
-
-export default connect()(Header);
--- a/src/common/components/header/index.js
+++ b/src/common/components/header/index.js
@@ -7,7 +7,7 @@ import styles from './header.css';
 
 const wordmark = 'https://assets.ubuntu.com/v1/d45097a4-snapcraft.io-logotype.svg';
 
-class Header extends Component {
+export class Header extends Component {
   render() {
     const { authenticated, user } = this.props;
 
@@ -20,7 +20,11 @@ class Header extends Component {
           { authenticated
             ?
               <div className={ styles.sideNav }>
-                { user && <span className={ styles.navItem } >Hi, {user.name}</span> }
+                { user &&
+                  <span className={ styles.username } >
+                    Hi, {user.name || user.login}
+                  </span>
+                }
                 <Link to="/dashboard" className={ styles.link }>Dashboard</Link>
                 <a
                   className={ styles.link }
--- a/src/common/components/user-avatar/index.js
+++ b/src/common/components/user-avatar/index.js
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import { HeadingOne, HeadingThree } from '../vanilla/heading';
 import styles from './user-avatar.css';
 
-class UserAvatar extends Component {
+export class UserAvatarView extends Component {
 
   render() {
     const { auth, user } = this.props;
@@ -29,7 +29,7 @@ class UserAvatar extends Component {
   }
 }
 
-UserAvatar.propTypes = {
+UserAvatarView.propTypes = {
   auth: PropTypes.object,
   user: PropTypes.object
 };
@@ -46,4 +46,4 @@ function mapStateToProps(state) {
   };
 }
 
-export default connect(mapStateToProps)(UserAvatar);
+export default connect(mapStateToProps)(UserAvatarView);
--- a/src/common/components/user-avatar/index.js
+++ b/src/common/components/user-avatar/index.js
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import { HeadingOne, HeadingThree } from '../vanilla/heading';
 import styles from './user-avatar.css';
 
-class UserAvatar extends Component {
+export class UserAvatar extends Component {
 
   render() {
     const { auth, user } = this.props;
--- a/src/common/components/user-avatar/index.js
+++ b/src/common/components/user-avatar/index.js
@@ -17,10 +17,12 @@ class UserAvatar extends Component {
           <img src={ user.avatar_url } />
         </div>
         <div className={ styles.headings }>
-          <HeadingOne>{ user.name }</HeadingOne>
-          <div className={ styles.login }>
-            <HeadingThree>{ user.login }</HeadingThree>
-          </div>
+          <HeadingOne>{ user.name || user.login }</HeadingOne>
+          { user.name &&
+            <div className={ styles.login }>
+              <HeadingThree>{ user.login }</HeadingThree>
+            </div>
+          }
         </div>
       </div>
     );
--- a/src/common/components/user-avatar/index.js
+++ b/src/common/components/user-avatar/index.js
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import { HeadingOne, HeadingThree } from '../vanilla/heading';
 import styles from './user-avatar.css';
 
-export class UserAvatar extends Component {
+class UserAvatar extends Component {
 
   render() {
     const { auth, user } = this.props;
--- a/src/common/components/user-avatar/user-avatar.css
+++ b/src/common/components/user-avatar/user-avatar.css
@@ -12,6 +12,8 @@ $login-colour: $dark-grey;
 
 .avatar > img {
   height: 115px;
+  width: 115px;
+  display: block;
   border-radius: 60px;
 }
 
--- a/src/common/reducers/user.js
+++ b/src/common/reducers/user.js
@@ -1,11 +1,11 @@
 // initialised with server side rendered state
 
-export const DEBUG_SET_USER = 'DEBUG_SET_USER';
+export const UPDATE_USER = 'UPDATE_USER';
 
 export function user(state = null, action) {
   switch(action.type) {
-    // set individual user properties for debugging
-    case DEBUG_SET_USER:
+    // set individual user properties, useful for debugging
+    case UPDATE_USER:
       return {
         ...state,
         ...action.payload
--- a/src/common/reducers/user.js
+++ b/src/common/reducers/user.js
@@ -1,4 +1,16 @@
 // initialised with server side rendered state
-export function user(state = null){
-  return state;
+
+export const DEBUG_SET_USER = 'DEBUG_SET_USER';
+
+export function user(state = null, action) {
+  switch(action.type) {
+    // set individual user properties for debugging
+    case DEBUG_SET_USER:
+      return {
+        ...state,
+        ...action.payload
+      };
+    default:
+      return state;
+  }
 }
--- a/test/unit/src/common/components/user-avatar/t_user-avatar.js
+++ b/test/unit/src/common/components/user-avatar/t_user-avatar.js
@@ -4,17 +4,22 @@ import { shallow } from 'enzyme';
 import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 
-import UserAvatar from '../../../../../../src/common/components/user-avatar';
+import UserAvatar, { UserAvatarView } from '../../../../../../src/common/components/user-avatar';
 import { HeadingOne, HeadingThree } from '../../../../../../src/common/components/vanilla/heading';
 
 const middlewares = [ thunk ];
 const mockStore = configureMockStore(middlewares);
 
-describe('<UserAvatar />', function() {
-  let element;
+describe('<UserAvatar />', () => {
+  const auth = {
+    authorized: true
+  };
+  const user = {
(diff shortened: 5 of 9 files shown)


src/common/components/header/index.js
```diff
@@ -7,7 +7,7 @@ import styles from './header.css';
 
 const wordmark = 'https://assets.ubuntu.com/v1/d45097a4-snapcraft.io-logotype.svg';
 
-class Header extends Component {
+export class Header extends Component {
```
bartaz: To test `Header` without need for the store.

src/common/components/user-avatar/index.js
```diff
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import { HeadingOne, HeadingThree } from '../vanilla/heading';
 import styles from './user-avatar.css';
 
-class UserAvatar extends Component {
+export class UserAvatar extends Component {
```
bartaz: To test `UserAvatar` without the need for the store.

src/common/components/user-avatar/user-avatar.css
```diff
@@ -12,6 +12,8 @@ $login-colour: $dark-grey;
 
 .avatar > img {
   height: 115px;
+  width: 115px;
```
bartaz: To make sure it's not flickering while loading the image.

src/common/reducers/user.js
```diff
@@ -1,4 +1,16 @@
 // initialised with server side rendered state
-export function user(state = null){
-  return state;
+
+export const DEBUG_SET_USER = 'DEBUG_SET_USER';
+
+export function user(state = null, action) {
+  switch(action.type) {
+    // set individual user properties for debugging
+    case DEBUG_SET_USER:
```
bartaz: We don't have any action creators for this (as it's only set serverside), but it was useful when debugging to be able to set/change any user info from Redux dev tools.

src/common/components/header/index.js
katiefenn: We should be binding the store further up the tree and passing in props using JSX instead of exporting twice. Subtle differences like this often lead to difficult-to-debug bugs - especially when some components export to default and some to named exports.

src/common/reducers/user.js
katiefenn: Is the "DEBUG_" prefix necessary? The comment should be sufficient.

src/common/components/header/index.js
bartaz: I'm not sure about moving binding up the tree. This will mean that top level component/container binds the data for all the child components and data is passed down the tree using props. 

Why should parent component bind user and auth if it's only header that needs that data?

There are benefits of both approaches obviously. Binding close to the component that needs the data helps to avoid passing too much data through props. But on the other hand we end up with many components bound to the store.

When it comes to testing and exporting named class.
Yes, it would be possible to remove binding from `Header` and move it up. Actually - that's how it was from the beginning. Now it seems to me that `connect` was only added here to make `dispatch` available (to dispatch logout action). Also, `Header` is rendered only in one place (app container), so it wouldn't be such an issue to pass props to it (as it is done already).

But issue will remain with other components, like `UserAvatar` here.
It's useful for user avatar to have its own binding to the store. Parent component doesn't need to worry about what data UserAvatar needs and why.
In cases of such components, that are bound to the store anyway - we need a way to test them without the need for the store. And pattern with exporting both default bound component and named pure component is very helpful for that.

src/common/reducers/user.js
bartaz: It's not.
I wanted it to be clear that it's meant mostly for debugging. Especially that it doesn't reset user data only overrides single properties, so in theory it doesn't really set user to new object, only updates user props.

But maybe `UPDATE_USER` is a good action name for that. So it may serve for debugging and possible use for future.

src/common/components/header/index.js
katiefenn: I think having two separate public interfaces for components is a mistake. It means we're not testing the same thing we're integrating, and it makes it easy to use the wrong interface in the wrong place. Any other approach is better than this, in my opinion.

I don't see the issue of passing in some props in JSX. It does delegate ownership of data up the tree, but in many cases this is what we want. The app container should say **"show a header component containing a logged-in navigation item"**. This allows for cleaner separation of presentational and container components and makes it easier to test.

bartaz: Maybe the issue here is that we have both presentional and connected component in one file.

Because that's what we do in most cases (where we do bind data):

```js
class SomeComponent extends Component {
  render() {
  }
}

SomeComponent.propTypes = {
}

const mapStateToProps = () => {
}

export default connect(mapStateToProps)(SomeComponent);
```

With such pattern we have a source code of pure react component `SomeComponent`, but we export only the bound one - because that's the one used by the app. These are 2 different components, but only one exported.

If we export both we end up with "two separate public interfaces", but at least it makes it possible to test the pure React one.
The other solution will be to split these into 2 files. One for pure react component, so we can easily import and test it. Another one for data binding only if needed.

Sure, in some cases (like with Header) it's fine to pass couple of props, but in more complicated cases it may become troublesome.

If we take the dashboard, so we really want the top container to handle all the data?
Does dashboard need to know about all the data it's child components need? Or should it only serve as a layout for components that handle data on their own?

We need to find some middle ground here I guess. Not exporting pure React components will likely result in not testing them. I don't know what test setup would be needed to test the connected one (mocking the store and passing it to connect), but it will make testing components much harder than it is for pure React components. And I don't think we should make all our components just pure components and leave data binding to parent container, because it will make containers responsible for all data of their child components which I don't think is good and necessary.

katiefenn: The top container doesn't need to own _all_ the data. Containing elements can be nested. 

Consider the repository list component. A container could be used to own the data the table needs. It could be dropped inside the dashboard page container, or any other page as necessary. It would separate presentational and functional concerns.

I'm arguing for binding _one or more levels above_ presentational elements, not at the _top level_.

test/unit/src/common/components/user-avatar/t_user-avatar.js
```diff
@@ -1,22 +1,41 @@
...
+  let user;
+
+  beforeEach(() => {
+    store = mockStore({
+      auth,
+      user
+    });
+
+    // shallow render container component with the store
+    element = shallow(<UserAvatar store={store} />)
```
bartaz: It's not very straight forward, because shallow render doesn't render child component, we need to find wrapped component by name and shallow render it too. But at least we are testing component connected to the store and we also see if properties are mapped as they should.

src/common/components/header/index.js
nottrobin: I've read this thread, the IRC conversation and [the article](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.5vryhppki) @katiefenn mentioned.

Since I'm still learning React, I don't have a specific opinion about this exact issue, but I do definitely believe that both encapsulation and a sensible separation of concerns are terribly important to try to prevent your codebase from growing into a spaghetti monster.

I very much like the sound of separating "Container" and "Presentational" logic, and also of strictly exporting only one thing from each component, and of the idea that we should be doing our thinking up-front to save future developers from having to think.

Trying to `export` more than one class sounds like it may be effectively trying to use global scope to make your life easier today. @katiefenn's suggestion on passing in properties instead, sounds similar to an example of [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection), which is a widely accepted pattern for keeping neat, encapsulated units which are easier to test.

It is true that this can lead you to sometimes end up injecting an inordinately large number of dependencies. But this can be mitigated by sensibly breaking up your functions into smaller components, and then using [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance).

As I said, I don't understand enough to have a strong opinion about this specific piece of code, but since we are no longer under the gun, I do think it's worth taking the time to come up with patterns which ensure that concerns are separated, encapsulation is clean, and logical structures are kept as simple as possible.

If you do happen to come up with good rules of thumb, feel free to encode them into law by proposing a "javascript.md" or "react.md" into the [webteam practices coding folder](https://github.com/ubuntudesign/webteam-practices/tree/master/coding).

test/unit/src/common/components/user-avatar/t_user-avatar.js
earnubs: Having had a look around and tried it myself, I'm -1 for testing the component with the store, it's essentially a test of redux. If the mapStateToProps function was complex I would export that as a named export along with the component and unit test it too.

src/common/components/header/index.js
earnubs: So I think we're all aiming for the same things, a nice, simple developer environment, that causes us as little bother as possible and allows as much energy to be put on design as possible.

In terms of design, HoC are a form of composition, and for me, breaking components up at this stage is over thinking the problem, but I may not fully understand the problem (I'm somewhat trying to understand it from the solutions people are providing).

Limiting to single exports I can't see the real benefit of at this time, so my vote is to export the component as a named export, the wrapped component as the default and test the named export. Unit testing the wrapped component is really just testing Redux's connect, and it already has unit tests.

We also have code style guides but they're well out of date, as we've been in a constant crunch for over 6 months now. These should be merged into the webteam repo https://github.com/canonical-ols/style-guides

bartaz: @earnubs @katiefenn @nottrobin 

It seems to me that we all agree on the principle of having a clean code and good separation of concerns. Where we differ is the interpretation of how to best implement this in React/Redux world. And unfortunately (or maybe fortunately) React/Redux leave a lot of space for interpretation, because they are not very opinionated and leave to developer how you use them.

I totally agree that we want simple presentational components that only render the data and container components that connect them to data and give more functionality. And that's exactly what React/Redux allows us.

Looking at UserAvatar example:

```js
class UserAvatar extends Component {

  render() {
    const { auth, user } = this.props;

    if (!auth.authenticated) return null;

    return (
      <div className={ styles.container }>
          // render component here...
      </div>
    );
  }
}

UserAvatar.propTypes = {
  auth: PropTypes.object,
  user: PropTypes.object
};
```

This is a presentational user avatar component. All it does is rendering some HTML based on user/auth data it gets in props.

```
function mapStateToProps(state) {
  const {
    auth,
    user
  } = state;

  return {
    auth,
    user
  };
}

export default connect(mapStateToProps)(UserAvatar);
```

This is a container component that connects data from Redux store and passes it to UserAvatar presentational component to render it. That's the "dependency injection".
No other component or container in the app, even if it wants to render user avatar needs to know what kind of data it needs.

Clearly separated concerns in 2 separate components. They just happen to be in one file.

By default we export container component because that's what we use in the app. Presentational component is kept 'hidden' in the module and not visible to outside world.

And now we are back to testing. We didn't do much testing of React/Redux components before because we didn't have time to and when we did it was just for pure presentational components.

Should we just test the container component? Container component is a wrapper, its only functionality is passing the data from the store to presentational component. So I believe we should be able to test presentational component as well, to check if it renders correctly based on data it gets in props.

And yes, separately we can also test container component, but we don't need to check how it renders (assuming we already tested presentational component). We only need to check if it maps the state from the store to the props as presentational component expects. Because that's the function of container component.

But to be able to test them separately, we need to export them both.
Maybe it was mistake we started to put both presentational and container components in one file - but this seemed to be a convention we borrowed from the examples we found.
Should we put them in separate files in same module? That module would still need to export both to make them visible for tests.
Should we separate them into totally separate component modules? In theory it would make sense, but from practical point of view it will make our components structure totally bloated with doubled components (and we would need some good naming convention for those).

test/unit/src/common/components/user-avatar/t_user-avatar.js
bartaz: @earnubs @katiefenn While the discussion we are having is important and interesting from technical point of view it's already quite far away from the purpose of this PR.

Could we focus on what we want to fix to make this PR landed?
I updated the UserAvatar test to test only container component without the need of exporting presentational part of it. It's doable (as in this example), but not very elegant and I kind of agree with @earnubs on that.

If we want an elegant test of presentational component we would need to export it.
And if we don't want to export both presentational and container from the same file we need to move `connect` call somewhere else (to dashboard / select-repositories?) which is also not perfect in my opinion (because these 'pages' shouldn't need to know what data components need).

earnubs: @bartaz I agree with your proposal, export the presentational component, test it in isolation.

bartaz: @katiefenn @earnubs 

Would it be better if we start call presentational component 'Views'? Like:

```js

export class UserAvatarView extends Component {
}

export default connect(...)(UserAvatarView);
```

This would be less confusing when importing.
Of course this leaves the question how do we call purely presentational components that don't have `connect` at all, like vanilla ones, etc.

src/common/components/header/index.js
earnubs: The docs also recommend exporting the component as a named export: http://redux.js.org/docs/recipes/WritingTests.html#connected-components

test/unit/src/common/components/user-avatar/t_user-avatar.js
```diff
@@ -0,0 +1,114 @@
+import React from 'react';
+import expect from 'expect';
+import { shallow } from 'enzyme';
+import configureMockStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
+
+import UserAvatar, { UserAvatarView } from '../../../../../../src/common/components/user-avatar';
```
bartaz: Testing both container and view components.

As suggested in redux docs (http://redux.js.org/docs/recipes/WritingTests.html#connected-components) we may not necessarily need `View` suffix on the name, but maybe it does things a bit more clear?

bartaz:
Merging, with technical aspects to be continued separately in https://trello.com/c/Y32bOilM/324-invesigate-solutions-for-dealing-with-container-and-presentational-components

PR #240, opened 2017-02-28
Tweak icon CSS to match MAAS Vanilla theme

cjwatson:
This puts the asset URL together with the rest of its presentation; and
this code layout works better when we need an icon with an action
associated with it, as we will for snap removals.

While I was at it, I made the asset URL use HTTPS rather than HTTP.

### DIFF
--- a/src/common/components/repository-row/index.js
+++ b/src/common/components/repository-row/index.js
@@ -21,12 +21,6 @@ const FILE_NAME_CLAIM_URL = 'https://myapps.developer.ubuntu.com/dev/click-apps/
 
 const LEARN_THE_BASICS_LINK = 'https://snapcraft.io/docs/build-snaps/your-first-snap';
 const INSTALL_IT_LINK = 'https://snapcraft.io/create/';
-const tickIcon = (
-  <img
-    src='http://assets.ubuntu.com/v1/6c395e6d-green-tick.svg'
-    className={ styles.tickIcon }
-  />
-);
 
 class RepositoryRow extends Component {
 
@@ -221,7 +215,13 @@ class RepositoryRow extends Component {
 
     let caption;
     if (registerNameStatus.success) {
-      caption = <div>{ tickIcon } Registered successfully</div>;
+      caption = (
+        <div>
+          <span
+            className={ `${styles.icon} ${styles.tickIcon}` }
+          /> Registered successfully
+        </div>
+      );
     } else if ( registerNameStatus.error
       && registerNameStatus.error.json
       && registerNameStatus.error.json.payload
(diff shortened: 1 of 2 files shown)


src/common/components/repository-row/index.js
```diff
@@ -21,12 +21,6 @@ const FILE_NAME_CLAIM_URL = 'https://myapps.developer.ubuntu.com/dev/click-apps/
 
 const LEARN_THE_BASICS_LINK = 'https://snapcraft.io/docs/build-snaps/your-first-snap';
 const INSTALL_IT_LINK = 'https://snapcraft.io/create/';
-const tickIcon = (
-  <img
```
bartaz: Seems like instead of removing it it could just be turned into:

```js
const tickIcon = (
   <span className={ `${styles.icon} ${styles.tickIcon}` } />
)
```

Which would simply work without the need to inline the `<span>` with class names everywhere.
Also, it would be easier for later to extract it as separate icon component that we might do.

src/common/components/repository-row/repositoryRow.css
```diff
@@ -8,9 +8,23 @@
   color: $warm-grey;
 }
 
-.tickIcon {
-  width: 22px;
+.icon {
+  width: 16px;
   height: 16px;
+  padding: 0;
+  border-bottom: 0 !important;
```
bartaz: Why the `!important`? It doesn't seem like it inherits border from anything.

src/common/components/repository-row/repositoryRow.css
```diff
@@ -8,9 +8,23 @@
...
+  text-indent: 999em;
+}
+
+.tickIcon {
+  background: url(https://assets.ubuntu.com/v1/6c395e6d-green-tick.svg) no-repeat;
+  background-size: 100% 100%;
+}
+
+span.icon {
+  margin-top: -3px;
```
bartaz: Why `snap.icon` is different from `.icon`? Why the special treatment (and placement)?
Are these a copy of some MAAS styles?

src/common/components/repository-row/repositoryRow.css
cjwatson: This was from MAAS.  I've removed it - as you say, it doesn't seem necessary.

src/common/components/repository-row/repositoryRow.css
cjwatson: Yes, this was from MAAS.  I've removed it for now as it seems fine without it.

src/common/components/repository-row/index.js
cjwatson: Fair enough; done.

Issue #234, opened 2017-02-27
choose repos inputs should be clickable on label


earnubs:
Currently clicking a repo label does not enable the repo, so the hit area is only the checkbox itself.

PR #246, opened 2017-02-28
Filter out enabled repos

earnubs:
Enabled repos should be dimmed, inputs checked and disabled.

### DIFF
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -54,7 +54,7 @@ class SelectRepositoryList extends Component {
   }
 
   renderRepository(repo) {
-    const { fullName } = repo;
+    const { fullName, enabled } = repo;
     const status = this.props.repositoriesStatus[fullName] || {};
     const { selectedRepos } = this.props.selectRepositoriesForm;
 
@@ -67,6 +67,7 @@ class SelectRepositoryList extends Component {
         onChange={ this.onSelectRepository.bind(this, repo) }
         errorMsg= { this.getErrorMessage(status.error) }
         checked={ hasRepository(selectedRepos, repo) }
+        isEnabled={ enabled }
       />
     );
   }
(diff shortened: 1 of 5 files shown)


src/common/components/select-repository-list/index.js
```diff
@@ -86,20 +87,46 @@ class SelectRepositoryList extends Component {
...
     const { selectedRepos } = this.props.selectRepositoriesForm;
-    const { repos } = this.props.repositories;
+    const { repos, success } = this.props.repositories;
     const pageLinks = this.renderPageLinks.call(this);
 
+    this.filterEnabledRepos(repos);
+    let renderedRepos = null;
+
+    if (success) {
+      renderedRepos = this.props.repositories.repos.map(this.renderRepository.bind(this));
```
earnubs: not necessary for this PR, but seemed like an OK improvement to make while here

src/common/components/select-repository-row/index.js
```diff
@@ -8,20 +9,29 @@ class SelectRepositoryRow extends Component {
...
             type="checkbox"
             onChange={ this.onChange.bind(this) }
-            checked={ checked }
+            checked={ checked || isEnabled }
+            disabled={ isEnabled }
           />
         }
         <div>
-          {repository.fullName}
+          <label htmlFor={ repository.fullName }>{repository.fullName}</label>
```
earnubs: fixes #234

src/common/components/select-repository-row/index.js
```diff
@@ -37,12 +47,18 @@ class SelectRepositoryRow extends Component {
   }
 }
 
+SelectRepositoryRow.defaultProps = {
+  checked: false,
```
earnubs: if the props on a form switch from undef to bool, react warns about controlled vrs uncontrolled form elements.

earnubs:
[[add new repo] filter out repositories that are already enabled](https://trello.com/c/jdl05qOn/174-add-new-repo-filter-out-repositories-that-are-already-enabled)

src/common/components/select-repository-list/index.js
```diff
@@ -86,20 +87,46 @@ class SelectRepositoryList extends Component {
     this.props.dispatch(fetchUserRepositories(pageNumber));
   }
 
+  filterEnabledRepos(repositories) {
+    const { success, snaps } = this.props.snaps;
+
+    if (success && snaps.length) {
+      for (let i = repositories.length; i--;) {
+        for (let j = snaps.length; j--;) {
+          let enabledRepo = snaps[j].git_repository_url;
```
cjwatson: `const`, I think.

earnubs: ta

PR #247, opened 2017-03-01
replace remove anchor and text with span only

earnubs:
Text-indent was causing the container to overflow the window.


earnubs:
[Bug: sign in SSO button no longer works](https://trello.com/c/tGpfBNEV/309-bug-sign-in-sso-button-no-longer-works)

PR #248, opened 2017-03-01
161: Landing page CTA authenticated text

katiefenn:


src/common/containers/landing.js
```diff
@@ -42,9 +42,16 @@ class Landing extends Component {
...
-                  Set up in minutes
-                </Anchor>
+                { !this.props.auth.authenticated &&
+                  <Anchor href="/auth/authenticate" icon={ octocat } flavour='embiggened' appearance='positive' >
+                    Set up in minutes
+                  </Anchor>
+                }
+                { this.props.auth.authenticated &&
+                  <div>
+                    Hi { this.props.user.name }, <a href="/dashboard">let’s check out your repos</a>.
```
bartaz: This will be prone to the same issue we solve in #239

`user.name` may be empty in GH, so we need to fall back to `user.login` in such case.

cjwatson: Should be `{ this.props.user.name || this.props.user.login }`, as in #239.

src/common/containers/landing.js
```diff
@@ -42,9 +42,16 @@ class Landing extends Component {
               </ul>
 
               <div className={ styles.bannerButton }>
-                <Anchor href="/auth/authenticate" icon={ octocat } flavour='embiggened' appearance='positive' >
-                  Set up in minutes
-                </Anchor>
+                { !this.props.auth.authenticated &&
```
bartaz: Wouldn't that be a bit more clear as `{ this.props.auth.authenticated ? ... : ... }`? 

I as one statement it will be more close to `if ... else` than it is now as 2 separate conditions.

PR #250, opened 2017-03-01
Clear snap cache when name registered

bartaz:
Launchpad sets the `store_name` for a snap once it's authorized.
So after a successful authorize call we need to clear all snap related caches to make sure they will be updated with fresh data about snap.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -12,6 +12,8 @@ If true, set secure session cookies.
 
 ### MEMCACHED\_HOST
 
+- **Example**: `localhost:11211`
+
 The host name of the memcached service.
 
 ### MEMCACHED\_SESSION\_SECRET
--- a/docs/memcached.md
+++ b/docs/memcached.md
@@ -1,0 +1,14 @@
+# Memchached entries
+
+## `url_prefix:https://github.com/:owner/`
+
+Caches list of snaps (enabled repositories) for given GitHub user (via URL prefix).
+
+## `url:https://github.com/:owner/:name`
+
+Caches snap data for given repository URL.
+
+## `snapcraft_data:https://github.com/:owner/:name`
+
+Caches JSON subset of data from `snapcraft.yaml` found in given repository.
+Currently it contains only a snap name.
--- a/docs/memcached.md
+++ b/docs/memcached.md
@@ -1,14 +1,14 @@
-# Memchached entries
+# Memcached entries
 
-## `url_prefix:https://github.com/:owner/`
+## `lp:url_prefix:https://github.com/:owner/`
 
 Caches list of snaps (enabled repositories) for given GitHub user (via URL prefix).
 
-## `url:https://github.com/:owner/:name`
+## `lp:url:https://github.com/:owner/:name`
 
 Caches snap data for given repository URL.
 
-## `snapcraft_data:https://github.com/:owner/:name`
+## `lp:snapcraft_data:https://github.com/:owner/:name`
 
 Caches JSON subset of data from `snapcraft.yaml` found in given repository.
 Currently it contains only a snap name.
--- a/src/common/containers/build-details.js
+++ b/src/common/containers/build-details.js
@@ -81,7 +81,7 @@ class BuildDetails extends Component {
         { isFetching &&
           <div className={styles.spinner}><Spinner /></div>
         }
-        { snap &&
+        { snap && snap.store_name &&
           <HelpInstallSnap
             headline='To test the latest successful build on your PC or cloud instance:'
             name={ snap.store_name }
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -439,6 +439,18 @@ export const findSnap = (req, res) => {
     .catch((error) => sendError(res, error));
 };
 
+const memcachedDel = (cachedId) => {
+  return new Promise((resolve, reject) => {
+    getMemcached().del(cachedId, (err) => {
+      if (err) {
+        logger.error(`Error deleting ${cachedId} from memcached:`, err);
+        reject(err);
+      }
+      resolve();
+    });
+  });
+};
+
 const clearSnapCache = (repositoryUrl) => {
   const repository = parseGitHubUrl(repositoryUrl);
   const enabledReposCacheId = getUrlPrefixCacheId(getRepoUrlPrefix(repository.owner));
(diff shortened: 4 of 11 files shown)


docs/environment-variables.md
```diff
@@ -12,6 +12,8 @@ If true, set secure session cookies.
 
 ### MEMCACHED\_HOST
 
+- **Example**: `localhost:11211`
```
bartaz: It took me a while to realize there shouldn't be `http://` in front of it...

docs/memcached.md
```diff
@@ -0,0 +1,14 @@
+# Memchached entries
```
bartaz: Some basic docs with memcached entries we use (at least those that I touched in this PR).

src/common/containers/build-details.js
```diff
@@ -81,7 +81,7 @@ class BuildDetails extends Component {
         { isFetching &&
           <div className={styles.spinner}><Spinner /></div>
         }
-        { snap &&
+        { snap && snap.store_name &&
```
bartaz: There still may be a case when user sees this page before snap has a name (before name is registered), so we for sure shouldn't show the box in such case.

We still have separate card to properly hide this box until snap is successfully published.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -8,7 +8,7 @@ import {
   resetMemcached,
   setupInMemoryMemcached
 } from '../../../../../src/server/helpers/memcached';
-import getSnapNameCacheId from '../../../../../src/server/helpers/github';
+import { getSnapNameCacheId } from '../../../../../src/server/handlers/github';
```
bartaz: Why other tests that use this didn't fail when it was undefined? ... it's a mystery.

docs/environment-variables.md
katiefenn: 👍 

We should discuss bringing the docs up to (and improving on) the standard of my.ubuntu.com

docs/memcached.md
cjwatson: "Memchached" is a typo.

docs/memcached.md
```diff
@@ -0,0 +1,14 @@
...
+
+## `url_prefix:https://github.com/:owner/`
```
cjwatson: They actually all have `lp:` prefixed (see `src/server/helpers/memcached.js`).  May be worth prefixing that here and below too for the sake of operators.

src/server/handlers/launchpad.js
```diff
@@ -435,6 +437,30 @@ export const findSnap = (req, res) => {
...
+  getMemcached().del(snapCacheId, (err) => {
+    if (err) {
+      logger.error(`Error deleting ${snapCacheId} from memcached:`, err);
+    }
+  });
+  getMemcached().del(snapNameCacheId, (err) => {
+    if (err) {
+      logger.error(`Error deleting ${snapNameCacheId} from memcached:`, err);
+    }
+  });
```
cjwatson: Won't this result in unpredictable behaviour in tests if the delete operation takes any time?  IMO this function should return a Promise that resolves only when all the deletions are done.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -952,6 +952,68 @@ describe('The Launchpad API endpoint', () => {
...
+                    getMemcached().get(getUrlPrefixCacheId(urlPrefix), (err, result) => {
+                      expect(result).toNotExist();
+                    }),
+                    getMemcached().get(getRepositoryUrlCacheId(repositoryUrl), (err, result) => {
+                      expect(result).toNotExist();
+                    }),
+                    getMemcached().get(getSnapNameCacheId(repositoryUrl), (err, result) => {
+                      expect(result).toNotExist();
+                    })
+                  ])
```
cjwatson: For the sake of concise tests, it might be better to just poke into the internals of the test helper that we know is in place, e.g.:

    expect(getMemcached().cache).toExcludeKey(getUrlPrefixCacheId(urlPrefix));

src/server/handlers/launchpad.js
bartaz: I didn't want to delay the response on memcached calls, but yes - in theory this could fail in tests (as memcached may be cleared after response is returned).

bartaz:
It would be good to have #253 merged before this one as it will simplify some bits.

src/server/handlers/launchpad.js
```diff
@@ -435,6 +437,33 @@ export const findSnap = (req, res) => {
...
+  const snapCacheId = getRepositoryUrlCacheId(repositoryUrl);
+  const snapNameCacheId = getSnapNameCacheId(repositoryUrl);
+
+  logger.info(`Clearing caches for ${repositoryUrl}: ${enabledReposCacheId}, ${snapCacheId}, ${snapNameCacheId}`);
+
+  return Promise.all([
+    memcachedDel(enabledReposCacheId),
+    memcachedDel(snapCacheId),
+    memcachedDel(snapNameCacheId)
+  ]);
```
katiefenn: I think snap names are cached twice, once on their own and again with the rest of the snap data. Is it worth consolidating these cache items?

test/routes/src/server/routes/t_launchpad.js
katiefenn: iiam

src/server/handlers/launchpad.js
```diff
@@ -437,6 +439,21 @@ export const findSnap = (req, res) => {
...
+  const snapNameCacheId = getSnapNameCacheId(repositoryUrl);
+
+  logger.info(`Clearing caches for ${repositoryUrl}: ${enabledReposCacheId}, ${snapCacheId}, ${snapNameCacheId}`);
+
+  return Promise.all([
+    getMemcached().del(enabledReposCacheId),
+    getMemcached().del(snapCacheId),
+    getMemcached().del(snapNameCacheId)
+  ]);
+};
```
katiefenn: I think snap names are cached twice, once on their own and again with the rest of the snap data. Is it worth consolidating these cache items?

bartaz: These are caches for 2 different LP API calls. One is listing of all snaps (enabled repos) for given user, and one is a single snap for given repository.

Yes, this data overlaps (for one user same snap is kept in both), but optimizing it may introduce unnecessary complexity in cache layer.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -944,6 +944,58 @@ describe('The Launchpad API endpoint', () => {
...
+
+              setupInMemoryMemcached();
+              // find snap by url cache will be filled by API call
+
+              // fill snap listing and snapcraft.yaml data caches
+              getMemcached().set(getUrlPrefixCacheId(urlPrefix), [snap]);
+              getMemcached().set(
+                getSnapNameCacheId(repositoryUrl),
+                { name: 'test-snap-name' }
+              );
```
cjwatson: These should just do `getMemcached().cache[...] = ...;` rather than getting promises involved.

test/routes/src/server/routes/t_launchpad.js
```diff
@@ -944,6 +944,58 @@ describe('The Launchpad API endpoint', () => {
                 done(err);
               });
           });
+
+          context('when snaps are memcached', () => {
+            const urlPrefix = 'https://github.com/anowner/';
+            const repositoryUrl = 'https://github.com/anowner/aname';
+            let snap;
+
+            before(() => {
```
cjwatson: Maybe `beforeEach` and `afterEach` here to avoid confusion in case somebody wants to add another test to this `context` block.

PR #252, opened 2017-03-01
request user snaps before redirecting to dashboard from

earnubs:
In the case where a user has no enabled repos, and visits dashboard/selected-repositories, the `repositories-list` component onMount has stale props. `mapStateToProps` is called both too early (when we indeed have no snaps) and too late (already redirected), and as such the component redirects back to dashboard/selected-repositories because the list of snaps is 0, but we've had a successful fetch.

By dispatching the fetchUserSnaps action early, when we know we're going to redirect, the `repositories-list` component has up to date props and does the right thing. In order to avoid fetching when we're already fetching, we check in the action if a dispatch is required based on `[snaps: isFetching]`.

This is a solution, not the solution. The solution involved figuring out why mapStateToProps isn't being called at the point we'd like it to, which will take a little more investigation.

## To QA

* remove all snaps
* enable 1 snap
* add selected snap
* dashboard/select-repositories should redirect to /dashboard (and not bounce back)

### DIFF
--- a/src/common/actions/snaps.js
+++ b/src/common/actions/snaps.js
@@ -42,6 +42,20 @@ export function fetchUserSnaps(owner) {
   };
 }
 
+function shouldFetchUserSnaps(state) {
+  const snaps = state.snaps;
+
+  return !snaps.isFetching;
+}
+
+export function fetchUserSnapsIfNeeded(owner) {
+  return function(dispatch, getState) {
+    if (shouldFetchUserSnaps(getState())) {
+      return dispatch(fetchUserSnaps(owner));
+    }
+  };
+}
+
 export function fetchSnapsSuccess(snaps) {
   return {
     type: FETCH_SNAPS_SUCCESS,
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -36,7 +36,7 @@ export class SelectRepositoryListComponent extends Component {
     const owner = nextProps.user.login;
     if (ids.length && ids.every((id) => repositoriesStatus[id].success)) {
       this.props.dispatch(fetchUserSnapsIfNeeded(owner));
-      this.props.router.push('/dashboard/');
+      this.props.router.push('/dashboard');
     }
   }
 
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -14,6 +14,7 @@ import PageLinks from '../page-links';
 import Button, { LinkButton } from '../vanilla/button';
 import { HeadingThree } from '../vanilla/heading';
 import { fetchUserRepositories } from '../../actions/repositories';
+import { fetchUserSnapsIfNeeded } from '../../actions/snaps';
 import { hasRepository } from '../../helpers/repositories';
 import styles from './styles.css';
 
@@ -32,8 +33,10 @@ export class SelectRepositoryListComponent extends Component {
   componentWillReceiveProps(nextProps) {
     const repositoriesStatus = nextProps.repositoriesStatus;
     const ids = Object.keys(repositoriesStatus);
+    const owner = nextProps.user.login;
     if (ids.length && ids.every((id) => repositoriesStatus[id].success)) {
-      this.props.router.push('/dashboard');
+      this.props.dispatch(fetchUserSnapsIfNeeded(owner));
+      this.props.router.push('/dashboard/');
     }
   }
 
@@ -86,6 +89,7 @@ export class SelectRepositoryListComponent extends Component {
 
   onSubmit() {
     const { selectedRepos } = this.props.selectRepositoriesForm;
+
     if (selectedRepos.length) {
       this.props.dispatch(createSnaps(selectedRepos));
     }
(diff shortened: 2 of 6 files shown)


src/common/components/select-repository-list/index.js
```diff
@@ -172,29 +176,34 @@ export class SelectRepositoryListComponent extends Component {
...
+  onSelectRepository: PropTypes.func,
   repositories: PropTypes.object,
   repositoriesStatus: PropTypes.object,
-  selectRepositoriesForm: PropTypes.object,
-  onSelectRepository: PropTypes.func,
   router: PropTypes.object.isRequired,
-  dispatch: PropTypes.func.isRequired
+  selectRepositoriesForm: PropTypes.object,
+  snaps: PropTypes.object,
+  user: PropTypes.object
```
earnubs: Just me sorted alphabetically.

src/common/components/select-repository-list/index.js
```diff
@@ -32,8 +33,10 @@ export class SelectRepositoryListComponent extends Component {
   componentWillReceiveProps(nextProps) {
     const repositoriesStatus = nextProps.repositoriesStatus;
     const ids = Object.keys(repositoriesStatus);
+    const owner = nextProps.user.login;
     if (ids.length && ids.every((id) => repositoriesStatus[id].success)) {
-      this.props.router.push('/dashboard');
+      this.props.dispatch(fetchUserSnapsIfNeeded(owner));
```
earnubs: This is the fix, everything else is really to support this line.

earnubs:
[Bug: Adding repo lands back on the add repo page](https://trello.com/c/JjrJINu8/308-bug-adding-repo-lands-back-on-the-add-repo-page)

src/common/actions/snaps.js
```diff
@@ -42,6 +42,20 @@ export function fetchUserSnaps(owner) {
...
 
+function shouldFetchUserSnaps(state) {
+  const snaps = state.snaps;
+
+  return !snaps.isFetching;
+}
+
+export function fetchUserSnapsIfNeeded(owner) {
+  return function(dispatch, getState) {
+    if (shouldFetchUserSnaps(getState())) {
```
bartaz: I'm not sure if it is a good place to do this check (and to access the state in the action creator).

Accessing the state in action is an anti-pattern (http://stackoverflow.com/a/35674575).
Even though this use case seem to fit the "conditional dispatch" exception that allows checking the state in action it still introduces a bit of dependency between action and the store. This action here needs to know where in the store we keep the data related to this particular fetch.

On the other hand every fetch should be 'conditional' - we should never fetch data that we already requested (`isFetching` is true). Right now we were checking it in the component that was requesting the data. But maybe we should aim to refactor all the "fetch" actions to use the state and check if `isFetching` is set?

I guess this comment doesn't provide any useful opinion ;)
I'm not agains this change, but it introduces some new patters we didn't have in the app before.

src/common/components/select-repository-list/index.js
```diff
@@ -32,8 +33,10 @@ export class SelectRepositoryListComponent extends Component {
...
+      this.props.router.push('/dashboard/');
```
bartaz: Routes define `/dashboard` without trailing slash, so it would be good to keep it the same here.

test/unit/src/common/actions/t_snaps.js
```diff
@@ -21,10 +22,12 @@ const mockStore = configureMockStore(middlewares);
 
 describe('repositories actions', () => {
   const initialState = {
-    isFetching: false,
-    success: false,
-    error: null,
-    snaps: null
+    snaps: {
```
bartaz: Oh... so it seems we do it wrong in all the other actions tests, because it's not initial state of single reducer, but whole store...

And only because we don't do anything with initial state it doesn't break anything. Sounds like another technical debt to fix.

bartaz: Created a card for that: https://trello.com/c/nnpyx939/320-actions-tests-use-wrong-initialstate

src/common/actions/snaps.js
earnubs: Can you show me the pattern we're using here? The pattern I've used is from redux docs, and also what I used in the prototype app before mu https://github.com/reactjs/redux/blob/master/examples/async/src/actions/index.js#L46

src/common/components/select-repository-list/index.js
earnubs: Oops, testing hangover, didn't notice it in diff :(

src/common/actions/snaps.js
bartaz: No sure if `auth` is a good example but for example here: https://github.com/canonical-ols/build.snapcraft.io/blob/master/src/common/components/repositories-list/index.js#L23-L37

All I mean is that by this time we had all the logic and conditional dispatching in components that were using the data.

But if it's a pattern from redux docs then I'm less scared that we try to do something unusual.
And, as mentioned above I think in theory we would need that for all the fetch actions.
Maybe it could even be the default - so instead of `fetchSnapsIfNeeded` being conditional simply `fetchSnaps` should fetch only if we are not fetching the data already.

But that's something we can think about later when we start to use this pattern more.

src/common/components/select-repository-list/index.js
```diff
@@ -32,7 +33,9 @@ export class SelectRepositoryListComponent extends Component {
   componentWillReceiveProps(nextProps) {
     const repositoriesStatus = nextProps.repositoriesStatus;
     const ids = Object.keys(repositoriesStatus);
+    const owner = nextProps.user.login;
     if (ids.length && ids.every((id) => repositoriesStatus[id].success)) {
+      this.props.dispatch(fetchUserSnapsIfNeeded(owner));
```
bartaz: This feels a bit like a workaround rather then solution, as it's quite clear that the problem is in the dashboard (we trigger redirect before we properly receive the data).

I had a look at the component flow there and indeed it's tricky. The `fetchUserSnaps` action is dispatched (so in theory it sets `isFetching` to true and `success` to false in the store), but the component still calls `fetchData` with its current state (success with zero repositories).

So it seem that the problematic line is this one:
https://github.com/canonical-ols/build.snapcraft.io/blob/master/src/common/components/repositories-home/index.js#L42

Removing it seems to fix the bug. By removing it we never trigger `fetchData` directly from `componentDidMount`, only from `componentWillRecieveProps`. In `componentDidMount` we always fetch user snaps, so eventually store will be updated and component will receive new props so it can decide if redirect should be made or builds should be fetched.

Unless I missed something and it will break in some other case...

bartaz:
I wasn't suggesting getting rid of the whole thing ;)

The `IfNeeded` pattern seemed like something that would be useful for us to adapt.

earnubs:
superseded by #254

PR #253, opened 2017-03-01
Convert all uses of memcached to promises

cjwatson:
This is much easier to use, especially in conjunction with other code
using promises.  The test stubs already used promises, but this had a
dangerous gotcha: code that expected memcached methods to return
promises would work under test but fail in production, as was the case
with the webhook implementation.  Using promises everywhere avoids this
problem.

I considered using one of the existing memcached/promise integration
libraries, but all the ones I could find depended on some specific
promise library, and it's so easy to do by hand that it wasn't worth the
hassle.

### DIFF
--- a/src/server/handlers/github.js
+++ b/src/server/handlers/github.js
@@ -92,7 +92,9 @@ export const getSnapcraftData = (repositoryUrl, token) => {
   const cacheId = getSnapNameCacheId(repositoryUrl);
 
   return getMemcached().get(cacheId)
-    .catch(() => undefined)
+    .catch((err) => {
+      logger.error(`Error getting ${cacheId} from memcached:`, err);
+    })
     .then((result) => {
       if (result !== undefined) {
         return result;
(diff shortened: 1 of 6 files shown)


src/server/handlers/github.js
```diff
@@ -91,11 +91,11 @@ export const getSnapcraftData = (repositoryUrl, token) => {
   const { owner, name } = parseGitHubRepoUrl(repositoryUrl);
   const cacheId = getSnapNameCacheId(repositoryUrl);
 
-  return new Promise((resolve) => {
-    getMemcached().get(cacheId, (err, result) => {
-
-      if (!err && result !== undefined) {
-        return resolve(result);
+  return getMemcached().get(cacheId)
+    .catch(() => undefined)
```
bartaz: This looks quite strange to silently ignore error and return `undefined`.
It at least asks for a comment in code to explain why it is like that or log error.

And as far as I can see you use such pattern later in launchpad handler:

```
.catch((err) => {
  logger.error(`Error deleting ${cacheId} from memcached:`, err);		            
})
```

This would handle the error at least by logging it and also (silently) return undefined.

src/server/handlers/webhook.js
```diff
@@ -64,6 +64,7 @@ export const notify = (req, res) => {
     // XXX cjwatson 2017-02-16: We could be smarter about this by looking at
     // the content of the push event.
     return getMemcached().del(getSnapNameCacheId(repositoryUrl))
+      .catch(() => undefined)
```
bartaz: As mentioned earlier - instead of silently ignoring error and returning undefined I'd prefer at least logging the error.

PR #255, opened 2017-03-02
237: Build status for unbuilt repository

katiefenn:
- Replaced status prop with a colour prop
- Added extra "Never built" build status for unconfigured repos

### DIFF
--- a/src/common/components/build-status/index.js
+++ b/src/common/components/build-status/index.js
@@ -2,24 +2,16 @@ import React, { PropTypes } from 'react';
 import { Link } from 'react-router';
 import moment from 'moment';
 
-import { BuildStatusConstants } from '../../helpers/snap-builds';
-
 import styles from './buildStatus.css';
 
 const BuildStatus = (props) => {
 
   const {
     link,
-    status,
+    colour,
     statusMessage,
     dateStarted
   } = props;
-
-  const statusStyle = {
-    [BuildStatusConstants.SUCCESS]: styles.success,
-    [BuildStatusConstants.ERROR]: styles.error,
-    [BuildStatusConstants.PENDING]: styles.pending
-  };
 
   let humanDateStarted;
 
@@ -35,7 +27,7 @@ const BuildStatus = (props) => {
   }
 
   return (
-    <div className={ `${styles.buildStatus} ${statusStyle[status]}` }>
+    <div className={ `${styles.buildStatus} ${styles[colour]}` }>
       { link
         ? <Link to={link}>{statusMessage}</Link>
         : <span>{statusMessage}</span>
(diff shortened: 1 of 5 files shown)


src/common/components/build-status/index.js
```diff
@@ -47,7 +39,7 @@ const BuildStatus = (props) => {
 
 BuildStatus.propTypes = {
   link: PropTypes.string,
-  status:  PropTypes.oneOf(['success', 'pending', 'error']),
+  colour: PropTypes.oneOf(['green', 'yellow', 'red', 'grey']),
```
katiefenn: I've replaced the "success", "pending", "error" statuses with a straight map to colours. The spec states that grey and blue will be added, and so colours will not map directly to the old Vanilla status colours.

src/common/components/repository-row/index.js
```diff
@@ -347,14 +347,20 @@ class RepositoryRow extends Component {
...
               statusMessage={ latestBuild.statusMessage }
               dateStarted={ latestBuild.dateStarted }
             />
           }
+          { !snap.snapcraft_data &&
+            <BuildStatus
+              colour="grey"
+              statusMessage="Never built"
+            />
+          }
```
katiefenn: Because the build data and snap data are stored in separate data structures in the store, this rather inelegant switching is the best that can be done for now.

It would be nice to redesign the store to unify snap and build data into a single data structure so that build status can be determined by a reducer instead of in the component.

katiefenn:
screen shot 2017-03-01 at 16 37 31

src/common/helpers/snap-builds.js
```diff
@@ -24,28 +24,29 @@
...
+  GREEN: 'green',
+  YELLOW: 'yellow',
+  RED: 'red',
+  GREY: 'grey'
 };
 
 // Based on BuildStatusConstants from LP API
 // https://git.launchpad.net/launchpad/tree/lib/lp/buildmaster/enums.py#n22
 //
 // mapping between build status from LP and pending/success/error internal status
```
bartaz: This comment should be updated to reflect the changes.

src/common/components/repository-row/index.js
```diff
@@ -347,14 +347,20 @@ class RepositoryRow extends Component {
...
+          { latestBuild && snap.snapcraft_data &&
             <BuildStatus
               link={ `/${fullName}/builds/${latestBuild.buildId}`}
-              status={ latestBuild.status }
+              colour={ latestBuild.status }
               statusMessage={ latestBuild.statusMessage }
               dateStarted={ latestBuild.dateStarted }
             />
           }
+          { !snap.snapcraft_data &&
```
bartaz: It would be better to have both of the conditions joined together to be sure the result is always one `<BuildStatus>`. 
When we have them separate like this it can lead to mistake when we change the condition on only on of them.
If the conditional is too complex to keep in JSX maybe it's worth extracting it to a variable or function the will return `BuildStatus` depending on the state.

PR #261, opened 2017-03-02
237: Fix prop-type warnings on build detail pages

katiefenn:


src/common/components/build-row/index.js
```diff
@@ -12,7 +12,7 @@ const BuildRow = (props) => {
     architecture,
     buildId,
     duration,
-    status,
+    colour,
```
katiefenn: Even though users aren't linked through to this page any more, I decided it's worth fixing in case we make it acessibile in the future.

src/common/components/repository-row/index.js
```diff
@@ -415,7 +415,16 @@ class RepositoryRow extends Component {
...
+        <Data col="27">
+          { hasBuilt
+            ? (
+              <Link to={ `/${fullName}/builds` }>{ fullName }</Link>
+            )
+            : (
+              <span>{ fullName }</span>
+            )
+          }
+        </Data>
```
katiefenn: Don't link users to the build pages if the repository has never been built.

PR #262, opened 2017-03-03
Detect name mismatch (and dropdowns refactoring)

bartaz:
Implementation of a first part of handling name mismatch between snapcraft.yaml and registered name. As in https://trello.com/c/zvLmZuMR/322-handle-registered-name-not-matching-snap-name-pt-1

screen shot 2017-03-03 at 10 21 11

This branch also includes refactoring of RepositoryRow code. Mostly moving dropdowns (and icons) to their separate modules.


src/common/components/repository-row/dropdowns/remove-repo-dropdown.js
```diff
@@ -0,0 +1,67 @@
...
+  let warningText;
+  if (latestBuild) {
+    warningText = (
+      'Removing this repo will delete all its builds and build logs.'
+    );
+  } else {
+    warningText = (
+      'Are you sure you want to remove this repo from the list?'
+    );
+  }
```
katiefenn: We could tidy up the RemoveRepoDropdown function by extractng the warning text to a function.

src/common/components/repository-row/dropdowns/unregistered-dropdown.js
```diff
@@ -0,0 +1,171 @@
...
+            onClick={actionOnClick}
+            isSpinner={actionSpinner}
+          >
+            { actionText }
+          </Button>
+        </div>
+      </Row>
+    </Dropdown>
+  );
+};
```
katiefenn: This function could really benefit from being broken down into separate functions. It's still fine to keep it in the same file, I think.

src/common/components/repository-row/dropdowns/remove-repo-dropdown.js
bartaz: Looking at it now it seems it could be moved up to the parent component and only passed as a string here.

This will remove the need for a bunch of store props passed into this component.

Issue #360, opened 2017-03-03
Add metric for GitHub accounts that have ever logged in [1]


nottrobin:
This is complicated because we don't quite keep state at this level, only once they actually create an account, and switch deploys mean that there isn't anywhere to keep that state right now.  Choices seem to be having a local Prometheus instance which handles the aggregation and federates with IS Prometheus, or possibly a little PostgreSQL instance which we can keep account settings in (which I suspect we may end up needing eventually, but it's probably more work than justified in the short term).

- [x] Add PostgreSQL instance to deployments
- [x] Add basic local database integration (Bookshelf.js)
- [x] Update charm to generate knex configuration based on PostgreSQL relation
- [x] Run migrations on each deployment
- [x] Add `/metrics` endpoint
- [x] Export GitHub user count metric
- [x] Get firewall access opened from IS Prometheus to BSI appservers
- [x] Connect metrics to `kpi.canonical.com`

---

Copied from Trello card: https://trello.com/c/8OsVtwzz/317-add-metric-for-github-accounts-that-have-ever-logged-in
Labels: MWC priority

cjwatson:
https://portal.admin.canonical.com/100730 is the next stage of this.

cjwatson:
There's a panel for this on https://snappy.kpi.canonical.com/dashboard/db/incoming-wip-store-kpis now.  It's a bit minimal, but we can refine it from here.

PR #371, opened 2017-03-05
Store GitHub user information in a database

cjwatson:
We now use the Bookshelf ORM and the Knex query builder, which gives us
a promise-based interface to multiple databases with minimal
boilerplate.

In development, we use SQLite by default, since that involves minimal
setup.  In staging and production, we use PostgreSQL, so this comes with
charm changes to ensure that we have access to a suitable database.

At the moment, user information is saved to both the session and the
database.  This is obviously duplicative, and we should perhaps consider
having just an ID in the session and fetching the rest from the database
as needed.  Similarly, perhaps some of the things we use memcached for
would make more sense in a more persistent database instead.

Part of #360.

### DIFF
--- a/migrations/20170302155852_user.js
+++ b/migrations/20170302155852_user.js
@@ -8,6 +8,7 @@ exports.up = function(knex, Promise) {
       table.text('name');
       table.text('login').notNullable();
       table.timestamps();
+      table.timestamp('last_login_at').notNullable();
     })
   ]);
 };
--- a/migrations/20170302155852_user.js
+++ b/migrations/20170302155852_user.js
@@ -1,0 +1,20 @@
+exports.up = function(knex, Promise) {
+  return Promise.all([
+    knex.schema.createTable('github_user', function(table) {
+      // The GitHub ID would do as a primary key, but this approach works
+      // better with Bookshelf's `Model.isNew` method.
+      table.increments();
+      table.integer('github_id').notNullable();
+      table.text('name');
+      table.text('login').notNullable();
+      table.text('avatar_url');
+      table.timestamps();
+    })
+  ]);
+};
+
+exports.down = function(knex, Promise) {
+  return Promise.all([
+    knex.schema.dropTable('github_user')
+  ]);
+};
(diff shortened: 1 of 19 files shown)


cjwatson:
Temporarily blocked because https://code.launchpad.net/~cjwatson/canonical-mojo-specs/snap-build-add-postgresql/+merge/318989 needs to land first and be deployed to both staging and production, otherwise we'll be in for pain; but I might as well get this up for review now.

cjwatson:
We have PostgreSQL units deployed to both staging and production now, so this is no longer blocked.

src/server/db/models/github-user.js
```diff
@@ -0,0 +1,6 @@
+import db from '../index';
+
+export const GitHubUser = db.Model.extend({
+  tableName: 'github_user',
```
bartaz: Is the schema of the table only defined in the migration file?

It would be useful to be able to find available fields in model file.

migrations/20170302155852_user.js
```diff
@@ -0,0 +1,20 @@
+exports.up = function(knex, Promise) {
+  return Promise.all([
+    knex.schema.createTable('github_user', function(table) {
+      // The GitHub ID would do as a primary key, but this approach works
+      // better with Bookshelf's `Model.isNew` method.
+      table.increments();
+      table.integer('github_id').notNullable();
+      table.text('name');
+      table.text('login').notNullable();
+      table.text('avatar_url');
```
bartaz: Do we need `avatar_url` stored?
We use it in UI, but we get it from GitHub through session, so not sure if we need to persist it.

On the other hand, as main reason for this table now is login metrics, would it be worth adding some `last_login_date` so we would be able to see which accounts are active?
But this of course may evolve to some bigger "activities" table later...

src/server/db/models/github-user.js
cjwatson: Yeah, this is an IMO somewhat annoying design decision of Bookshelf.  I've added a block comment as a workaround.

PR #372, opened 2017-03-06
Fixes #266: Fixed mocked requests in login route tests

### LINKED ISSUE #266: Tests randomly failing in dev environment
Some routes tests fail randomly on before/after hooks with message "Cannot read property 'call' of undefined".

They pass just fine on Travis.

This is possibly related to nock mocks. Maybe not all needed requests are properly mocked, or not mocks are satisfied? But this is just guessing.

```
  login routes
    authenticate
      with no options
        ✓ should redirect from /login/authenticate to SSO
        ✓ should include verify url in redirect header
        ✓ should not include macaroon extension in redirect header
      with options
        ✓ should redirect from /login/authenticate to SSO
        ✓ should include verify url in redirect header
        ✓ should include macaroon extension in redirect header
    sso-discharge
      if session has discharge macaroon
        ✓ GET returns the macaroon
        ✓ DELETE deletes the macaroon
      if user is logged in but has no SSO discharge
        ✓ GET returns 404
        1) "after each" hook
      2) "before each" hook

  The store API endpoint
    3) "after each" hook
    register-name route

  145 passing (975ms)
  3 failing

  1) login routes "after each" hook:
     TypeError: Cannot read property 'call' of undefined
  

  2) login routes sso-discharge if user is logged in but has no SSO discharge "before each" hook:
     TypeError: Cannot read property 'call' of undefined
  

  3) login routes "after each" hook:
     TypeError: Cannot read property 'call' of undefined
```

---

Copied from Trello card: https://trello.com/c/E2R3xwVI/247-tests-randomly-failing-in-dev-environment
Labels:

bartaz:
Fix for #266

Login route tests were failing randomly on some dev environments.

It turned out it was issue with usage of `nock`.

Tests for authentication routes were mocking SSO requests, but these mock were also being created (and never called) for discharge macaroon routes. And for some reason it made nock go nuts sometimes.

This PR moves mocks into authentication context, so they are only created when they are needed.
Also added checks if all mocks are satisfied.

### DIFF
--- a/test/routes/src/server/routes/t_login.js
+++ b/test/routes/src/server/routes/t_login.js
@@ -67,8 +67,8 @@ describe('login routes', () => {
       it('should redirect from /login/authenticate to SSO', () => {
         return supertest(app)
           .get('/login/authenticate')
-          .expect(res => {
-            expect(res.statusCode).toEqual(302);
+          .expect(302)
+          .then((res) => {
             const expectedBaseUrl = url.parse(UBUNTU_SSO_URL);
             expect(url.parse(res.header.location)).toMatch({
               protocol: expectedBaseUrl.protocol,
@@ -81,7 +81,7 @@ describe('login routes', () => {
       it('should include verify url in redirect header', () => {
         return supertest(app)
           .get('/login/authenticate')
-          .expect(res => {
+          .then((res) => {
             const parsedLocation = url.parse(res.header.location, true);
             expect(parsedLocation.query['openid.return_to'])
               .toEqual(OPENID_VERIFY_URL);
@@ -93,7 +93,7 @@ describe('login routes', () => {
          'header', () => {
         return supertest(app)
           .get('/login/authenticate')
-          .expect(res => {
+          .then((res) => {
             const parsedLocation = url.parse(res.header['location'], true);
             expect(parsedLocation.query).toExcludeKey('openid.ns.macaroon');
             expect(parsedLocation.query)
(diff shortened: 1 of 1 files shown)


bartaz:
[Tests randomly failing in dev environment](https://trello.com/c/E2R3xwVI/247-tests-randomly-failing-in-dev-environment)

test/routes/src/server/routes/t_login.js
```diff
@@ -19,49 +19,50 @@ describe('login routes', () => {
...
-    nock.cleanAll();
-  });
+    afterEach(() => {
+      nock.cleanAll();
+    });
 
-  describe('authenticate', () => {
     context('with no options', () => {
       it('should redirect from /login/authenticate to SSO', (done) => {
         supertest(app)
```
earnubs: From a mocha pov this might be neater with Promises and result in one less done(), but looks like it means adding `supertest-with-promises`, which is more  a refactoring task than a fixing task.

test/routes/src/server/routes/t_login.js
```diff
@@ -64,8 +64,8 @@ describe('login routes', () => {
     });
 
     context('with no options', () => {
-      it('should redirect from /login/authenticate to SSO', (done) => {
-        supertest(app)
+      it('should redirect from /login/authenticate to SSO', () => {
+        return supertest(app)
           .get('/login/authenticate')
           .expect(res => {
             expect(res.statusCode).toEqual(302);
```
earnubs: Can this be:

```
.expect(302)
.expect(res => { ... })
```

PR #415, opened 2017-03-06
Fixes for passing repository to build details page

### LINKED ISSUE #373: Navigating to another build page fails to show details
Steps to reproduce:

- Reload the dashboard
- Click on any latest build
- Click on the dashboard
- Click on any other latest build

What happens:

- The build details are not shown
- The build log is shown
- The install instructions are shown

What should happen:

- The build details are shown
- The build log is shown

bartaz:
Fixes #373

The issue was caused by `withRepository` container that was passing repository data from the store to build details page even if the repository in the store was different from the one in the URL.

With this PR we make sure to properly set repository in the store when it doesn't match the one in the URL. Also if these repos don't match child component is not rendered (until value in the store is properly propagated).

### DIFF
--- a/src/common/containers/with-repository.js
+++ b/src/common/containers/with-repository.js
@@ -8,7 +8,8 @@ function withRepository(WrappedComponent) {
   class WithRepository extends Component {
 
     componentDidMount() {
-      if (!this.props.repository && this.props.fullName) {
+      if ((!this.props.repository && this.props.fullName) ||
+          (this.props.repository.fullName !== this.props.fullName)) {
         this.props.dispatch(setGitHubRepository(this.props.fullName));
       }
     }
@@ -22,7 +23,11 @@ function withRepository(WrappedComponent) {
     render() {
       const { fullName, ...passThroughProps } = this.props; // eslint-disable-line no-unused-vars
 
-      return (this.props.repository
+      // only render if current repo in the store matches repo in URL params
+      const { repository } = this.props;
+      const isReady = repository && repository.fullName === fullName;
+
+      return (isReady
         ? <WrappedComponent {...passThroughProps} />
         : null
       );
--- a/test/unit/src/common/containers/t_with-repository.js
+++ b/test/unit/src/common/containers/t_with-repository.js
@@ -25,26 +25,13 @@ describe('WithRepository', () => {
 
   let store;
 
+  // renderDummy helper shallow renders wrapped component and assigns these:
   let wrapper;  // WithRepository rendered wrapper
   let instance; // WithRepository wrapper instance
   let wrapped;  // wrapped Dummy component
 
-  const setupDummy = (repository = REPO, params = PARAMS) => {
-    store = mockStore({ repository });
-
-    const DummyWithRepo = withRepository(Dummy);
-
-    // shallow render container component
-    const dummyWithRepo = shallow(<DummyWithRepo store={store} params={params}/>);
-
-    // get child components/instances from rendered component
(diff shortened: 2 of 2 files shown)


test/unit/src/common/containers/t_with-repository.js
```diff
@@ -0,0 +1,110 @@
...
+
+    const DummyWithRepo = withRepository(Dummy);
+
+    // shallow render container component
+    const dummyWithRepo = shallow(<DummyWithRepo store={store} params={params}/>);
+
+    // get child components/instances from rendered component
+    wrapper = dummyWithRepo.first();
+    instance = wrapper.shallow().instance();
+    wrapped = wrapper.dive();
```
bartaz: Higher order components make components nested, so to properly test them with shallow rendering we need to dive into the children.

src/common/containers/with-repository.js
```diff
@@ -22,7 +23,11 @@ function withRepository(WrappedComponent) {
     render() {
       const { fullName, ...passThroughProps } = this.props; // eslint-disable-line no-unused-vars
 
-      return (this.props.repository
+      // only render if current repo in the store matches repo in URL params
+      const { repository } = this.props;
+      const isReady = repository && repository.fullName === fullName;
+
+      return (isReady
```
katiefenn: Won't this hide the `WrappedComponent` on occasions that the repository is being fetched? What if a child component one day wants to show a spinner while repo data is being loaded?

test/unit/src/common/containers/t_with-repository.js
```diff
@@ -0,0 +1,110 @@
...
+    fullName: 'anowner/aname'
+  };
+
+  let store;
+
+  let wrapper;  // WithRepository rendered wrapper
+  let instance; // WithRepository wrapper instance
+  let wrapped;  // wrapped Dummy component
+
+  const setupDummy = (repository = REPO, params = PARAMS) => {
```
katiefenn: Extracting helper functions to the bottom of the file might make the body of the unit test files a bit easier to read.

src/common/containers/with-repository.js
bartaz: This data is not being fetched. It's built from URL params and set to the store.

So the owner/name pair from the URL is turned in the reducer into full repository object (with owner, name, fullName, url). 

This way WrappedComponent can rely on full repository object being available (and not parse it from URL itself).

test/unit/src/common/containers/t_with-repository.js
katiefenn: Can this helper function be reworked to return `wrapper` instead of writing it to a global? It would make it possible to move the function to the bottom of the file, making the body of the unit test easier to read.

src/common/containers/with-repository.js
katiefenn: Cool, that makes sense now 👍

test/unit/src/common/containers/t_with-repository.js
bartaz: Possibly it could, but what it actually prepares is `wrapper`, `instance` and `wrapped` objects that are used in tests.

If it only returns `wrapper`, each time we call setupDummy we would need to recreate those manually.

If that would help this helper function can be moved to the end of `describe`, so it will be kept in the scope of the variables it uses, but will not take space at the top of tests file.

katiefenn: I think the extra lines to manually set `instance` and `wrapped` would be worth it, but moving it to the bottom of the closure would also work.

PR #417, opened 2017-03-06
Convert server-side handlers to async/await

cjwatson:
`internalGetSnapcraftYaml` and anything that uses memcached particularly
benefit, but in general most things end up being more straightforward
and less deeply-indented.

### DIFF
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -282,11 +282,7 @@ export const newSnap = async (req, res) => {
     const urlPrefix = getRepoUrlPrefix(owner);
     const cacheId = getUrlPrefixCacheId(urlPrefix);
 
-    try {
-      await getMemcached().del(cacheId);
-    } catch (error) {
-      logger.error(`Error deleting ${cacheId} from memcached: ${error}`);
-    }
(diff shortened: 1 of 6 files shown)


src/server/handlers/launchpad.js
```diff
@@ -513,85 +490,85 @@ export const getSnapBuilds = (req, res) => {
...
+    );
+    const snap = await internalFindSnap(req.body.repository_url);
+    await snap.lp_delete();
+
+    const urlPrefix = getRepoUrlPrefix(owner);
+    const prefixCacheId = getUrlPrefixCacheId(urlPrefix);
+    const repoCacheId = getRepositoryUrlCacheId(req.body.repository_url);
+    try {
+      await getMemcached().del(prefixCacheId);
+    } catch (error) {
```
bartaz: In a simple cases like these, can we actually keep the `.catch()` notation from promise (which should be allowed on await calls)?

```js
await getMemcached().del(prefixCacheId)
  .catch((error) => logger.error( ... ));
```

This will allow to avoid repeated and nested `try ... catch ...`.

I'm not sure if I've seen any other case apart from `getMemcached().del()` where it was so visible, but maybe there are some.

PR #419, opened 2017-03-06
Add a /metrics endpoint

cjwatson:
For now, this just exposes the basic set of metrics provided by
`express-prom-bundle`.  In future we'll add our own as well.

We now pay attention to the `trusted_networks` charm configuration
option provided by `layer:ols`.  On staging and production, that will be
used to restrict access to the `/metrics` endpoint.

Part of #360.


src/server/server.js
```diff
@@ -46,8 +48,11 @@ app.use(expressWinston.logger({
 app.use(helmet());
 app.use(session(sessionConfig(conf)));
 app.use(Express.static(__dirname + '/../public', { maxAge: '365d' }));
+const metricsBundle = promBundle({ autoregister: false });
+app.use(metricsBundle);
 
 // routes
+app.use('/metrics', trustedNetworks, metricsBundle.metricsMiddleware);
```
bartaz: Does this mean we only allow metrics to be called from internal network, not from anyone?

cjwatson: That's the idea!  It's not very important with the minimal set available at the moment, but likely will be eventually.

PR #428, opened 2017-03-07
Link only to completed builds

### LINKED ISSUE #416: Don't take user to build page until log is available
When the build has started but hasn't yet finished, there will be no log to view. We shouldn't send the user to the build page (where all they can see is an empty log). Instead we should just tell them that the build is in progress on the dashboard page.

bartaz:
Fixes #416 

Because we currently only show log of completed builds, we want to 'hide' build details pages until log is available.

This mean links from dashboard or repo details to build pages should only be rendered when build log is already available.

### DIFF
--- a/src/common/components/repository-row/index.js
+++ b/src/common/components/repository-row/index.js
@@ -197,7 +197,14 @@ class RepositoryRow extends Component {
       registerNameStatus.snapName : snap.store_name
     );
 
-    const hasBuilt = latestBuild && snap.snapcraft_data;
+    const hasBuilt = !!(latestBuild && snap.snapcraft_data);
+    const hasLog = !!(hasBuilt && latestBuild.buildLogUrl);
+
+    // only link to builds that have log available
+    const latestBuildUrl = hasLog
+      ? `/${fullName}/builds/${latestBuild.buildId}`
+      : null;
+
     const isActive = (
       showNameMismatchDropdown ||
       showUnconfiguredDropdown ||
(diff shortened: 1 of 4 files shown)


src/common/components/repository-row/index.js
```diff
@@ -354,31 +357,29 @@ class RepositoryRow extends Component {
...
     git_repository_url: PropTypes.string,
-    self_link: PropTypes.string,
     store_name: PropTypes.string,
     snapcraft_data: PropTypes.object
   }),
   latestBuild: PropTypes.shape({
     buildId: PropTypes.string,
-    status: PropTypes.string,
+    buildLogUrl: PropTypes.string,
+    colour: PropTypes.string,
```
bartaz: Cleaning up in `PropTypes`. I think only properties that are used by given component should be listed here.

src/common/components/repository-row/index.js
```diff
@@ -354,31 +357,29 @@ class RepositoryRow extends Component {
...
-    error: PropTypes.object
+    success: PropTypes.bool
   }),
   dispatch: PropTypes.func.isRequired
 };
 
+// FIXME:
+// `connect` shouldn't be used just to add dispatch to props
+// this makes it much harder to test it as it get wrapped
 export default connect()(RepositoryRow);
```
bartaz: I think that's an antipattern we should avoid.

`RepositoryRow` gets all it's data in props from parent component, so in theory could be a pure presentational component. This `connect` here doesn't really bind it to data, it's only used to pass `dispatch`, so it can call various redux actions (which it does quite a lot, as there is lot's of functionality there).

I don't want to change it now (that's why I just added a FIXME), but I think it should be either pure react component that will not get dispatch from connect (so we either pass dispatch or different action handlers from parent component) or we make it into fully connected component that reads its data from the store.

But that's probably something that can be part of component structure discussion we have in plans.

PR #430, opened 2017-03-07
Add a bsi_github_users_total metric

cjwatson:
This shows the number of GitHub users who have ever logged in (since the
creation of the database table, at least).

Part of #360.

### DIFF
--- a/src/server/metrics/github-users.js
+++ b/src/server/metrics/github-users.js
@@ -1,0 +1,12 @@
+import promClient from 'prom-client';
+
+import { GitHubUser } from '../db/models/github-user';
+
+const gitHubUsersTotal = new promClient.Gauge(
+  'bsi_github_users_total',
+  'Total number of GitHub users who have ever logged in.'
+);
+
+export default async function updateGitHubUsersTotal() {
+  gitHubUsersTotal.set(await GitHubUser.count());
+}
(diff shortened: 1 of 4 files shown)


src/server/metrics/index.js
```diff
@@ -0,0 +1,17 @@
+import updateGitHubUsersTotal from './github-users';
+
+let existingInterval = null;
+
+export default async function startAllMetrics() {
+  if (existingInterval !== null) {
+    clearInterval(existingInterval);
+  }
+
+  async function updateAllMetrics() {
```
bartaz: Nit picking a bit, but can't `updateAllMetrics` be defined outside of `startAllMetrics` scope? It doesn't seem to rely on any internal scope, so it doesn't need to be redefined every time `startAllMetrics` is called.

Of course, we call `startAllMetrics` just once, so it shouldn't really be an issue.

src/server/metrics/github-users.js
```diff
@@ -0,0 +1,12 @@
+import promClient from 'prom-client';
+
+import { GitHubUser } from '../db/models/github-user';
+
+const gitHubUsersTotal = new promClient.Gauge(
```
bartaz: Not sure if it is worth the effort (because I guess promClient would need to be mocked), but maybe it would be worth having at least simple unit test that verifies that we expose the metrics?

PR #434, opened 2017-03-08
Convert auth-store action tests to async/await

cjwatson:


test/unit/src/common/actions/t_auth-store.js
```diff
@@ -161,16 +161,14 @@ describe('store authentication actions', () => {
...
 
-        return store.dispatch(getSSODischarge())
-          .then(() => {
-            api.done();
-            const action = store.getActions().filter(
-              (a) => a.type === ActionTypes.GET_SSO_DISCHARGE_ERROR)[0];
-            expect(action.payload.message).toBe(expectedMessage);
-          });
+        await store.dispatch(getSSODischarge());
+        api.done();
```
bartaz: `api.done()` calls can be moved to `afterEach` hook, so it will make sure all tests in given context will satisfy mocks without repeating it in each of them.

Not really part of this conversion, but would be nice touch.

PR #435, opened 2017-03-08
Added snapcraft.yaml path do snapcraft_data

### LINKED ISSUE #297: [name not matching] get location of snapcraft.yaml name
Handling name mismatch requires a link to snapcraft.yaml file in GitHub.

Currently we only know that file exists, but it may be in one of 3 allowed locations (/snap/snapcraft.yaml, snapcraft.yaml or .snapcraft.yaml), so we need to know where the file is in current repo.

We already have snapcraft_data with a name in it, it should probably be extended to include location of the file.

---

Copied from Trello card: https://trello.com/c/aTaRuaMq/325-name-not-matching-get-location-of-snapcraft-yaml-name
Labels:

bartaz:
Fixes #297

It's not a UI change, but something that is needed to allow us to provide a link to edit snapcraft.yaml file in repo.

Snapcraft.yaml can currently be in one of 3 possible locations, so we not only need to know that file exists, but also where it is in repository.

`snapcraft_data` that we pass together with snap (enabled repository) currently contains just a snap `name` from snapcraft.yaml. While it was meant to be a representation of contents of snapcraft.yaml it seems the only reasonable place to add the file path.

### DIFF
--- a/src/server/handlers/github.js
+++ b/src/server/handlers/github.js
@@ -101,12 +101,19 @@ export const getSnapcraftData = async (repositoryUrl, token) => {
   try {
     const snapcraftYaml = await internalGetSnapcraftYaml(owner, name, token);
     const snapcraftData = {};
-    for (const index of Object.keys(snapcraftYaml)) {
+    for (const index of Object.keys(snapcraftYaml.contents)) {
       if (SNAPCRAFT_INFO_WHITELIST.indexOf(index) >= 0) {
-        snapcraftData[index] = snapcraftYaml[index];
+        snapcraftData[index] = snapcraftYaml.contents[index];
       }
     }
 
+    // copy snapcraft.yaml path from repo into snapcraftData
+    //
+    // XXX we are mixing our custom `path` into data from snapcraft.yaml file
+    // currently there is no `path` defined in snapcraft syntax
+    // https://snapcraft.io/docs/build-snaps/syntax
+    // and also we whitelist only `name`, so no collision should occur
+    snapcraftData.path = snapcraftYaml.path;
     await getMemcached().set(cacheId, snapcraftData, 3600);
     return snapcraftData;
   } catch (error) {
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -211,7 +211,10 @@ export const internalGetSnapcraftYaml = async (owner, name, token) => {
     try {
       const response = await fetchSnapcraftYaml(path, owner, name, token);
       try {
-        return yaml.safeLoad(response.body);
+        return {
+          contents: yaml.safeLoad(response.body),
+          path
+        };
       } catch (e) {
         throw new PreparedError(400, RESPONSE_SNAPCRAFT_YAML_PARSE_FAILED);
       }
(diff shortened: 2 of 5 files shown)


test/routes/src/server/routes/t_github.js
```diff
@@ -34,24 +40,26 @@ describe('The GitHub API endpoint', () => {
         nock.cleanAll();
       });
 
-      it('should successfully return', (done) => {
-        supertest(app)
-          .get('/github/snapcraft-yaml/anowner/aname')
-          .expect(200, done);
+      it('should satisfy all request mocks', async () => {
```
earnubs: I wonder if moving this into it's own test isn't detrimental to debugging failed tests? Knowing with confidence that the mock was satisfied in a test is a good thing, and even though these tests hit the same endpoint without messing around perhaps this isn't a good pattern to establish? Maybe move to afterEach?

bartaz: Good point. I didn't want to repeat them in all tests, and separate test case sounded bit artificial.

Moving to `afterEach` sounds as a good compromise. But in the end it will probably also not really tell which test case made it fail.

bartaz: I was wrong, it does tell:

`The Launchpad API endpoint list snaps route when LP returns error "after each" hook for "should satisfy all request mocks"`

So to `afterEach` it goes...

src/server/handlers/github.js
```diff
@@ -101,12 +101,19 @@ export const getSnapcraftData = async (repositoryUrl, token) => {
...
       }
     }
 
+    // copy snapcraft.yaml path from repo into snapcraftData
+    //
+    // XXX we are mixing our custom `path` into data from snapcraft.yaml file
+    // currently there is no `path` defined in snapcraft syntax
+    // https://snapcraft.io/docs/build-snaps/syntax
+    // and also we whitelist only `name`, so no collision should occur
+    snapcraftData.path = snapcraftYaml.path;
```
bartaz: This is a bit questionable part.

`snapcraftData` was meant to keep (whitelisted) contents of snapcraft.yaml (only name ATM). But it's the only reasonable place to put path info.
Right now I just put it in `path` field, because it doesn't collide with anything from real snapcraft.yaml. But in theory, in future it may (if snapcraft.yaml adds some path property and we would whitelist it).

So the question is, do we keep it under `path`, as it's convenient and in the end we are controlling exactly what part of snapcraft.yaml goes into `snapcraftData` or so we want to protect ourselves from possible future by "obfuscating" the prop name as something like `snapcraft_yaml_path` or similar?

src/server/handlers/launchpad.js
```diff
@@ -211,7 +211,10 @@ export const internalGetSnapcraftYaml = async (owner, name, token) => {
     try {
       const response = await fetchSnapcraftYaml(path, owner, name, token);
       try {
-        return yaml.safeLoad(response.body);
+        return {
+          contents: yaml.safeLoad(response.body),
+          path
```
bartaz: Return both contents of snapcraft.yaml and its path internally.

src/server/handlers/github.js
bartaz: The other option would be to separate contents of snapcraft.yaml and the path and propagate this change all down to the client app, but because we only read the name it would make it unnecessarily nested as `snapcraft_data.contents.name` when accessing.

Issue #293, opened 2017-03-03
[dashboard] Arrival or removal of snapcraft.yaml in repo isn’t auto-detected


nottrobin:
A1\. Add a repo that doesn’t have a snapcraft.yaml.
A2\. Add the snapcraft.yaml.

What happens:

A1\. The “Configured” cell says “Not configured”.
A2\. The “Configured” cell still says “Not configured” until you reload the page.

What should happen:

A2\. The “Configured” cell changes to a checkmark without you having to reload the page.
___

B1\. Add a repo that does have a snapcraft.yaml.
B2\. Remove the snapcraft.yaml.

What happens:

B1\. The “Configured” cell has a checkmark..
B2\. The “Configured” cell still has a checkmarkuntil you reload the page.

What should happen:

B2\. The “Configured” cell changes to “Not configured” without you having to reload the page.

---

Copied from Trello card: https://trello.com/c/ZhOfeleQ/256-dashboard-arrival-of-snapcraft-yaml-in-repo-isn-t-auto-detected
Labels: Bug

PR #436, opened 2017-03-08
#293: Auto-update of configured status

katiefenn:
Fixes: #293 
- Added polled update of configured status to dashboard


src/common/actions/snaps.js
```diff
@@ -12,14 +12,16 @@ export const REMOVE_SNAP = 'REMOVE_SNAP';
 export const REMOVE_SNAP_SUCCESS = 'REMOVE_SNAP_SUCCESS';
 export const REMOVE_SNAP_ERROR = 'REMOVE_SNAP_ERROR';
 
-export function fetchUserSnaps(owner) {
+export function fetchUserSnaps(owner, showFetching = true) {
```
katiefenn: New param stops fetching status (i.e. the spinner) from being shown while fetching new snap data. This is useful when polling the service for updates.

src/common/components/repositories-home/index.js
```diff
@@ -33,12 +36,22 @@ class RepositoriesHome extends Component {
...
     const owner = this.props.user.login;
 
     if (authenticated) {
-      this.props.dispatch(fetchUserSnaps(owner));
+      dispatch(fetchUserSnaps(owner));
+      if (!interval) {
+        interval = setInterval(() => {
+          dispatch(fetchUserSnaps(owner, false));
+        }, SNAP_POLL_PERIOD);
+      }
```
katiefenn: Poll `fetchUserSnaps` with a fifteen second interval, only if an existing timer has not been set up.

src/common/components/repositories-home/index.js
```diff
@@ -33,12 +36,22 @@ class RepositoriesHome extends Component {
...
     }
+  }
 
+  componentWillUnmount() {
+    clearInterval(interval);
+    interval = null;
```
katiefenn: Clear timer, preventing snap data being fetched when user navigates away from the page.

src/common/actions/snaps.js
earnubs: How do you check if you're fetching?

bartaz: Not sure if I agree with such approach. This makes the state in store little bit inconsistent, because if we call it with `showFetching = false` data is fetching (we have request in progress), but the store doesn't reflect this change.

If we don't want to show spinner during polling (and I agree with that) we should make UI make this decision based on state in the store - not make the store pretend there is no request in progress.

In build pages (where we poll for builds statuses) it's done as `showLoading = isFetching && !success` - but this assumes that action that inits fetching doesn't clear the success status in store. This is not the case in `snaps` reducer as far as I can see. And I guess it's debatable what should be our convention in such case (should we clear success when we set isFetching)?

Other option would be do do `showLoading = isFetching && !snaps`. So we will only show loading if we don't have snaps data already.

So, to summarize my point here: I think the state should always have isFetching set when we wait for response (for example so we can detect it and not fetch again until completed), and it's UI component that should decide about showing loading based on what is in the store. I don't think we should create 'silent' action that fetch requests, but hide some info from the store.

earnubs: Another option is that the UI always shows the isFetching state but in a less obtrusive way, for example if the spinner is behind the content (only visible when we have no content). Cheating, I know.

src/common/components/repositories-home/index.js
```diff
@@ -89,10 +89,11 @@ class RepositoriesHome extends Component {
...
   render() {
     const { snaps } = this.props;
+    const hasSnaps = (snaps.snaps && Object.keys(snaps.snaps).length > 0);
     // show spinner until we know if user has any enabled repos
-    return (snaps.success || snaps.error)
-        ? this.renderRepositoriesList()
-        : this.renderSpinner();
+    return (snaps.isFetching && !hasSnaps)
+      ? this.renderSpinner()
+      : this.renderRepositoriesList();
```
katiefenn: I've changed the criteria for showing the repositories list to avoid explicitly overriding the "isFetching" status.

`snaps.success` is not a reliable criteria, because it changes whenever we poll the service for updated snap information. Instead, I've used the `isFetching` status coupled with the existence of snaps.

test/unit/src/common/components/repositories-home/t_repositories-home.js
```diff
@@ -0,0 +1,85 @@
...
+import expect from 'expect';
+import { shallow } from 'enzyme';
+import proxyquire from 'proxyquire';
+import sinon from 'sinon';
+
+const RepositoriesHome = proxyquire(
+  '../../../../../../src/common/components/repositories-home',
+  {
+    '../../actions/snaps': {
+      fetchUserSnaps: sinon.stub().returns({ type: 'FETCH_SNAP_REPOSITORIES' })
```
bartaz: I guess that's where `mapDispatchToProps` would help as suggested by @earnubs.

We would be able to pass mocked dispatch calls to raw component via props without ugly proxyquire calls.

src/common/components/repositories-home/index.js
```diff
@@ -109,3 +123,4 @@ function mapStateToProps(state) {
 }
 
 export default connect(mapStateToProps)(withRouter(RepositoriesHome));
+export const raw = RepositoriesHome;
```
bartaz: I know we didn't settle down on any convention for this, but is basically the same as exporting `RepositoriesHome` directly - maybe just under different name, like `RepositoriesHomeView`.

But `RepositoriesHomeRaw` works for me as well. Simply - I would be for exporting it under full uppercase class name. It doesn't matter much in this case with proxyquire, but it does with import.

src/common/components/repositories-list/index.js
```diff
@@ -100,15 +99,11 @@ class RepositoriesList extends Component {
...
 
   render() {
-    const isLoading = this.props.snaps.isFetching;
+    const { snaps } = this.props.snaps;
+    const hasSnaps = (snaps && Object.keys(snaps).length > 0);
 
     return (
       <div className={styles.repositoriesList}>
-        { isLoading &&
-          <div className={ spinnerStyles }>
```
bartaz: I guess this throws on linter, as `spinnerStyles` are still imported but not used.

test/unit/src/common/components/repositories-home/t_repositories-home.js
katiefenn: Yep, that would be good.

src/common/components/repositories-list/index.js
katiefenn: This should now be resolved.

src/common/components/repositories-home/index.js
earnubs: Why not an explicit poll action, with associated state?

katiefenn: @earnubs: It was something I considered, but as @bartaz suggested we already have state in the store that can tell us when we're polling and when we're fetching for the initial status. In fact, we could probably write a selector to do this for us.

Issue #363, opened 2017-03-03
[Dashboard] Repo table row doesn’t auto-expand when no repos have a name yet


nottrobin:
Panel for registering snap name for the first repo opens automatically when user has at least one repo, but none have a name yet (and no other repos have ever been removed).

---

Copied from Trello card: https://trello.com/c/HAFdLZOo/234-dashboard-repo-table-row-doesn-t-auto-expand-when-no-repos-have-a-name-yet
Labels:

earnubs:
Fixed in #450 & #484

PR #450, opened 2017-03-09
computing derived data: use reselect to memoize 'hasNamedSnaps' selector

earnubs:
See http://redux.js.org/docs/recipes/ComputingDerivedData.html
and https://github.com/reactjs/reselect

Then use that to determine if the first row in reposlist should be opened on register name.

### DIFF
--- a/src/common/components/repositories-list/index.js
+++ b/src/common/components/repositories-list/index.js
@@ -1,6 +1,7 @@
 import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
 
+import { hasNamedSnaps } from '../../selectors'
 import { conf } from '../../helpers/config';
 import {
   checkSignedIntoStore,
@@ -69,17 +70,18 @@ class RepositoriesList extends Component {
     }
   }
 
-  renderRow(snap) {
+  renderRow(snap, index) {
+    const { hasNamedSnaps, registerName, snapBuilds, authStore } = this.props;
     const { fullName } = parseGitHubRepoUrl(snap.git_repository_url);
 
     let latestBuild = null;
-    const snapBuilds = this.props.snapBuilds[fullName];
+    const snapBuild = snapBuilds[fullName];
 
-    if (snapBuilds && snapBuilds.success && snapBuilds.builds.length) {
+    if (snapBuild && snapBuild.success && snapBuild.builds.length) {
       latestBuild = snapBuilds.builds[0];
     }
 
-    const registerNameStatus = this.props.registerName[fullName] || {};
+    const registerNameStatus = registerName[fullName] || {};
 
     return (
       <RepositoryRow
@@ -87,8 +89,9 @@ class RepositoriesList extends Component {
         snap={ snap }
         latestBuild={ latestBuild }
         fullName={ fullName }
-        authStore={ this.props.authStore }
+        authStore={ authStore }
         registerNameStatus={ registerNameStatus }
+        registerNameIsOpen={ !index && !hasNamedSnaps }
       />
     );
   }
@@ -152,7 +155,8 @@ function mapStateToProps(state) {
     authStore,
     registerName,
     snaps,
-    snapBuilds
+    snapBuilds,
+    hasNamedSnaps: hasNamedSnaps(state)
   };
 }
 
--- a/src/common/components/repositories-list/index.js
+++ b/src/common/components/repositories-list/index.js
@@ -1,7 +1,7 @@
 import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
 
-import { hasNamedSnaps } from '../../selectors'
+import { hasNoRegisteredNames } from '../../selectors';
 import { conf } from '../../helpers/config';
 import {
   checkSignedIntoStore,
@@ -71,7 +71,7 @@ class RepositoriesList extends Component {
   }
 
   renderRow(snap, index) {
-    const { hasNamedSnaps, registerName, snapBuilds, authStore } = this.props;
+    const { hasNoRegisteredNames, registerName, snapBuilds, authStore } = this.props;
     const { fullName } = parseGitHubRepoUrl(snap.git_repository_url);
 
     let latestBuild = null;
@@ -91,7 +91,7 @@ class RepositoriesList extends Component {
         fullName={ fullName }
         authStore={ authStore }
         registerNameStatus={ registerNameStatus }
-        registerNameIsOpen={ !index && !hasNamedSnaps }
+        registerNameIsOpen={ !index && hasNoRegisteredNames }
       />
     );
   }
@@ -138,7 +138,8 @@ RepositoriesList.propTypes = {
   registerName: PropTypes.object,
   snaps: PropTypes.object,
   snapBuilds: PropTypes.object,
-  dispatch: PropTypes.func.isRequired
+  dispatch: PropTypes.func.isRequired,
+  hasNoRegisteredNames: PropTypes.bool
 };
 
 function mapStateToProps(state) {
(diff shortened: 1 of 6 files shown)


bartaz:
I like the idea of selectors and I think it's one of missing part in our app that would make it easier for us to refactor the store and access data. Selectors would allow us to be more independent of the structure of the store and have a common 'API' for accessing data in it.

src/common/selectors/index.js
```diff
@@ -0,0 +1,20 @@
...
+const getSnaps = state => state.snaps;
+const getNames = state => state.registerName;
+
+export const hasNamedSnaps = createSelector(
+  [getSnaps, getNames],
+  (snaps, names) => {
+    return snaps.snaps.some((snap) => {
+      if (typeof(snap.store_name) === 'string') return true;
+
+      const { fullName } = parseGitHubRepoUrl(snap.git_repository_url);
```
bartaz: I think this is (or at least should be) unnecessary.
Snaps reducer already listens to register action and updates `store_name` directly in the store, so we don't even need to re-fetch snaps from LP to get this change: https://github.com/canonical-ols/build.snapcraft.io/blob/master/src/common/reducers/snaps.js#L61-L65

So if the name was registered by user in BSI, it should immediately be updated in the stored snap.

src/common/components/repositories-list/index.js
```diff
@@ -69,26 +70,28 @@ class RepositoriesList extends Component {
...
   }
 
-  renderRow(snap) {
+  renderRow(snap, index) {
+    const { hasNamedSnaps, registerName, snapBuilds, authStore } = this.props;
     const { fullName } = parseGitHubRepoUrl(snap.git_repository_url);
 
     let latestBuild = null;
-    const snapBuilds = this.props.snapBuilds[fullName];
+    const snapBuild = snapBuilds[fullName];
```
bartaz: Name is confusing a bit, as this is a list of all builds for given snap.
But I get the confusion with `snapBuilds` from store (which is all builds for all snaps).

Maybe `currentSnapBuilds`?

earnubs:
@bartaz I think some of what I wanted to refactor in the store is actually just

src/common/components/repositories-list/index.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
 
+import { hasNamedSnaps } from '../../selectors'
```
earnubs: Actually, this should probably be in the inverse?

src/common/selectors/index.js
```diff
@@ -0,0 +1,15 @@
+import { createSelector } from 'reselect'
+
+import { parseGitHubRepoUrl } from '../helpers/github-url.js';
```
earnubs: delete

src/common/selectors/index.js
```diff
@@ -0,0 +1,12 @@
+import { createSelector } from 'reselect';
+
+const getSnaps = state => state.snaps;
+
+export const hasNoRegisteredNames = createSelector(
+  [getSnaps],
+  (snaps) => {
+    return !snaps.snaps.some((snap) => {
+      return snap.store_name;
```
earnubs: snaps snaps some snap snap store ...

src/common/components/repositories-list/index.js
```diff
@@ -69,26 +70,28 @@ class RepositoriesList extends Component {
...
     return (
       <RepositoryRow
         key={ `repo_${fullName}` }
         snap={ snap }
         latestBuild={ latestBuild }
         fullName={ fullName }
-        authStore={ this.props.authStore }
+        authStore={ authStore }
         registerNameStatus={ registerNameStatus }
+        registerNameIsOpen={ !index && hasNoRegisteredNames }
```
bartaz: As far as I understand this will work only for first repo from the list (index = 0), right?

My understanding of requirements was that it should open first repo that doesn't have name registered. Not first repo if it doesn't have a name registered.

So if first has name registered, but second doesn't, it should open the second one. (as it will be the first on the list that doesn't have it registered)...

But maybe it's overinterpratation...

src/common/components/repositories-list/index.js
```diff
@@ -69,26 +70,28 @@ class RepositoriesList extends Component {
...
       />
     );
   }
```
bartaz: I know it's not part of this PR, but the `onButtonClick` method just below that doesn't seem to be used anywhere and looks like it was just copied from somewhere else.
Also `createSnap` action dispatched from it shouldn't be needed at all in this component.

test/unit/src/common/selectors/t_selectors.js
```diff
@@ -0,0 +1,37 @@
...
+  it('should be false when no names in store', function() {
+    expect(hasNoRegisteredNames(stateWithNoName)).toBe(true);
+  });
+
+  it('should be true when name in snaps', function() {
+    expect(hasNoRegisteredNames(stateWithName)).toBe(false);
+  });
+
+  it('should not memoize the selector', function() {
+    expect(hasNoRegisteredNames.recomputations()).toBe(2);
```
bartaz: Not sure if I understand this test.
What it's checking, why expected is 2?

nottrobin:
This fixes #363

src/common/components/repositories-list/index.js
earnubs: The spec has "Panel for registering snap name for the first repo opens " which could be interpreted as both I guess, but seems to me to mean the first repo in the list.

@matthewpaulthomas can you clarify?

test/unit/src/common/selectors/t_selectors.js
earnubs: See https://github.com/reactjs/reselect#q-how-do-i-test-a-selector

This is the number of times it recomputed the selector, and it should have done it twice because we changed state twice, next test doesn't change state so we get the memoized value.

src/common/components/repositories-list/index.js
earnubs: The complete spec is "At least one repo, but none have a name yet ... Panel for registering snap name for the first repo opens"

Which can only be the first repo in the list because none of the repos have a name, right?

katiefenn: > Which can only be the first repo in the list because none of the repos have a name, right?

This was my reading of it. I think we should make the test slightly more specific, though:
`registerNameIsOpen={ index === 0 && hasNoRegisteredNames }`

`index === 0` could also be extracted into a variable named `isFirst` or `isFirstInList`.

test/unit/src/common/selectors/t_selectors.js
bartaz: Does this mean that result of this test depends on the fact that 2 previous tests are run? Are these dependent? The state is not cleared between the tests?

earnubs: Yes, but that's intended. I'll wrap it up into it's own describe block and use `resetRecomputations` to make that clear.

test/unit/src/common/selectors/t_selectors.js
```diff
@@ -0,0 +1,53 @@
...
+    snaps: {
+      snaps: [{
+        store_name: 'bsi-test-ii'
+      }, {}]
+    }
+  };
+
+  // XXX must run first in suite to avoid memoizing a selector and being
+  // off-by-one in recomputations()
+  context('is a memoized selector', function() {
```
earnubs: I've tried to make this clearer, but the computations count is dependent on tests which ran before even if you reset the count, because the selector may be memoized, and if you give it the same state again, the count won't increment.

earnubs: NB. we don't actually need these tests, they're covering what is already covered in reselect unit tests

PR #452, opened 2017-03-09
Change registered name UI

### LINKED ISSUE #301: [name not matching] change registered name
Handle "If none of the above apply:" case from the spec.

This includes linking to "change registered name" flow.

---

Copied from Trello card: https://trello.com/c/NNK4WJ56/329-name-not-matching-change-registered-name
Labels:

bartaz:
Fix #301 

TODO:
- [x] clarify design questions from spec doc
- [x] update styles
- [x] ~~clean up dropdown code~~

### DIFF
--- a/src/common/components/repository-row/dropdowns/dropdowns.css
+++ b/src/common/components/repository-row/dropdowns/dropdowns.css
@@ -17,3 +17,13 @@
   background-size: 0.7em;
   padding-right: 0.9em;
 }
+
+.caption ul {
+  list-style: disc;
+  margin-left: 2em;
+  margin-bottom: 1em;
+}
+
+.snapNameInput {
+  width: 250px;
+}
--- a/src/common/components/repository-row/dropdowns/register-name-dropdown.js
+++ b/src/common/components/repository-row/dropdowns/register-name-dropdown.js
@@ -57,6 +57,12 @@ const Caption = (props) => {
   let changeForm;
   let message;
 
+  const helpText = (
+    <div className={ styles.helpText }>
+      Lower-case letters, numbers, and hyphens only.
+    </div>
+  );
+
   const changeRegisteredName = !!registeredName;
 
   // only show "change name" info and form if there is a name registered already,
@@ -67,16 +73,18 @@ const Caption = (props) => {
     !registerNameStatus.success
   );
 
-  if (!changeRegisteredName) {
+  const showHelpText = !showChangeForm && (authStoreFetchingDischarge || authStore.authenticated);
+
+  if (changeRegisteredName) {
+    // if user is changing already registered name
+    if (!authStore.authenticated) {
+      message = 'You need to sign in to Ubuntu One to change a snap’s registered name.';
+    }
+  } else {
     // if user is registering new name
     message = 'To publish to the snap store, this repo needs a registered name.';
     if (!authStore.authenticated) {
       message += ' You need to sign in to Ubuntu One to register a name.';
-    }
-  } else {
-    // if user is changing already registered name
-    if (!authStore.authenticated) {
-      message = 'You need to sign in to Ubuntu One to change a snap’s registered name.';
     }
   }
 
@@ -101,9 +109,7 @@ const Caption = (props) => {
             />
           </label>
         </p>
-        <div className={ styles.helpText }>
-          Lower-case letters, numbers, and hyphens only.
-        </div>
+        { helpText }
       </div>
     );
   }
@@ -135,11 +141,7 @@ const Caption = (props) => {
     caption = (
       <div>
         { message }
-        { !showChangeForm && (authStoreFetchingDischarge || authStore.authenticated) &&
-          <div className={ styles.helpText }>
-            Lower-case letters, numbers, and hyphens only.
-          </div>
-        }
+        { showHelpText && helpText }
         { userMustSignAgreement && <Agreement onChange={onSignAgreementChange}/> }
       </div>
     );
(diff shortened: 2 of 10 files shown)


src/common/components/repository-row/index.js
```diff
@@ -149,9 +149,10 @@ class RepositoryRow extends Component {
   componentWillReceiveProps(nextProps) {
     if (nextProps.registerNameStatus.success &&
         !this.props.registerNameStatus.success) {
-      this.closeUnregisteredTimerID = window.setTimeout(
-        this.closeUnregisteredDropdown.bind(this), 2000
-      );
+      this.closeUnregisteredTimerID = window.setTimeout(() => {
+        this.closeUnregisteredDropdown();
+        this.props.dispatch(registerNameClear(this.props.fullName));
```
bartaz: We need to clear state of register name API calls, because user may open up the dropdown again to change the name (and shouldn't see 'success' status there).

src/common/components/repository-row/dropdowns/register-name-dropdown.js
```diff
@@ -47,6 +54,60 @@ const Caption = (props) => {
...
+      message += ' You need to sign in to Ubuntu One to register a name.';
+    }
+  } else {
+    // if user is changing already registered name
+    if (!authStore.authenticated) {
+      message = 'You need to sign in to Ubuntu One to change a snap’s registered name.';
+    }
+  }
+
+  if (showChangeForm) {
```
bartaz: It's a bit messy now.

This dropdown now needs to support both name registering and changing name (which have different UIs) and signing to the SSO (with 2 different messages depending on the state).

This means a lot of conditionals here...

There is no much to do with this until we have more consistent UI for registering and changing name, which may come later.

src/common/components/repository-row/dropdowns/dropdowns.css
```diff
@@ -17,3 +17,13 @@
   background-size: 0.7em;
   padding-right: 0.9em;
 }
+
+.caption ul {
```
bartaz: I don't really like it this way, but we don't have any global list styles...

src/common/components/repository-row/dropdowns/register-name-dropdown.js
```diff
@@ -47,6 +54,60 @@ const Caption = (props) => {
...
+              className={ styles.snapNameInput }
+              type='text'
+              value={ snapName }
+              onChange={ onSnapNameChange }
+            />
+          </label>
+        </p>
+        <div className={ styles.helpText }>
+          Lower-case letters, numbers, and hyphens only.
+        </div>
```
cjwatson: Maybe pull this `div` out so that it can be common with the new registration case?

src/common/components/repository-row/dropdowns/register-name-dropdown.js
```diff
@@ -47,6 +54,60 @@ const Caption = (props) => {
...
+        </ul>
+        <p>
+          <label>New name:
+            {' ' /* force space between inline elements */}
+            <input
+              className={ styles.snapNameInput }
+              type='text'
+              value={ snapName }
+              onChange={ onSnapNameChange }
+            />
```
cjwatson: Is there any reasonable way to make this common with the similar `input` element in `src/common/components/repository-row/index.js`?

src/common/components/repository-row/dropdowns/register-name-dropdown.js
```diff
@@ -47,6 +54,60 @@ const Caption = (props) => {
...
+
+  // only show "change name" info and form if there is a name registered already,
+  // user is authenticated and we are not showing register success message
+  const showChangeForm = (
+    changeRegisteredName &&
+    authStore.authenticated &&
+    !registerNameStatus.success
+  );
+
+  if (!changeRegisteredName) {
```
cjwatson: I usually find this kind of conditional easier to read when the positive case comes first.

src/common/components/repository-row/dropdowns/register-name-dropdown.js
```diff
@@ -129,7 +194,7 @@ const ActionButtons = (props) => {
       snapName === '' ||
       !!registerNameStatus.error
     );
-    actionText = 'Register name';
+    actionText = !registeredName ? 'Register name' : 'Register new name';
```
cjwatson: As above, I usually find this kind of conditional easier to read when the positive case comes first.

src/common/components/repository-row/dropdowns/register-name-dropdown.js
```diff
@@ -58,28 +119,23 @@ const Caption = (props) => {
...
   } else {
     caption = (
       <div>
-        To publish to the snap store, this repo needs a registered name.
-        { !authStore.authenticated &&
-          ' You need to sign in to Ubuntu One to register a name.'
-        }
-        { (authStoreFetchingDischarge || authStore.authenticated) &&
+        { message }
+        { !showChangeForm && (authStoreFetchingDischarge || authStore.authenticated) &&
```
earnubs: Ouch.

bartaz: Just created one so we don't loose track of it #468

src/common/components/repository-row/dropdowns/register-name-dropdown.js
bartaz: At this point it would be hard.
Im not sure if we want to inject whole input from repository row to the dropdown (or the other way around).

With #468 coming at some point both UIs should be made consistent, so we will likely end up with input only in the dropdown, and not in table itself.

PR #456, opened 2017-03-10
Sanity-check pgsql.master before using it

cjwatson:
It can be None here when running from the `db-admin-relation-changed`
hook.


charm/reactive/snap-build.py
```diff
@@ -85,6 +85,9 @@ def migrate(pgsql):
     KNEXFILE_NORMAL: ['snap-build'],
     }, stopstart=True)
 def configure(pgsql, cache):
+    db_name = hookenv.config('db_name')
+    if pgsql.master is None or pgsql.master.dbname != db_name:
```
earnubs: Should/could we log something here?

PR #460, opened 2017-03-12
refine look of traffic lights

### LINKED ISSUE #354: [dashboard] Using colours in the progress meter
Grazina thought it was odd that the progress meter uses three colours (green, blue, and grey) rather than two.

This is especially since the first time you see the meter, you are already on step two. At that point there’s one step in each colour, making it less obvious that they’re progress colours (rather than, say, step colours).

- Also look at the amount of colour (more delicate)

---

Copied from Trello card: https://trello.com/c/MUQPy1BO/298-dashboard-using-colours-in-the-progress-meter
Labels: design

pmahnke:
## Done

* removed the blue colour
* made the dots smaller
* removed the 1,2,3 label … can be removed if we like this better
* other design tweaks

## QA

1. go to the initial repo enrollment and check if the traffic lights match the design in the related issue

## Issue/Card

Fixes #354 

## Screenshots

image

### DIFF
--- a/src/common/components/traffic-lights/index.js
+++ b/src/common/components/traffic-lights/index.js
@@ -18,14 +18,12 @@ const stateStyles = [
 export class Signal extends Component {
 
   render() {
-    const { state, label, message } = this.props;
+    const { state, message } = this.props;
     const signalStyle = stateStyles[state];
 
     return (
       <div className={ styles.box }>
-        <div className={ classNames(styles.signal, signalStyle) }>
-          { (state === SIGNALS.DONE) ? '' : label }
-        </div>
+        <div className={ classNames(styles.signal, signalStyle) }></div>
         <p>{ message }</p>
       </div>
     );
@@ -45,13 +43,10 @@ export default class TrafficLights extends Component {
 
     this.state = {
       signals: [{
-        label: '',
         message: 'Connect to Github'
       }, {
-        label: '',
         message: 'Choose a repo'
       }, {
-        label: '',
         message: 'Name and YAML'
       }]
     };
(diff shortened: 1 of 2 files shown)


src/common/components/traffic-lights/index.js
```diff
@@ -45,13 +45,13 @@ export default class TrafficLights extends Component {
 
     this.state = {
       signals: [{
-        label: '1',
+        label: '',
```
bartaz: If we don't use the label, let's make it optional (in Signal component), so we don't have to specify it here at all.

earnubs: Or give it default props.

src/common/components/traffic-lights/traffic-light.css
```diff
@@ -8,15 +8,14 @@ $signal-colour-default: $mid-grey;
 
 .signal {
   background: $signal-colour-default;
-  border-radius: 50%;
+  border-radius: 16px;
```
earnubs: 50% should still work here, or 100%, no?

bartaz: I think it was just following spec: https://github.com/canonical-ols/build.snapcraft.io/issues/354#issuecomment-285355591

But yes, 50% would just work fine (as it's 50% of 32px)

src/common/components/traffic-lights/index.js
pmahnke: @earnubs  and @bartaz I updated this... did I do it ok

jamiedawsonyoung:
[Using colours in the progress meter #354](https://trello.com/c/q3HbBkTA/601-using-colours-in-the-progress-meter-354)

pmahnke:
@bartaz thanks... however, I don't have write access and cannot merge.

PR #469, opened 2017-03-14
Fixes #303 - further first time steps shouldn't show when adding new repo

### LINKED ISSUE #303: [dashboard] Name registration suggestion persists if you choose to add another repo instead [1]
1\. On the front page choose “Set up in minutes”.
2\. Authorize a GitHub account that hasn’t used build.snapcraft.io before.
3\. Add your first repo.
4\. Choose “Add repos…”.

What happens:
4\. The progress is still visible above the “Add repos” dialog, with an instruction that you can’t immediately follow because you’re doing something else: “Great! Next, register a snap name for publishing.”

What should happen:
4\. While the “Add Repos” dialog is open, the progress meter and text should not be present.

[Spec](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.q0b3jpi4vtot): “As an exception to these conditions, the first-time flow should not appear if: … you have at least one repo added, and you reopen the ‘Add Repos’ UI.”

---

Copied from Trello card: https://trello.com/c/vbgELi5h/264-dashboard-name-registration-suggestion-persists-if-you-choose-to-add-another-repo-instead
Labels: Bug

bartaz:
Fix for #303

### DIFF
--- a/src/common/components/first-time-heading/index.js
+++ b/src/common/components/first-time-heading/index.js
@@ -1,52 +1,51 @@
 import React, { Component, PropTypes } from 'react';
+import { connect } from 'react-redux';
 
-import { parseGitHubRepoUrl } from '../../helpers/github-url';
 import { HeadingThree } from '../vanilla/heading';
 import TrafficLights, { SIGNALS } from '../traffic-lights';
 
-import styles from './firstTimeHeading.css';
+import {
+  hasNoRegisteredNames,
+  snapsWithRegisteredNameAndSnapcraftData,
+  snapsWithRegisteredNameAndNoSnapcraftData,
+  snapsWithNoBuilds,
+  hasNoSnaps
+} from '../../selectors';
 
-const hasStoreName = (snap) => snap.store_name;
-const hasStoreNameAndSnapcraftData = (snap) => snap.store_name && snap.snapcraft_data;
-const hasStoreNameButNotSnapcraftData = (snap) => snap.store_name && !snap.snapcraft_data;
+import styles from './firstTimeHeading.css';
 
 class FirstTimeHeading extends Component {
 
-  hasNoBuilds(snap) {
-    const { fullName } = parseGitHubRepoUrl(snap.git_repository_url);
-    const repoBuilds = this.props.snapBuilds[fullName];
-
-    if (repoBuilds && repoBuilds.success) {
-      // if builds for given repo were fetched but there aren't any builds yet
-      return repoBuilds.builds.length === 0;
-    }
-    return false;
-  }
-
   getCurrentState() {
-    const snapsStore = this.props.snaps;
-    const snaps = snapsStore.snaps;
+    const {
+      snapsLoaded,
+      hasNoSnaps,
+      hasNoRegisteredNames,
+      hasSnapsWithRegisteredNameAndNoSnapcraftData,
+      hasNoSnapsWithRegisteredNameAndSnapcraftData,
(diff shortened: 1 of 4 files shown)


src/common/components/first-time-heading/index.js
```diff
@@ -35,19 +35,21 @@ class FirstTimeHeading extends Component {
...
-      // at least one repo has a name but no snapcraft.yaml, and none have both
-      } else if (snaps.filter(hasStoreNameButNotSnapcraftData).length &&
-                 snaps.filter(hasStoreNameAndSnapcraftData).length === 0) {
-        message = 'Okay, your repo is registered. Now push a snapcraft.yaml file, and building will start.';
-        progress = [SIGNALS.DONE, SIGNALS.DONE, SIGNALS.ACTIVE];
-        // only one repo has both a name and snapcraft.yaml, and it hasn’t had a build yet
-      } else if (snaps.filter(hasStoreNameAndSnapcraftData).filter(this.hasNoBuilds.bind(this)).length === 1) {
-        message = 'All set up! Your first build is on the way.';
-        progress = [SIGNALS.DONE, SIGNALS.DONE, SIGNALS.DONE];
+      } else if (this.props.isOnMyRepos) { // further steps are only visible on 'My Repos' page
```
bartaz: "Dashboard" is about to be changed to "My repos", so we probably should stop using this name internally in code too.

src/common/components/first-time-heading/index.js
```diff
@@ -35,19 +35,21 @@ class FirstTimeHeading extends Component {
       if (snaps.length === 0) {
         message = 'Let’s get started! First, choose one or more GitHub repos for building.';
         progress = [SIGNALS.DONE, SIGNALS.ACTIVE, SIGNALS.DEFAULT];
-      // at least one repo, but none have a name yet
-      } else if (snaps.filter(hasStoreName).length === 0) {
```
earnubs: perhaps consider selectors here?

src/common/selectors/index.js
```diff
@@ -1,12 +1,65 @@
...
   (snaps) => {
-    return !snaps.snaps.some((snap) => {
+    return snaps.snaps ? !snaps.snaps.some((snap) => {
       return snap.store_name;
+    }) : true;
+  }
+);
+
+// returns array - snaps that have both registered name and snapcraft data
+export const snapsWithRegisteredNameAndSnapcraftData = createSelector(
```
earnubs: Could we decompose to `getSnapsWithRegisteredName` and `getSnapsWithSnapcraftData` and then use those 2 selectors to compose this and more ... or just leave it as those 2?

src/common/selectors/index.js
```diff
@@ -1,12 +1,65 @@
...
+  [getSnaps],
+  (snaps) => snaps.snaps ? snaps.snaps.length === 0 : true
+);
+
+// returns bool - true if there are no snaps with registered name in state
 export const hasNoRegisteredNames = createSelector(
   [getSnaps],
   (snaps) => {
-    return !snaps.snaps.some((snap) => {
+    return snaps.snaps ? !snaps.snaps.some((snap) => {
```
earnubs: This catches the case where snaps.snaps is null, as it is in initial state? Oops, naughty me. :(

I think it would be simpler if initial/default state of lists was [] rather than null, but that's for another day.

PR #472, opened 2017-03-14
rel=noopener

earnubs:
`<a target=_blank />` is considered a security risk, we don't really have user generated content but doesn't hurt to have the rule.

More about `rel=noopener` here: https://mathiasbynens.github.io/rel-noopener/

Lint rule is defined here: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md

### DIFF
--- a/test/unit/src/server/middleware/t_trusted-networks.js
+++ b/test/unit/src/server/middleware/t_trusted-networks.js
@@ -3,7 +3,7 @@ import ipaddr from 'ipaddr.js';
 
 import { matchNetworks } from '../../../../../src/server/middleware/trusted-networks';
 
-describe.only('matchNetworks', () => {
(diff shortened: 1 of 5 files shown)


test/unit/src/server/middleware/t_trusted-networks.js
```diff
@@ -3,7 +3,7 @@ import ipaddr from 'ipaddr.js';
 
 import { matchNetworks } from '../../../../../src/server/middleware/trusted-networks';
 
-describe.only('matchNetworks', () => {
+describe('matchNetworks', () => {
```
earnubs: I ... don't know how this got here? I rebased and this appears. :(

PR #475, opened 2017-03-14
Add eslint plugin to forbid .only in tests

cjwatson:


cjwatson:
Ah, `eslint-plugin-mocha` has some other useful rules we might as well enable, so I've switched to that.  Is this still OK?

PR #477, opened 2017-03-14
add eslint rule for no-target-blank in jsx, fix occurrences

earnubs:
`<a target=_blank />` is considered a security risk, we don't really have user generated content but doesn't hurt to have the rule.

More about `rel=noopener` here: https://mathiasbynens.github.io/rel-noopener/

Lint rule is defined here: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md


src/common/components/repository-row/dropdowns/register-name-dropdown.js
```diff
@@ -129,7 +134,14 @@ const Caption = (props) => {
         <p><ErrorIcon /> Sorry, that name is already taken. Try a different name.</p>
         <p className={ styles.helpText }>
           If you think you should have sole rights to the name,
-          you can <a href={ FILE_NAME_CLAIM_URL } target='_blank'>file a claim</a>.
+          you can
```
cjwatson: I think you need `{' '}` here too.

PR #484, opened 2017-03-14
unconfigured dropdown first-time user flow

earnubs:
<s>This is blocked because presently the user can never get in this state as they cannot register a name for a repo with no snapcraft.yaml?</s>

### DIFF
--- a/src/common/components/repositories-list/index.js
+++ b/src/common/components/repositories-list/index.js
@@ -1,7 +1,7 @@
 import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
 
-import { hasNoRegisteredNames } from '../../selectors';
+import { hasNoRegisteredNames, hasNoConfiguredSnaps } from '../../selectors';
 import { conf } from '../../helpers/config';
 import {
   checkSignedIntoStore,
@@ -16,7 +16,8 @@ import styles from './repositoriesList.css';
 
 const SNAP_NAME_NOT_REGISTERED_ERROR_CODE = 'snap-name-not-registered';
 
-class RepositoriesList extends Component {
+export class RepositoriesListView extends Component {
+
   fetchAuthData(authStore) {
     if (authStore.authenticated === null) {
       this.props.dispatch(checkSignedIntoStore());
@@ -67,9 +68,17 @@ class RepositoriesList extends Component {
   }
 
   renderRow(snap, index) {
-    const { hasNoRegisteredNames, registerName, snapBuilds, authStore } = this.props;
+    const { hasNoConfiguredSnaps, hasNoRegisteredNames, registerName, snapBuilds, authStore } = this.props;
     const { fullName } = parseGitHubRepoUrl(snap.git_repository_url);
     const isFirstInList = index === 0;
+
+    let registerNameIsOpen = false;
+    let configureIsOpen = false;
+
+    if (isFirstInList) {
+      registerNameIsOpen = hasNoRegisteredNames;
+      configureIsOpen = registerNameIsOpen ? false : hasNoConfiguredSnaps;
+    }
 
     let latestBuild = null;
     const currentSnapBuilds = snapBuilds[fullName];
@@ -88,7 +97,8 @@ class RepositoriesList extends Component {
         fullName={ fullName }
         authStore={ authStore }
         registerNameStatus={ registerNameStatus }
-        registerNameIsOpen={ isFirstInList && hasNoRegisteredNames }
+        registerNameIsOpen={ registerNameIsOpen }
+        configureIsOpen={ configureIsOpen }
       />
     );
   }
@@ -119,14 +129,25 @@ class RepositoriesList extends Component {
   }
 }
 
-RepositoriesList.propTypes = {
+RepositoriesListView.defaultProps = {
+  auth: {},
+  authStore: {},
+  hasNoConfiguredSnaps: true,
+  hasNoRegisteredNames: true,
+  registerName: {},
+  snapBuilds: {},
+  snaps: {}
+};
+
+RepositoriesListView.propTypes = {
   auth: PropTypes.object.isRequired,
   authStore: PropTypes.object.isRequired,
+  dispatch: PropTypes.func,
+  hasNoConfiguredSnaps: PropTypes.bool,
+  hasNoRegisteredNames: PropTypes.bool,
   registerName: PropTypes.object,
-  snaps: PropTypes.object,
   snapBuilds: PropTypes.object,
-  dispatch: PropTypes.func.isRequired,
-  hasNoRegisteredNames: PropTypes.bool
+  snaps: PropTypes.object
 };
 
 function mapStateToProps(state) {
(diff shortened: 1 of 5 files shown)


bartaz:
@earnubs Why user can't register name without snapcraft.yaml?

If you add a repo with no snapcraft.yaml you have both "Not configured" and "Not registered" status, when you click on "Not registered" you can normally sign in and register a name of your choice. No snapcraft.yaml is needed for that:

screen shot 2017-03-15 at 08 47 41

earnubs:
@bartaz hmm, I was seeing an error last night about not having a snapcraft.yaml, will look again today.

earnubs:
cannot-find-snapcraft

src/common/components/repository-row/index.js
```diff
@@ -377,10 +376,8 @@ RepositoryRow.propTypes = {
...
   }),
   dispatch: PropTypes.func.isRequired,
-  registerNameIsOpen: PropTypes.bool
+  registerNameIsOpen: PropTypes.bool,
+  configureIsOpen: PropTypes.bool
 };
 
-// FIXME:
-// `connect` shouldn't be used just to add dispatch to props
-// this makes it much harder to test it as it get wrapped
```
earnubs: I don't agree!

test/unit/src/common/components/repository-list/t_repository-list.js
```diff
@@ -0,0 +1,116 @@
...
+      git_repository_url: 'https://github.com/earnubs/bsi-test-ii'
+    }, {
+      git_repository_url: 'https://github.com/earnubs/bsi-test-iii'
+    }, {
+      git_repository_url: 'https://github.com/earnubs/bsi-test-iv'
+    }, {
+      git_repository_url: 'https://github.com/earnubs/bsi-test-v'
+    }]
+  },
+  registerName: {},
```
earnubs: This makes me wonder if we used defaultProps more could most of these fixtures go away?

test/unit/src/common/components/repository-list/t_repository-list.js
```diff
@@ -0,0 +1,116 @@
...
+  context('with no snaps in list', function() {
+    let wrapper;
+
+    beforeEach(function() {
+      initialState.snaps.snaps = null;
+      wrapper = shallow(<RepositoriesList { ...initialState } />);
+    });
+
+    // XXX Spec says it should be thus:
+    xit('should display \'No repos\' message', function() {
```
earnubs: Only added this because I noticed it in the spec, it's easier to add the test than the feature though.

src/common/components/repository-row/index.js
```diff
@@ -65,12 +65,16 @@ class RepositoryRow extends Component {
   }
 
   componentDidMount() {
-    this.loadState();
+    if (typeof window !== 'undefined') {
```
earnubs: prevent localForage warnings on server side

src/common/components/repositories-list/index.js
```diff
@@ -119,14 +129,25 @@ class RepositoriesList extends Component {
   }
 }
 
-RepositoriesList.propTypes = {
+RepositoriesListView.defaultProps = {
```
earnubs: Decouples the component a little more from store, and makes for fewer fixtures in tests

test/unit/src/common/components/repositories-list/t_repositories-list.js
```diff
@@ -0,0 +1,115 @@
...
+
+  context('with no snaps in list', function() {
+    let wrapper;
+
+    beforeEach(function() {
+      initialState.snaps.snaps = null;
+      wrapper = shallow(<RepositoriesListView { ...initialState } />);
+    });
+
+    xit('should display \'No repos\' message', function() {
```
earnubs: It's in the spec, I'm only adding a test while here.

test/unit/src/common/components/repositories-list/t_repositories-list.js
```diff
@@ -0,0 +1,115 @@
...
+
+    beforeEach(function() {
+      wrapper = getWrapper(initialState);
+    });
+
+    it('should display snap rows', function() {
+      expect(wrapper.children().length).toBe(5);
+    });
+
+    xit('should display snap rows in alphabetical order', function() {
```
earnubs: Also in the spec, I'm only adding a test while here.

src/common/components/repository-row/index.js
bartaz: Sure, but in case of passing `dispatch` only I would probably prefer passing some callback from parent component or using `mapDispatchToProps`.

src/common/components/repository-row/index.js
```diff
@@ -376,11 +391,9 @@ RepositoryRow.propTypes = {
...
+  dispatch: PropTypes.func,
+  registerNameIsOpen: PropTypes.bool,
+  configureIsOpen: PropTypes.bool
 };
 
-// FIXME:
-// `connect` shouldn't be used just to add dispatch to props
-// this makes it much harder to test it as it get wrapped
-export default connect()(RepositoryRow);
+export default connect()(RepositoryRowView);
```
bartaz: What I don't like about this approach is that using empty `connect` call doesn't explicitly define what is connected (what part of state is passed to the component).

I know it's just to pass `dispatch`, but I would prefer it to be more explicit in `connect` call, for example by using `mapDispatchToProps`.

src/common/selectors/index.js
```diff
@@ -5,19 +5,25 @@ import { parseGitHubRepoUrl } from '../helpers/github-url';
 const getSnaps = state => state.snaps;
 const getSnapBuilds = state => state.snapBuilds;
 
-// returns bool - true if there are any snaps in state
+/**
+ * @returns {Boolean} true if there are any snaps in state
```
bartaz: Thank you, I was too lazy to give any proper formatting to these comments ;)

test/unit/src/common/components/repository-row/t_repository-row.js
```diff
@@ -1,90 +1,116 @@
...
-        <RepositoryRow
-          store={store}
-          snap={snap}
-          latestBuild={buildWithoutLog}
-          fullName={fullName}
-          authStore={authStore}
-          registerNameStatus={registerNameStatus}
-        />)
-        .find('RepositoryRow').dive();
+      Object.assign(props.latestBuild, { buildLogUrl: null });
```
bartaz: This overrides `props.latestBuild`, so it may affect other tests that use `props` outside of this context.

earnubs:
Part of #363

src/common/components/repository-row/index.js
```diff
@@ -405,9 +406,19 @@ RepositoryRowView.propTypes = {
...
+  configureIsOpen: PropTypes.bool,
+  authActions: PropTypes.object,
+  nameActions: PropTypes.object,
+  snapActions: PropTypes.object
 };
 
-export default connect()(RepositoryRowView);
+function mapDispatchToProps(dispatch) {
+  return {
+    authActions: bindActionCreators(authStoreActionCreators, dispatch),
```
bartaz: I kind of prefer the idea of only passing that is needed by component, not all actions from given module.

But I don't mind. Lets try different patterns and come up with some conventions later on.

earnubs: Yep, my 2 pence worth, I like the simplicity of the wildcard import.

PR #490, opened 2017-03-15
489: Bugfix: "dispatch" is undefined on dashboard

katiefenn:
- Migrated remaining dispatch reference to mapDispatchToProps function

### DIFF
--- a/src/common/components/repositories-home/index.js
+++ b/src/common/components/repositories-home/index.js
@@ -29,7 +29,7 @@ class RepositoriesHome extends Component {
       }
 
       snaps.snaps.forEach((snap) => {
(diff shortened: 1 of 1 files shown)


src/common/components/repositories-home/index.js
```diff
@@ -29,7 +29,7 @@ class RepositoriesHome extends Component {
       }
 
       snaps.snaps.forEach((snap) => {
-        this.props.dispatch(fetchBuilds(snap.git_repository_url, snap.self_link));
+        this.props.fetchBuilds(snap.git_repository_url, snap.self_link);
```
bartaz: `fetchBuilds` could just take snap as parameter and extract repo URL and self link from it, but that's fine too.

Issue #320, opened 2017-03-03
“Dashboard” is a confusing name [1]


nottrobin:
We’ll probably have an actual dashboard for ISVs in the future, so we shouldn’t call our main screen “Dashboard”.

---

Copied from Trello card: https://trello.com/c/ra654Drs/258-dashboard-is-a-confusing-name
Labels:

matthewpaulthomas:
In Trello, Jamie wrote:
> Some thoughts:
> - Build list
> - Builds

I wrote:
> It’s true that the Dashboard lists builds — but so do the pages for each individual repo. So “Build list” or “Builds” probably wouldn’t help distinguish which page you’re on.
>
> Teresa suggested “My repos”.

matthewpaulthomas:
[Spec updated](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.or9ei8m7j2f0): changed to “My repos”.

pmahnke:
@matthewpaulthomas is this about changing the url as well? or just the title on the page?

bartaz:
There is no "My repos" title on the page. Currently "Dashboard" is only used in header link and in the URL.

So without changing URL this task would be only about changing header link (which is about to be removed anyway in #318.

I think that if we are getting rid of 'dashboard' as a name, we should also update URL , as it's clearly specified as `/user/:username`in spec doc.

bartaz:
Fixed by #492

PR #492, opened 2017-03-15
Update URLs and remove references to dashboard

bartaz:
Fix: #320 

New URLs:

old | new
--------------|----------------
`/dashboard` | `/user/:owner`
`/dashboard/select-repositories` | `/select-repositories` (to avoid GH repo name clash)
`/:owner/:name/builds` | `/user/:owner/:name`
`/:owner/:name/builds/:id` | `/user/:owner/:name/:id`


src/common/routes.js
```diff
@@ -3,21 +3,24 @@ import { Route } from 'react-router';
...
+import EnsureLoggedIn from './containers/routes/ensure-logged-in';
+import EnsureSameUser from './containers/routes/ensure-same-user';
 
 export default (
   <Route component={App}>
     <Route path="/" component={Landing}/>
     <Route component={ EnsureLoggedIn }>
-      <Route path="/dashboard" component={Dashboard}/>
-      <Route path="/dashboard/select-repositories" component={SelectRepositories}/>
+      <Route component={ EnsureSameUser }>
```
bartaz: It would be nice to do it with `onEnter` handler, but it doesn't have access to the store, so wrapping with route and component seems to do the trick.

katiefenn: Nice solution 👍

PR #501, opened 2017-03-16
Clean up database handling code

cjwatson:
* Use the Bookshelf registry plugin

This will mean fewer problems with circular dependencies if the database
schema gets more complicated.

* Run database queries in transactions

At the moment we have no particular need for cleanly rolling back
transactions, but we will soon, and we might as well put good practice
in place now.

I considered using `bookshelf-transaction-manager` to avoid the
`{ transacting: trx }` boilerplate, but it ended up being much more
trouble than it was worth: it doesn't currently wrap all the Bookshelf
model/collection methods, which makes it very confusing and error-prone
to use.

### DIFF
--- a/src/server/db/index.js
+++ b/src/server/db/index.js
@@ -25,5 +25,8 @@ if (!config) {
 
 const connection = knex(config);
 const db = new Bookshelf(connection);
+db.plugin('registry');
 
 export default db;
+
+require('./models');
--- a/src/server/db/index.js
+++ b/src/server/db/index.js
@@ -27,7 +27,6 @@ if (!config) {
 const connection = knex(config);
 const db = new Bookshelf(connection);
 db.plugin('registry');
+registerModels(db);
 
 export default db;
-
-registerModels(db);
--- a/src/server/db/index.js
+++ b/src/server/db/index.js
@@ -4,6 +4,7 @@ import Bookshelf from 'bookshelf';
 
 import { conf } from '../helpers/config';
 import logging from '../logging';
+import registerModels from './models';
 
 let knexConfigPath = conf.get('KNEX_CONFIG_PATH');
 if (!path.isAbsolute(knexConfigPath)) {
@@ -29,7 +30,4 @@ db.plugin('registry');
 
 export default db;
 
-// Import all our models here to ensure that they are added to the registry.
-// This needs to be require rather than import as otherwise we have problems
-// with circular dependencies.
-require('./models');
+registerModels(db);
--- a/src/server/db/index.js
+++ b/src/server/db/index.js
@@ -29,4 +29,7 @@ db.plugin('registry');
 
 export default db;
 
+// Import all our models here to ensure that they are added to the registry.
+// This needs to be require rather than import as otherwise we have problems
+// with circular dependencies.
 require('./models');
--- a/src/server/db/models/github-user.js
+++ b/src/server/db/models/github-user.js
@@ -9,7 +9,7 @@ import db from '../index';
  *   updated_at: update date
  *   last_login_at: last login date
  */
-export const GitHubUser = db.model('GitHubUser', {
+db.model('GitHubUser', {
   tableName: 'github_user',
   hasTimestamps: true
 });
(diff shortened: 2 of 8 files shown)


src/server/db/index.js
```diff
@@ -25,5 +25,8 @@ if (!config) {
 
 const connection = knex(config);
 const db = new Bookshelf(connection);
+db.plugin('registry');
 
 export default db;
+
+require('./models');
```
bartaz: What is this `require` supposed to do? It's not assigned anywhere, it doesn't seem to be exported...

Why not `import`?

I suppose it's some kind of a convention for bookshelf, but it looks quite confusing when looking at the code.

src/server/db/models/github-user.js
```diff
@@ -9,7 +9,7 @@ import db from '../index';
  *   updated_at: update date
  *   last_login_at: last login date
  */
-export const GitHubUser = db.Model.extend({
+export const GitHubUser = db.model('GitHubUser', {
```
bartaz: Does it need to be exported? It seems like we don't import it directly anymore (as it's always used via `db`.

src/server/db/index.js
cjwatson: I've left a comment about what this is doing and why it needs to be `require`.

src/server/db/models/github-user.js
cjwatson: Indeed, that was an oversight; thanks.

cjwatson:
Well, it's not inextricably tied to transactions as such, though of course it overlaps a lot.

The reason it's worth doing it this way is because of circular dependencies.  As soon as the database schema gets much more complicated, tables are likely to have foreign keys referring to other tables, and it ends up being quite common in practice for those relationships to go both ways.  At that point it's not possible to refer to models consistently by way of straightforward imports, and in fact you can get some very mysterious errors from that approach (apparently sometimes the import will appear to work but the symbol will actually be undefined).

It would be possible to take both approaches (export the model *and* add it to a registry just for circular cases), but I think it's less confusing and less error-prone to pick one approach and stick to it.

src/server/db/models/github-user.js
```diff
@@ -9,7 +9,7 @@ import db from '../index';
  *   updated_at: update date
  *   last_login_at: last login date
  */
-export const GitHubUser = db.Model.extend({
+db.model('GitHubUser', {
```
bartaz: I was thinking about a pattern that would solve the issue of circular dependency between models and db.

What I think would help would be to wrap model declarations in function that will get db in param (rather then importing it directly).

```js
export default function registerGitHubUser(db) {
  db.model('GitHubUser', { ... });
};
```

It's not perfect, but it would allow models to be properly imported in `db`:

```js
import * as models from `./models`;

// db module code

models.registerGitHubUser(db);

export default db;
```

Still `registerGitHubUser` has side-effects (changed db that is passed to it), but at least circular dependency is removed and relation between model and db is a bit more explicit.

But I leave it to you to decide what pattern you prefer.

bartaz: Or maybe - to push this idea to the edge, make models that export just a schema, and db module will register them:

`/models/github-user.js`:
```js
export default {
  tableName: 'github_user',
  hasTimestamps: true
}
```
`/index.js`:
```
import * as models from `./models`;

// db module code

db.model('GitHubUser, models.GitHubUser);

export default db;
```

But this makes model module almost empty and useless (as the whole schema is defined in migration file anyway)...

cjwatson: I definitely don't like the second option.  The model name belongs with its definition, IMO.

I did think about approaches along the lines of the first option, but they seem gratuitously different from what the other bookshelf-using codebases I could find are doing, and it's not clear that they really buy us all that much ...

katiefenn: Is it just the ownership of the name that you find problematic? Why not make the model module export `name` and a `schema` properties?

**src/server/db/index.js**
```
import * as models from './models';
...
Object.keys(models).forEach((key) => {
  db.model(models[key]['name'], models[key]['schema']);
});
```

**src/server/db/models/index.js**
```
import * as GitHubUser from './github-user';

export {
  GitHubUser as GitHubUser
};

```

**src/server/db/models/github-user.js**
```
export const name = "GitHubUser";

export const schema = {
  tableName: 'github_user',
  hasTimestamps: true
};
```

PR #503, opened 2017-03-16
338: Edit snapcraft.yml feature

katiefenn:
- Added new repository row dropdown for editing snapcraft.yml
- Changed unconfigured dropdown to only display if snapcraft.yml doesn't
  exist

### DIFF
--- a/src/common/components/repository-row/index.js
+++ b/src/common/components/repository-row/index.js
@@ -10,6 +10,7 @@ import {
   NameMismatchDropdown,
   RemoveRepoDropdown,
   UnconfiguredDropdown,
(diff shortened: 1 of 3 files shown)


src/common/components/repository-row/index.js
```diff
@@ -195,8 +201,8 @@ class RepositoryRow extends Component {
       registerNameStatus
     } = this.props;
 
-    const unconfigured = true;
-    const showUnconfiguredDropdown = unconfigured && this.state.unconfiguredDropdownExpanded;
```
katiefenn: Not sure why `unconfigured` is used here. It will always be `true`.

PR #512, opened 2017-03-17
Convert repositories/snaps tests to async/await

cjwatson:


test/unit/src/common/actions/t_snap-builds.js
```diff
@@ -233,21 +223,18 @@ describe('snap builds actions', () => {
...
-      it('should pass error message from repsponse to action', () => {
-        return store.dispatch(fetchSnap(repositoryUrl))
-          .then(() => {
-            const action = store.getActions().filter(a => a.type === ActionTypes.FETCH_BUILDS_ERROR)[0];
-            expect(action.payload.error.message).toBe('Bad repo URL');
-          });
+      it('should pass error message from repsponse to action', async () => {
+        await store.dispatch(fetchSnap(repositoryUrl));
+        const action = store.getActions().filter(
+          a => a.type === ActionTypes.FETCH_BUILDS_ERROR)[0];
```
bartaz: As much as I like expressiveness of arrow functions, I hate parsing statements like this.

Adding some brackets `(a) => a.type === ActionTypes.FETCH_BUILDS_ERROR` doesn't seem to help a lot, but I think moving `.filter` down one line make is a bit more readable:

```js
const action = store.getActions()
  .filter((a) => a.type === ActionTypes.FETCH_BUILDS_ERROR)[0];
```

PR #516, opened 2017-03-17
Properly show spinner when user has no repositories

### LINKED ISSUE #507: [Regression] 'Empty' my repos flashes before data is loaded
After user signs in (or when my repos page is fully reloaded) empty repos table and first time heading show up for a sec before data starts loading.

progress-bug

This is a regression bug.

bartaz:
Fixes #507 

Showing spinner on 'My repos' until snaps data is properly fetched.
If user has no repos (because they never added one or removed all of them) they will be redirected to select repositories.

Current solution has one case when it works a bit weird (may be considered a bug):
When user has some repositories and adds new one, then they move back from 'select repositories' to 'my repositories' they will see their old list of repos (without the new one) until list gets refreshed.
There is also no spinner of any kind (because we don't want to show spinner during polling).
So there is a visual delay between adding a repo and it being rendered no the list.

### DIFF
--- a/src/common/components/repositories-home/index.js
+++ b/src/common/components/repositories-home/index.js
@@ -55,7 +55,10 @@ class RepositoriesHome extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (this.props.snaps.success !== nextProps.snaps.success) {
+    const snapsLength = nextProps.snaps.snaps && nextProps.snaps.snaps.length;
+
+    if ((this.props.snaps.success !== nextProps.snaps.success)
+        || (snapsLength === 0)) {
       this.fetchData(nextProps);
     }
   }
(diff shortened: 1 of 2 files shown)


src/common/components/repositories-home/index.js
```diff
@@ -55,7 +55,10 @@ class RepositoriesHome extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (this.props.snaps.success !== nextProps.snaps.success) {
+    const snapsLength = nextProps.snaps.snaps && nextProps.snaps.snaps.length;
+
+    if ((this.props.snaps.success !== nextProps.snaps.success)
+        || (snapsLength === 0)) {
```
bartaz: Make sure user is redirected to select repositories view when all snaps are removed.

src/common/components/repositories-home/index.js
```diff
@@ -89,9 +92,14 @@ class RepositoriesHome extends Component {
...
-    // show spinner until we know if user has any enabled repos
-    return (snaps.isFetching && !hasSnaps)
+    const hasSnaps = (snaps.snaps && snaps.snaps.length > 0);
+
+    // show spinner if snaps data was not yet fetched (snaps list is empty)
+    // (to avoid spinner during polling we are not checking `success` or `isFetching`
+    //
+    // when snaps are loaded and user doesn't have any, they will be redirected
+    // to select repositories (so spinner won't be showing endlessly)
+    return !hasSnaps
```
bartaz: In case of 'My repos' we don't care about `isFetching` or success.

If we don't know if user has any repos (initial state) we show spinner.
If we know they don't have any, we redirect to select repositories.

Otherwise we render repositories list. There is no spinner shown during polling.

PR #518, opened 2017-03-18
Add developer uptake metrics

### LINKED ISSUE #446: Gather metrics for KPI dashboard [2]
We need basic visibility on uptake and fall-off points, tracking how many developers are at each major step (github logged in, name registered, snap pushed to edge) from the KPI dashboard

cjwatson:
We now track developers adding/removing snaps, registering snap names,
and requesting builds, and emit aggregate metrics indicating how many
developers have reached various stages in the process.

Fixes #446.

### DIFF
--- a/src/server/db/models/github-user.js
+++ b/src/server/db/models/github-user.js
@@ -7,9 +7,32 @@ export default function register(db) {
    *   created_at: creation date
    *   updated_at: update date
    *   last_login_at: last login date
+   *   snaps_added: number of snaps added
+   *   snaps_removed: number of snaps removed
+   *   names_registered: number of snap names registered
+   *   builds_requested: number of snap builds requested
    */
   db.model('GitHubUser', {
     tableName: 'github_user',
     hasTimestamps: true
+  }, {
+    // Increment a metric for a user.  Where possible, this should be called
+    // at the *start* of a transaction spanning any external API calls that
+    // are tracked by the metric in question: this means that either a
+    // failure to increment the metric or a failure in the external API call
+    // will cause both to be rolled back.
+    incrementMetric: async (query, metricName, delta, options) => {
+      const row = await db.model('GitHubUser').where(query).fetch(options);
+      if (row) {
+        const value = row.get(metricName);
+        // Metrics may often require non-trivial initialization, for example
+        // when a metric is introduced to the site after the feature that
+        // it's measuring.  To cope with this, only increment a metric if it
+        // has already been initialized.
+        if (value !== undefined && value !== null) {
+          await row.save({ [metricName]: value + delta }, options);
+        }
+      }
+    }
   });
 }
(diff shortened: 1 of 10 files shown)


src/server/db/models/github-user.js
```diff
@@ -7,9 +7,32 @@ export default function register(db) {
    *   created_at: creation date
    *   updated_at: update date
    *   last_login_at: last login date
+   *   snaps_added: number of snaps added
+   *   snaps_removed: number of snaps removed
+   *   names_registered: number of snap names registered
```
bartaz: These are just an incremental counters per user? It doesn't matter if I add and remove same snap over and over again, it will bump up the stat?

cjwatson: Yes.  To quote Evan, "we can add" [or presumably actually subtract] "them together at the level of Grafana".

For the moment the metric only cares about when this is non-zero; the idea is to count developers who've got past certain steps in the process.

src/server/handlers/store.js
```diff
@@ -1,23 +1,35 @@
...
-  const response = await fetch(`${conf.get('STORE_API_URL')}/register-name/`, {
-    method: 'POST',
-    headers: {
-      'Authorization': `Macaroon root="${root}", discharge="${discharge}"`,
-      'Content-Type': 'application/json',
-      'Accept': 'application/json'
-    },
-    body: JSON.stringify({ snap_name: snapName })
+  await db.transaction(async (trx) => {
+    // XXX cjwatson 2017-03-17: This may need to move somewhere else in
```
bartaz: Maybe it can be moved to `authorizeSnap` in `handlers/launchpad.js` as it seems we call authorize directly after registering name.

src/server/db/models/github-user.js
```diff
@@ -7,9 +7,32 @@ export default function register(db) {
...
+    // will cause both to be rolled back.
+    incrementMetric: async (query, metricName, delta, options) => {
+      const row = await db.model('GitHubUser').where(query).fetch(options);
+      if (row) {
+        const value = row.get(metricName);
+        // Metrics may often require non-trivial initialization, for example
+        // when a metric is introduced to the site after the feature that
+        // it's measuring.  To cope with this, only increment a metric if it
+        // has already been initialized.
+        if (value !== undefined && value !== null) {
```
bartaz: In theory this could be shortened to `value != null` because `undefined == null` (note double not tripple eq), but I'm not sure if we want to have the an exception from using `===` just for this null comparison case.

src/server/handlers/launchpad.js
```diff
@@ -384,6 +394,60 @@ const internalFindSnapsByPrefix = async (urlPrefix) => {
...
+        row.set({ snaps_removed: 0 });
+      }
+      if (rowData.names_registered === undefined ||
+          rowData.names_registered === null) {
+        let namesRegistered = 0;
+        for (const snap of snaps) {
+          if (snap.store_name) {
+            namesRegistered += 1;
+          }
+        }
```
katiefenn: Could `namesRegistered` be generated using `Array.prototype.find` and an arrow function?

```
const namesRegistered = snaps.filter((snap) => snap.store_name).length;
```

src/server/db/models/github-user.js
cjwatson: Mm.  I think this is a bit too magic and easy to get wrong for me.

PR #522, opened 2017-03-20
Convert remaining code to async/await

cjwatson:
This covers the last few unit tests, as well as various odds and ends of
code elsewhere.

### DIFF
--- a/test/unit/src/common/actions/t_create-snap.js
+++ b/test/unit/src/common/actions/t_create-snap.js
@@ -73,46 +73,43 @@ describe('create snap actions', () => {
     });
 
     afterEach(() => {
+      scope.done();
       nock.cleanAll();
     });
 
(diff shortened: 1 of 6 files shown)


test/unit/src/common/actions/t_create-snap.js
```diff
@@ -73,46 +73,43 @@ describe('create snap actions', () => {
...
+    it('succeeds if creating webhook says already created', async () => {
       scope
         .post('/api/github/webhook', { owner: 'foo', name: 'bar' })
         .reply(422, {
           status: 'error',
           payload: { code: 'github-already-created' }
         });
-      return createWebhook(repository)
-        .then(() => scope.done());
+      await createWebhook(repository);
```
bartaz: It's a bit confusing that there is no `expect` in this test at all.

But I guess it expects not to throw any error...

PR #530, opened 2017-03-21
Disable Add button when no repos are selected

### LINKED ISSUE #374: “Add” clickable and does nothing when no repos are selected [1]
1. Go to the dashboard.
2. Choose “Add repos…”

What you see: The “Add” button is opaque and clickable.

What you should see: The “Add” button is semitransparent and unclickable.

<https://goo.gl/B9Wkdb>: “‘Add’ should be disabled whenever no repos are selected in the dialog.”

bartaz:
Fixes #374

### DIFF
--- a/test/unit/src/common/components/select-repository-list/t_select-repository-list.js
+++ b/test/unit/src/common/components/select-repository-list/t_select-repository-list.js
@@ -7,37 +7,88 @@ import {
 } from '../../../../../../src/common/components/select-repository-list';
 
 describe('<SelectRepositoryListComponent /> instance', function() {
+  let props;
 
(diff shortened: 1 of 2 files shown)


test/unit/src/common/components/select-repository-list/t_select-repository-list.js
```diff
@@ -7,37 +7,88 @@ import {
...
+        snaps: []
+      }
+    };
+  });
+
+  context('when user doesn’t have any repos selected', function() {
+    let wrapper;
+
+    beforeEach(function() {
+      props.selectRepositoriesForm = {
```
earnubs: do you need this? it looks the same as in original props?

Issue #318, opened 2017-03-03
[header] Three adjacent header elements look the same, behave differently [1]


nottrobin:
Currently the header contains:

- “Hi, {name}”, static text
- “Dashboard”, a link
- “Sign out”, a button

These three items are styled exactly the same but behave differently.

Possible partial solutions:

- Replace “Hi, ” with your GitHub profile picture
- Remove “Dashboard” from the header altogether
- Style “Sign out” like a button

---

Copied from Trello card: https://trello.com/c/4tFtWj7S/253-header-three-adjacent-header-elements-look-the-same-behave-differently
Labels: design

matthewpaulthomas:
> Replace “Hi, ” with your GitHub profile picture

Now reported as #426.

> Remove “Dashboard” from the header altogether

This part can be done once #310 (consistent hierarchy navigation) is implemented.

jamiedawsonyoung:
## Design proposal
- Replaced "Hi" with GitHub avatar
- Removed the dashboard link

#### Signed out
![header-signed-out](https://cloud.githubusercontent.com/assets/6909617/23711168/fdacbce0-0416-11e7-947a-288d8279662f.png)

#### Signed in
![header-signed-in](https://cloud.githubusercontent.com/assets/6909617/23711227/2af4d5f2-0417-11e7-943b-c617e4f9d75d.png)

jamiedawsonyoung:
@matthewpaulthomas Could you have a look and +1 this please too.

nottrobin:
#305 may become relevant.

jamiedawsonyoung:
Reviewed with @matthewpaulthomas (and got +1) so the spec is below: 

### Sign in
- Styling as current build

### Signed in
Avatar from GitHub
- height: 24px
- width: 24px
- border-radius: 12px 
- padding: 10px

Username from GitHub
- Text color: $color-mid-dark
- Font size: 14px
- padding: 10px

*Let me know if you need any more styling*

jamiedawsonyoung:
![](https://github.trello.services/images/mini-trello-icon.png) [#318 Three adjacent header elements look the same, behave differently](https://trello.com/c/CI8zpkRN/595-318-three-adjacent-header-elements-look-the-same-behave-differently)

magdalena-mirowicz:
possibly reduce to simpler solution for the launch @matthewpaulthomas - to be discussed in standup

magdalena-mirowicz:
should be solved by #305

Issue #310, opened 2017-03-03
[repo and build details] Navigation breadcrumb [2]


nottrobin:
---

Copied from Trello card: https://trello.com/c/I69KPPVZ/146-repo-and-build-details-navigation-breadcrumb
Labels:

matthewpaulthomas:
[Spec updated](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.15nyc5j48m1e): “Hierarchy navigation”.
![navigation hierarchy on repo page](https://cloud.githubusercontent.com/assets/19801137/23655743/49045b6a-032e-11e7-9053-1015c31843ba.png)
![navigation hierarchy on build page](https://cloud.githubusercontent.com/assets/19801137/23655752/5130b81a-032e-11e7-96d2-2e34966a1f05.png)

bartaz:
The breadcrumb should follow the pattern from Vanilla:
https://github.com/ubuntudesign/vanilla-design/tree/master/Breadcrumbs

PR #531, opened 2017-03-21
Removes 'My repos' link from the header

matthewpaulthomas:
A _partial_ fix for #318. The link in the header should no longer be necessary, because (thanks to #310) it is now present in the breadcrumbs on every other page.

### DIFF
--- a/src/common/components/header/index.js
+++ b/src/common/components/header/index.js
@@ -24,7 +24,6 @@ export default class Header extends Component {
                     Hi, {user.name || user.login}
                   </span>
                 }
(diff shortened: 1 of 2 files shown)


bartaz:
Thanks for the fix.

Removing the link made some test fail, so they would need to be updated too.

We have 2 test cases to check if "My repos" link is rendered in header, so all that needs to be done is removing them from
https://github.com/canonical-ols/build.snapcraft.io/blob/master/test/unit/src/common/components/header/t_header.js#L20-L22
and https://github.com/canonical-ols/build.snapcraft.io/blob/master/test/unit/src/common/components/header/t_header.js#L47-L49

PR #536, opened 2017-03-23
Added spinner on 'Add' repository button

### LINKED ISSUE #313: Spinner is 'in flow' which means when it appears it has to move elements below it down the page. [1]
Spinner should not be in flow and should sit in z-index with some kind of overlay, avoiding the case where buttons move away from the users pointer because the spinner appeared.

---

Copied from Trello card: https://trello.com/c/CTH7MTvB/307-spinner-is-in-flow-which-means-when-it-appears-it-has-to-move-elements-below-it-down-the-page
Labels:

bartaz:
Fixes #313 

Adding spinner to 'Add' repository button and disabling whole repositories form while waiting for response.

> https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.43owh6tjv12q: “When you choose ‘Add’, everything in the dialog should become disabled, and the text of the ‘Add’ button should change to a spinner.”

### DIFF
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -14,7 +14,7 @@ import PageLinks from '../page-links';
 import Button, { LinkButton } from '../vanilla/button';
 import { HeadingThree } from '../vanilla/heading';
 import { fetchUserRepositories } from '../../actions/repositories';
-import { hasRepository } from '../../helpers/repositories';
+import { hasRepository, hasSnapForRepository } from '../../helpers/repositories';
 import { isAddingSnaps } from '../../selectors';
 
 import styles from './styles.css';
@@ -65,11 +65,12 @@ export class SelectRepositoryListComponent extends Component {
   }
 
   renderRepository(repo) {
+    const { repositoriesStatus, selectRepositoriesForm, snaps } = this.props;
     const { fullName } = repo;
-    const status = this.props.repositoriesStatus[fullName] || {};
+    const status = repositoriesStatus[fullName] || {};
 
-    const selected = this.props.isRepoSelected(repo);
-    const enabled = this.props.isRepoEnabled(repo);
+    const selected = hasRepository(selectRepositoriesForm.selectedRepos, repo);
+    const enabled = snaps.success && hasSnapForRepository(snaps.snaps, repo);
 
     return (
       <SelectRepositoryRow
@@ -169,8 +170,6 @@ SelectRepositoryListComponent.propTypes = {
   repositoriesStatus: PropTypes.object,
   selectRepositoriesForm: PropTypes.object,
   isAddingSnaps: PropTypes.bool,
-  isRepoSelected: PropTypes.func,
-  isRepoEnabled: PropTypes.func,
   onSelectRepository: PropTypes.func,
   router: PropTypes.object.isRequired,
   dispatch: PropTypes.func.isRequired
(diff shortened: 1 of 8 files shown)


src/common/components/select-repository-list/index.js
```diff
@@ -63,20 +65,22 @@ export class SelectRepositoryListComponent extends Component {
...
     return (
       <SelectRepositoryRow
         key={ `repo_${repo.fullName}` }
         repository={ repo }
-        buttonLabel={ status.isFetching ? 'Creating...' : 'Create' }
-        buttonDisabled={ status.isFetching }
         onChange={ this.onSelectRepository.bind(this, repo) }
         errorMsg= { this.getErrorMessage(status.error) }
-        checked={ hasRepository(selectedRepos, repo) }
-        isEnabled={ enabled }
```
bartaz: It was confusing to have 2 properties (`checked` and `isEnabled`) that would define is checkbox is checked.

src/common/components/select-repository-list/index.js
```diff
@@ -96,41 +100,21 @@ export class SelectRepositoryListComponent extends Component {
...
-  filterEnabledRepos(repositories) {
-    const { success, snaps } = this.props.snaps;
-
-    if (success && snaps.length) {
-      for (let i = repositories.length; i--;) {
-        for (let j = snaps.length; j--;) {
-          const enabledRepo = snaps[j].git_repository_url;
-
-          if (enabledRepo === repositories[i].url) {
-            repositories[i].enabled = true;
```
bartaz: This was having a side effect of modifying redux state (without triggering any action).

Values passed from store as props are still references to objects in the store, so any modifications to them (or adding new properly) would appear in the store.

src/common/components/select-repository-list/index.js
```diff
@@ -202,7 +190,16 @@ function mapStateToProps(state) {
     snaps,
     repositories,
     repositoriesStatus,
-    selectRepositoriesForm
+    selectRepositoriesForm,
+    isAddingSnaps: isAddingSnaps(state),
+    isRepoSelected: (repo) => hasRepository(selectRepositoriesForm.selectedRepos, repo),
```
bartaz: reselect memoized selectors don't support passing any parameters to them (they say all data needed for selector should be in the store), so to check particular repo we make some fake 'selectors' here.

katiefenn: Is this what selectors are really for? [The docs](http://redux.js.org/docs/recipes/ComputingDerivedData.html) state that selectors are for computing derived data from the props. This returns a function, it does not return transformed data.

This seems like a classic helper function, it could be a pure function defined after the component class and fulfil the same role.

src/common/components/select-repository-list/index.js
```diff
@@ -202,7 +190,16 @@ function mapStateToProps(state) {
...
+    isRepoEnabled: (repo) => {
+      const { success, snaps } = state.snaps;
+      if (success && snaps.length) {
+        return snaps.some((snap) => snap.git_repository_url === repo.url);
+      }
+      return false;
+    }
```
katiefenn: Ditto.

src/common/components/select-repository-list/index.js
bartaz: Selectors (as far as I currently understand them) allow to compute some derived data based on current state.

These functions here (`isRepoSelected` and `isRepoEnabled`) also compute derived data based on a state, but only for given repository.

Yes, they are kind of helper functions and could live outside of the class and be passed the data instead of getting it from state here.
But for me they felt similar enough to selectors to include them as props here.

As mentioned by @earnubs they probably would be simplified (or not needed at all) once snap store is refactored. So maybe we can revisit it then?

PR #541, opened 2017-03-24
Refactor redux part i

earnubs:
Anything commented XXX will be addressed/deleted or moved to issue/conversation.

- [x] merge API_CALL middleware (@katiefenn)
- [x] <s>s/repository/repo -- this isn't very important and can be dropped?</s>
- [x]  rename repository build actions and comment (@bartaz)
- [x]  tidy up entities.js (@bartaz)
- [x]  more test coverage for select-repository-list component, inc test and fix pageSlice method (@earnubs)
- [x]  more test coverage for repository reducers and actions (@earnubs)
- [x]  add processStrategy to repo and owner entities to limit props and camelCase them (@bartaz)
- [x] remove prefix from props `__isSelected` etc (@bartaz)
- [x] when repo is removed in BSI it still is enabled in select repos list


src/common/reducers/repositories.js
```diff
@@ -1,42 +1,69 @@
...
+}
 
 export function repositories(state = {
-  // has no server side content, so initial state is always 'fetching'
   isFetching: false,
-  success: false,
-  error: null,
-  repos: [],
+  //success: false, // implicit in !!error
+  error: false,
```
earnubs: Should be null </bartek>

src/common/components/select-repository-list/index.js
```diff
@@ -61,27 +60,28 @@ export class SelectRepositoryListComponent extends Component {
...
-        checked={ hasRepository(selectedRepos, repo) }
         isEnabled={ enabled }
       />
     );
   }
 
-  onSelectRepository(repository) {
-    this.props.dispatch(toggleRepository(repository));
+  onSelectRepository(id) {
+    this.props.dispatch(selectRepository(id));
```
earnubs: put back to toggle

src/common/actions/repositories.js
```diff
@@ -39,10 +41,9 @@ export function fetchUserRepositories(pageNumber) {
       if (result.status !== 'success') {
         throw getError(response, result);
       }
-      dispatch(setRepositories({
-        repos: result.payload.repos,
-        links: result.pageLinks
-      }));
+
+      // XXX deleted chunk should be handled in reducer?
+      dispatch(setRepositories(result));
```
earnubs: Handle the whole result in the reducer, rather than deconstructing it in the action creator. This keeps things simpler conceptually I think.

src/common/actions/repositories.js
```diff
@@ -53,8 +54,124 @@ export function fetchUserRepositories(pageNumber) {
...
   return {
-    type: FETCH_REPOSITORIES_ERROR,
+    type: REPOSITORIES_FAILURE,
     payload: error,
     error: true
   };
 }
+
+// XXX may need to split out the fetch from the success and failure dispatch
+// to ensure than setRepositories happens in sync with refreshed user snaps
```
earnubs: I need to come back to this.

src/common/actions/repositories.js
```diff
@@ -53,8 +54,124 @@ export function fetchUserRepositories(pageNumber) {
...
+      }
+    );
+    return Promise.all(promises).then(() => dispatch(push('/dashboard')));
+  };
+}
+
+export function buildRepository(repository) {
+  const { id, url, owner, name } = repository;
+
+  return async (dispatch) => {
```
earnubs: I am wondering about how this would look as an action that did the async bits, then an action that wrapped it, returning a promise:

```
function fetchFoo() {
  return async (dispatch) => { ... }
}

function orchestrateFoo() {
  return dispatch(fetchFoo()
    .then(
      thing => dispatch(fooSuccess(thing)),
      error => dispatch(fooFailure(error())
  )
)}

File: src/common/actions/repositories.js [hunk 6]
```diff
@@ -53,8 +54,124 @@ export function fetchUserRepositories(pageNumber) {
...
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ repository_url: url }),
+        credentials: 'same-origin'
+      });
+
+      await checkStatus(response);
+      dispatch(buildRepositorySuccess(id));
+    } catch (error) {
+      dispatch(buildRepositoryError(id, error));
+      return Promise.reject(error);
```
earnubs: Do reject the promise, either throw or `Promise.reject`, I kind of feel we should be handling these errors rather than throwing them but it's out of scope for this PR.

src/common/components/select-repositories-page/index.js
```diff
@@ -52,6 +42,16 @@ class SelectRepositoriesPage extends Component {
       </div>
     );
   }
+
+  fetchData(props) {
```
earnubs: I wonder if it's possible to tell eslint to make sure custom methods come after render?

src/common/components/select-repository-list/index.js
```diff
@@ -3,18 +3,23 @@ import { connect } from 'react-redux';
...
+  fetchUserRepositories,
+  resetRepository,
+  toggleRepositorySelection,
+} from '../../actions/repositories';
+import {
+  getEnabledRepositories,
+  getSelectedRepositories,
+  getRepositoriesToBuild,
+  hasFailedRepositories
+} from '../../selectors/index.js';
```
earnubs: More selectors, less state. Hopefully. :)

src/common/components/select-repository-list/index.js
```diff
@@ -25,18 +30,13 @@ const SNAP_NAME_NOT_REGISTERED_ERROR_CODE = 'snap-name-not-registered';
...
   componentDidMount() {
-    this.props.dispatch(createSnapsClear());
-    this.props.dispatch(unselectAllRepositories());
-  }
-
-  componentWillReceiveProps(nextProps) {
-    const repositoriesStatus = nextProps.repositoriesStatus;
-    const ids = Object.keys(repositoriesStatus);
-    if (ids.length && ids.every((id) => repositoriesStatus[id].success)) {
-      this.props.router.push('/dashboard');
```
earnubs: This now happens if the async actions to buildRepositories all resolve.

src/common/components/select-repository-list/index.js
```diff
@@ -50,99 +50,104 @@ export class SelectRepositoryListComponent extends Component {
...
 
-  renderRepository(repo) {
-    const { fullName, enabled } = repo;
-    const status = this.props.repositoriesStatus[fullName] || {};
-    const { selectedRepos } = this.props.selectRepositoriesForm;
+  renderRepository(id) {
+    const repository = this.props.entities.repos[id];
+    const { full_name } = repository;
+
+    const isEnabled = !!this.props.enabledRepositories[id];
```
earnubs: isEnabled is now just a lookup on a selector.

src/common/components/select-repository-list/index.js
```diff
@@ -50,99 +50,104 @@ export class SelectRepositoryListComponent extends Component {
...
-
-          if (enabledRepo === repositories[i].url) {
-            repositories[i].enabled = true;
-            break;
-          } else {
-            repositories[i].enabled = false;
-          }
-        }
-      }
+  pageSlice(array, pageLinks) {
```
earnubs: repository entities in state are now merged with `lodash.union`, so we actually preserve older results, which means slicing the list to handle pagination.

src/common/helpers/api.js
```diff
@@ -12,7 +12,7 @@ export function getError(response, json) {
 
 export function checkStatus(response) {
   if (response.status >= 200 && response.status < 300) {
-    return response;
+    return false;
```
earnubs: this looks like an artifact of testing, will revert.

src/common/reducers/repositories.js
```diff
@@ -1,42 +1,117 @@
...
+import union from 'lodash/union';
+
 import * as ActionTypes from '../actions/repositories';
-import { parseGitHubRepoUrl } from '../helpers/github-url';
+
+// XXX move to entities.js
+// XXX should we be handling snaps here too?
+// no, because it's not part of the normalized result from github
+// so this should really be called github_entities
+// no, decompose instead ...
```
earnubs: needs some tidy ups still...

src/common/reducers/repositories.js
```diff
@@ -1,42 +1,117 @@
...
+export function entities(state = { snaps: {}, repos: {} }, action) {
+  if (action.payload && action.payload.entities) {
+    return merge({}, state, action.payload.entities);
+  }
+
+
+  const reposInitialState = {
+    __isSelected: false,
```
earnubs: I am not wedded to __format, but happy with whatever namespace distinguishes props from api and client props to avoid clobbering.

src/common/reducers/repositories.js
```diff
@@ -1,42 +1,117 @@
...
+        ...state,
+        repos: {
+          ...state.repos,
+          [action.payload.id]: {
+            ...state.repos[action.payload.id] || {},
+            ...reposInitialState, // clear state from building by deselecting
+            __isSelected: !wasSelected
+          }
+        }
+      };
```
earnubs: These can be tidied up.

src/common/reducers/snaps.js
```diff
@@ -39,14 +42,17 @@ export function snaps(state = {
...
     case ActionTypes.FETCH_SNAPS_SUCCESS:
       return {
         ...state,
         isFetching: false,
         success: true,
         snaps: [
-          ...action.payload
+          ...action.payload.payload.snaps
         ],
+        ids: union(state.ids, action.payload.result),
```
earnubs: Adding this along side the current implementation, with the goal of moving to similar pattern for snaps as established here in repositories.

src/common/selectors/index.js
```diff
@@ -1,8 +1,14 @@
...
+import pick from 'lodash/pick';
 
 import { parseGitHubRepoUrl } from '../helpers/github-url';
 
+const getRepositoriesIndex = state => state.repositories.ids;
+const getRepositories = state => state.entities.repos;
+const getRepositoryOwners = state => state.entities.owners;
 const getSnaps = state => state.snaps;
+const _getSnapsIndex = state => state.snaps.ids;
+const _getSnaps = state => state.entities.snaps;
```
earnubs: underscored because they're the newer pattern and would need to refactor out the non underscore version in the work to tackle snaps.

src/server/handlers/github.js
```diff
@@ -150,10 +152,9 @@ export const listRepositories = async (req, res) => {
 
   const body = {
     status: 'success',
-    payload: {
-      code: 'github-list-repositories',
-      repos: response.body
-    }
+    code: 'github-list-repositories',
+    // XXX drop response specific prop, make response a little more general purpose
+    ...normalize(response.body, repoList)
```
earnubs: I find payload in the response confusing when the action sets payload.

src/server/handlers/launchpad.js
```diff
@@ -400,7 +402,8 @@ export const findSnaps = async (req, res) => {
       payload: {
         code: 'snaps-found',
         snaps: snaps
-      }
+      },
+      ...normalize(snaps, snapList)
```
earnubs: In this case we're not touching the payload, snaps refactor is for another pr, but we need snaps as entities for repositories to select on enabled repos.

src/server/handlers/schema.js
```diff
@@ -0,0 +1,13 @@
+import { schema } from 'normalizr';
+
+export const owner = new schema.Entity('owners');
+// XXX repository or repo?
```
earnubs: I've tried to consistently use `repository` except here ...

src/common/actions/repositories.js
```diff
@@ -1,18 +1,20 @@
...
-export const SET_REPOSITORIES = 'SET_REPOSITORIES';
-export const SET_REPOSITORY_PAGE_LINKS = 'SET_REPOSITORY_PAGE_LINKS';
+export const REPOSITORIES_REQUEST = 'REPOSITORIES_REQUEST';
+export const REPOSITORIES_SUCCESS = 'REPOSITORIES_SUCCESS';
+export const REPOSITORIES_FAILURE = 'REPOSITORIES_FAILURE';
+
 
 export function setRepositories(repos) {
   return {
-    type: SET_REPOSITORIES,
```
earnubs: The action doesn't set repositories, the reducer does that, the action is some excitable teenager shouting "hey guys, we got our repositories list from github!" with no clue what to do with that list.

src/common/actions/repositories.js
bartaz: Does `orchestrateFoo` need to be based on promise rather then async/await notation?

This should be more or less equivalent?
```
async function orchestrateFoo() {
  try {
     const thing = await dispatch(fetchFoo());
    dispatch(fooSuccess(thing));
  } catch (error) {
    dispatch(fooFailure(error())
  }
)}
```

earnubs: I think you're right those are equivalent, although the Promise that orchestrateFoo returns will never reject (because you've caught the error and don't explicitly reject).

I'm not even sure there's an issue here, it's just a thought based on wanting to wrap getting snaps and repositories and responding to both being complete differently... but it seems so far to just work by wrapping them in Promise.all as they are, pending more testing :)

src/common/actions/repositories.js
bartaz: In this case it was probably meant to be resolved - error was handled (action dispatched) so no need to reject it.

I get your point that we may want to combine them, you may end up thinking they were successfuly resolved when they actually resulted in error.
But does the parent 'action' (that does Promise.all) need to know what was the result of child promises - I mean does it care if they were successes or failures or just that they finished?

Because each action results on it's own in state change, so error result is in state. So maybe we don't need to know if they failed or succeeded on 'parent' level, just that they finished fetching - and we can get the result from the state?

src/common/components/select-repository-list/index.js
earnubs: Which seems to have the side effect of there being a delay on /dashboard from the intial render to the new lists of snaps appearing.

earnubs: This could be addressed with some kind of state loading message, because we know the user selected some repos, we know that was successful, we just don't know the result yet.

src/common/components/select-repository-list/index.js
earnubs: Which leads to potentially dropping/changing paging of select-repositories in future to be more like "Show more..." @matthewpaulthomas @jamiedawsonyoung

src/common/reducers/repositories.js
earnubs: I want to make the entities reducer it's own module/folder in reducers/, and then it can be split up and import helpers for the various functions.

src/common/actions/repositories.js
earnubs: Good question. The pattern we have is that if an async action succeeds we do `X`, and if it fails we do `Y`. Where we have batch actions we switch to then looking at state, which is valid but has some complexities, but why not just continue the same pattern? If the batch action is successful we do 'Z', etc.

The complexity of looking at state for errors is, I think, that you need to then include `success` in state, or some indication that you performed an action on the item, because checking for errors doesn't tell you if an action was ever performed (a fresh list has no errors), so in the case where you want to redirect if everything was successful, it just seems easier to stick to the the resolve/reject pattern than to decorate state.

earnubs: This is treading on what do to in action creators vrs what to do in reducers territory. :)

src/common/components/select-repository-list/index.js
earnubs: "Show more" + "refresh".

earnubs: #564 conversation for related bugs

src/common/actions/repositories.js
bartaz: > Logic to generate an action resides in action creator, logic to react to it resides in the reducer, reducer must be pure.
http://stackoverflow.com/questions/34098125/how-to-divide-the-logic-between-redux-reducers-and-action-creators

I trust this guy, he seems to know what he is talking about ;)

src/common/actions/repositories.js
bartaz: Isn't the promise-based orchestrateFoo do the same?

```
function orchestrateFoo() {
  return dispatch(fetchFoo()
    .then(
      thing => dispatch(fooSuccess(thing)),
      error => dispatch(fooFailure(error())
  )
)}
```

You specify rejected handler here, which just dispatches an action, so no other error or rejected promise is returned. What is redurned by `.then` is always a resolved promise, because any errors that may cause `fetchFoo` to reject will be caught by error handler in the same way try-catch would catch in in async-await).

src/common/components/select-repositories-page/index.js
bartaz: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md

src/common/reducers/repositories.js
bartaz: I'm not big fan of `__` prefix as it's going to make usages of this look like `entitites.repos[is].__isSelected` which is a bit confusing - why I need to remember which properties are custom with `__` and which are not?

Similar issue is with camelCase vs snake_case in names. Object returned by GH and LP APIs are snake_cased while the propetries we use (and add here to the same objects) are camelCased. So it ends up like:`entities.repo[i].full_name` but `entities.repo[i].__isSelected`.

But I'm not sure how to make those API and internal properties compatible in notation but without possible clobbering - apart from additional nesting which is not perfect either.

src/common/reducers/repositories.js
bartaz: I think I've seen in redux examples that sometimes they would have a separate reducer for single entity (repo in our case), that will internally deal with state change for a single repo, and `entities` reducer will only need to select the entity based on id.

Mind my simplified pseudo-code:
```js
repo = (state, action) => {
  switch(action.type) {
  case REPOSITORY_TOGGLE_SELECT:
    return {
      ...state,
      __isSelected: !state.__isSelected
    }
  }
}

entities = (state, action) => {
switch(action.type) {
  case REPOSITORY_TOGGLE_SELECT:
    return {
      ...state,
      repos: {
        ...state.repos,
        [action.payload.id]: repo(state.repos[action.payload.id], action)
      }
    }
  }
}
```

src/server/handlers/github.js
bartaz: I don't find it confusing, but unnecessary for sure, especially when the actions blindly pass whole response to reducer which results in `action.payload.payload` type of calls.

But we use this pattern for all of our responses from BSI APIs, which means a lot of places to update if we decide to switch to new scheme and very confusing 'in-between' state in the meantime.

src/server/handlers/launchpad.js
bartaz: Line of comment in the code would be useful to anyone doing the snaps refactoring later.

src/server/handlers/schema.js
bartaz: They are called 'repos' even in the UI, so I'm fine with making it consistent everywhere.

src/common/reducers/repositories.js
earnubs: To be fair, it's unlikely that isSelected is every likely to be clobbered by gh, but a namespace of some kind does tell you that you're dealing with a property sourced from the client rather than a 3rd party source.

Where I was going with this earlier on, and maybe I should have stuck at it, was another object keyed on id, but for client state of entities.

src/common/actions/repositories.js
earnubs: Right, but I think the point I was making was that you have async/await fetch dispatches with which to compose up to this kind of thing, rather than being all wrapped up from the outset.

src/common/components/select-repositories-page/index.js
earnubs: Added as a question in 569

src/server/handlers/github.js
earnubs: I'm thinking based on your last point of rolling this change back.

src/common/actions/repositories.js
earnubs: > "In this case [the] ... error was handled"

What I meant is that perhaps we shouldn't be throwing an error at all:

> The general rule is that a function may deliver operational errors synchronously (e.g., by throwing) or asynchronously (by passing them to a callback or emitting error on an EventEmitter), but it should not do both. 

https://www.joyent.com/node-js/production/design/errors

But since the error is based back via await, it should simply be part of the response?

src/common/components/select-repository-list/index.js
bartaz: @earnubs So if I understand it correctly, right now, with our current "First, Prev, Next, Last" pagination ids may get out of order by using union?

For example user loads first page of repos first (obviously), then uses "Last" to load last page (which is merged by `union` into the result) skipping any pages in between, so moving prev/next to any other pages later means other pages will be merged to the end of ids list while they should be in the middle?

src/server/handlers/github.js
bartaz: The 'CALL_API' middleware nests response in `action.payload.response`, so maybe there will be not so many places where we 'manually' push responses in actions (because we will use middleware for most API calls).

src/common/reducers/entities.js
```diff
@@ -0,0 +1,62 @@
+import merge from 'lodash/merge';
+
+import * as ActionTypes from '../actions/repositories';
+
+// export only for testing
+export function repository(state={
+  __isSelected: false,
```
earnubs: @bartaz further to comments on prefixing (https://github.com/canonical-ols/build.snapcraft.io/pull/541#discussion_r109158954), I was thinking over the weekend that if we whitelisted props from the api in the schema (normalize) we'd have a number of benefits:

- control over properties, no clobbering
- smaller api from server to client
- invariant api, so no chance of performance variation because source adds more data (that we don't use)

and is such, `isSelected` et al no longer need a prefix

src/common/actions/repositories.js
katiefenn: tumblr_mtr8wln9qe1s3b47zo1_500

katiefenn: But seriously, nice analogy.

src/common/actions/repositories.js
katiefenn: Is this not a moot discussion? This seems to be a case of an async action creator that will be replaced by the FETCH_API middleware.

katiefenn: Silly question time: is there any reason we can't encapsulate building multiple repositories behind a single API call? For instance:

```
fetch('/api/launchpad/snaps', { body: JSON.stringify(respositoryUrls) });
```
Dealing with this complexity in the Express service is probably much easier than dealing with it in the client.

src/common/reducers/entities.js
bartaz: Sounds good. But, if API is `snake_case`, so should be `is_selected`. Unless normalize can do camelCasing too for whitelisted props...

earnubs: @bartaz we can I think remap to camelCase if you'd like?

bartaz: The only reason I'm asking about camelCase is that we (used to) have best practice coding style defined as camelCase for JavaScript.
We don't enforce it with eslint either way, but right now it's mixed because of API objects have snake_case.

We could leave it as it is, and just use snake_case when it's returned as such from API, or take a chance and make it consistent if we normalize it anyway.

src/common/actions/repositories.js
earnubs: @katiefenn I think if it were handled in lp it would be great, but handling it in express wouldn't, i don't think, get us much (still has roughly the same network overhead, but now we likely have a less responsive and/or more work to build api) ... i think

src/common/reducers/entities.js
```diff
@@ -0,0 +1,66 @@
...
+      };
+    }
+    case ActionTypes.REPOSITORY_BUILD: {
+      return {
+        ...state,
+        __isFetching: true,
+        __isSelected: true
+      };
+    }
+    case ActionTypes.REPOSITORY_SUCCESS:
```
katiefenn: `REPOSITORY_SUCCESS` and `REPOSITORY_FAILURE` are a bit vague. What succeeded and failed?

Can I suggest changing this to:
`REPOSITORY_BUILD_REQUEST`, `REPOSITORY_BUILD_SUCCESS`, and `REPOSITORY_BUILD_FAILURE`?

src/common/reducers/entities.js
```diff
@@ -0,0 +1,66 @@
...
+import * as ActionTypes from '../actions/repository';
+
+// export only for testing
+export function repository(state={
+  __isSelected: false,
+  __isFetching: false,
+  __error: null
+}, action) {
+
+  switch(action.type) {
```
katiefenn: I don't fully understand why a reducer for the selected repository appears in here and not in a separate module.

earnubs: it's actually more of a helper function for a reducer, my plan is `entities/index.js` `entities/repository.js`, `entities/snap.js`

src/common/selectors/index.js
earnubs: Add a comment in the code for this.

src/common/reducers/entities.js
earnubs: What we're doing here is creating a snap from a given repository, and the request returns a new snap, which we can insert into a refactored snap state and have the correct data on the next view (dashboard).

More accurate name would be:

```
// add a repository to launchpad's build queue, returns a snap representation
const REPOSITORY_ADD = 'REPOSITORY_ADD';
```

src/common/reducers/entities.js
earnubs: Make this split when we move to refactor snaps.

src/common/reducers/entities/index.js
```diff
@@ -0,0 +1,27 @@
...
+export function entities(state = { snaps: {}, repos: {} }, action) {
+  if (action.payload && action.payload.entities) {
+    return merge({}, state, action.payload.entities);
+  }
+
+  // only modify repos if action is one of REPO_ types
+  if (action.payload && action.payload.id && RepoActionTypes[action.type]) {
+    return {
+      ...state,
+      repos: {
```
bartaz: I'm not sure if it will be good to have one reducer (one 'slice' of state) to contain all different entities' types from the app.

One point against it is likely weird logic to detect which part of `entities` should be updated by an action (like we do here by checking for `RepoActionTypes`).

Other one is that any component that needs access to any data (so most of them) would need a connect binding to whole `entities`, even if it's interested only in small part of it (for example repos or snaps only).

But let's not worry about it too much now and see how it goes when we add snaps to the equation.

earnubs: I've flip-flopped in my opinion on this a few times during this work. FWIW I don't agree that the logic here is weird and better use of reducers means components will be looking at this directly less, but I agree that we'll find out more when we start on snaps.

src/server/handlers/launchpad.js
earnubs: see #593

PR #544, opened 2017-03-27
Add instance option for LP credentials

cjwatson:
BSI can be usefully configured locally to point to Launchpad instances
other than production for the purpose of testing various things: staging
mostly works (modulo a snap-proxy problem that'll be fixed soon),
dogfood works, and it's even possible to use a local development
instance.  To make such things easier, add a `--launchpad` option to
`scripts/create-launchpad-credentials`.

Production is still the right choice for BSI staging for the moment,
since Launchpad staging is reset weekly and would require some
reconfiguration to preserve BSI credentials across the reset, but it may
be worth looking into that in future.

### DIFF
--- a/scripts/create-launchpad-credentials
+++ b/scripts/create-launchpad-credentials
@@ -43,7 +43,10 @@ class AuthorizeRequestTokenRemotely(RequestTokenAuthorizationEngine):
 def main():
     parser = ArgumentParser()
     parser.add_argument(
-        "-l", "--launchpad", dest="launchpad_instance", default="production")
+        "-l", "--launchpad", dest="launchpad_instance", default="production",
+        help=(
+            "Launchpad instance (production, staging, qastaging, dogfood, "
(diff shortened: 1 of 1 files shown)


scripts/create-launchpad-credentials
```diff
@@ -39,19 +41,25 @@ class AuthorizeRequestTokenRemotely(RequestTokenAuthorizationEngine):
 
 
 def main():
+    parser = ArgumentParser()
+    parser.add_argument(
+        "-l", "--launchpad", dest="launchpad_instance", default="production")
```
bartaz: What are the values it accepts?

I see "production" as default, so I also assume "staging", but for local instances it does it by IP?

cjwatson: I've added a bit of documentation.  The process of setting up Launchpad locally typically involves adding `launchpad.dev` etc. to `/etc/hosts`, but a full URI works too if you're doing something even stranger.

PR #546, opened 2017-03-27
Beta notification

### LINKED ISSUE #534: [beta] banner asking for feedback [1]

bartaz:
Fixes #534

screen shot 2017-03-27 at 17 02 40

This adds a BetaNotification component that renders notification (if redux state is properly set to true).
The other component BetaNotificationTrigger is used in app container to setup a delay of 2 minutes to show the notification.


src/common/components/beta-notification/index.js
```diff
@@ -0,0 +1,55 @@
+import React, { Component, PropTypes } from 'react';
+import { connect } from 'react-redux';
+import localforage from 'localforage';
+
+import { BETA_NOTIFICATION_TOGGLE } from '../../reducers/beta-notification';
```
bartaz: Action type imported directly from reducer. This is such a simple (and temporary) feature, that there is no need to add action creators boilerplate to it.

src/common/components/beta-notification/index.js
```diff
@@ -0,0 +1,55 @@
...
+    event.preventDefault();
+
+    this.props.dismissNotification();
+    this.storeNotificationDismissed();
+  }
+
+  render() {
+    return this.props.isVisible
+      ? (
+        <Notification onRemoveClick={this.onRemoveClick.bind(this)}>
```
bartaz: Simply renders our Vanilla notification.

Component is added to top of the page container - in document flow - meaning it will push down the content when it appears.

src/common/components/vanilla/notification/index.js
```diff
@@ -13,25 +13,14 @@ export default class Notification extends Component {
           { status && this.getStatus(this.props.status) }
           { message }
           { this.props.children }
-          { action && actionText &&
```
bartaz: Notification reused from what we have in MU.
But we don't use 'actions' anywhere here. Instead we have new icon X button with remove callback.

test/unit/src/common/components/beta-notification/t_beta-notification.js
```diff
@@ -0,0 +1,70 @@
...
+      wrapper = shallow(<BetaNotificationTriggerView authenticated={true} showNotification={callback}/>);
+
+      // componentDidMount is async because of internal localforage usage
+      // and because mocked localforage uses setTimeout to call async callbacks
+      // we need to first store reference to componentDidMount promise:
+      const componentDidMount = wrapper.instance().componentDidMount();
+      // and then tick the mocked clock to make localforage resolve properly
+      clock.tick(1);
+      // and finally we can wait for componentDidMount to fully resolve
+      await componentDidMount;
```
bartaz: This was a much harder beast to test than I thought it would.

We are using mocked setTimeout (from `clock`), but internally used `localforage` stub also uses `setTimeout` to call its callbacks asynchronously. So localforage promises don't resolve until we tick the clock. And only then we can wait till async componentDidMount resolves...

src/common/components/beta-notification/trigger.js
```diff
@@ -0,0 +1,48 @@
...
+const BETA_NOTIFICATION_DELAY = 2 * 6 * 1000; // 2 minutes
+
+export class BetaNotificationTriggerView extends Component {
+  notificationTimeout = null;
+
+  async componentDidMount() {
+    const notificationDismissed = await localforage.getItem(BETA_NOTIFICATION_DISMISSED_KEY);
+
+    if (this.props.authenticated && !notificationDismissed) {
+      this.notificationTimeout = setTimeout(this.props.showNotification, BETA_NOTIFICATION_DELAY);
```
earnubs: This dismisses the notification after 2 minutes if the user hasn't? 
2 minutes is a long time, am I correct is saying a user could potentially be further down the page and then the banner dismisses, in flow, the UI reflows and the user is wondering why?
Even if it didn't do that, what's the gain from timeout of this banner? It seems like work for no gain.

earnubs: OK, so it's the reverse!

PR #553, opened 2017-03-28
495: CSRF tokens

katiefenn:
- Added new Express middleware to generate and verify CSRF tokens
- Added global window.__CSRF_TOKEN__ variable for application to consume
  token from server
- Added new API_CALL redux middleware to encapsulate API requests and pass tokens
- Migrated FETCH_SNAPS and REMOVE_SNAPS chain of events to API_CALL
- Fixed snaps reducer to work with API_CALL middleware
- Added unit tests

### DIFF
--- a/src/common/actions/snaps.js
+++ b/src/common/actions/snaps.js
@@ -1,9 +1,6 @@
 import 'isomorphic-fetch';
 
-import { checkStatus, getError } from '../helpers/api';
-import { conf } from '../helpers/config';
-
-const BASE_URL = conf.get('BASE_URL');
+import { CALL_API } from '../middleware/call-api';
 
 export const FETCH_SNAPS = 'FETCH_SNAP_REPOSITORIES';
 export const FETCH_SNAPS_SUCCESS = 'FETCH_SNAPS_SUCCESS';
@@ -14,42 +11,19 @@ export const REMOVE_SNAP_ERROR = 'REMOVE_SNAP_ERROR';
 
 export function fetchUserSnaps(owner) {
   return async (dispatch) => {
-    const url = `${BASE_URL}/api/launchpad/snaps/list`;
     const query = `owner=${encodeURIComponent(owner)}`;
 
     dispatch({
-      type: FETCH_SNAPS
-    });
-
-    try {
-      const response = await fetch(`${url}?${query}`, {
-        headers: { 'Content-Type': 'application/json' },
-        credentials: 'same-origin'
-      });
-      await checkStatus(response);
-      const result = await response.json();
-      if (result.status !== 'success') {
-        throw getError(response, result);
+      type: FETCH_SNAPS,
+      [CALL_API]: {
+        types: [ FETCH_SNAPS_SUCCESS, FETCH_SNAPS_ERROR ],
+        path: `/api/launchpad/snaps/list?${query}`,
+        options: {
+          headers: { 'Content-Type': 'application/json' },
+          credentials: 'same-origin'
+        }
       }
-      dispatch(fetchSnapsSuccess(result.payload.snaps));
-    } catch (error) {
-      dispatch(fetchSnapsError(error));
-    }
-  };
-}
-
-export function fetchSnapsSuccess(snaps) {
-  return {
-    type: FETCH_SNAPS_SUCCESS,
-    payload: snaps
-  };
-}
-
-export function fetchSnapsError(error) {
-  return {
-    type: FETCH_SNAPS_ERROR,
-    payload: error,
-    error: true
+    });
   };
 }
 
@@ -57,42 +31,17 @@ export function removeSnap(repositoryUrl) {
   return async (dispatch) => {
     dispatch({
       type: REMOVE_SNAP,
-      payload: { repository_url: repositoryUrl }
-    });
-
-    try {
-      const response = await fetch(`${BASE_URL}/api/launchpad/snaps/delete`, {
-        method: 'POST',
-        headers: { 'Content-Type': 'application/json' },
-        body: JSON.stringify({ repository_url: repositoryUrl }),
-        credentials: 'same-origin'
-      });
-      await checkStatus(response);
-      const result = await response.json();
-      if (result.status !== 'success') {
-        throw getError(response, result);
+      payload: { repository_url: repositoryUrl },
+      [ CALL_API ]: {
+        types: [ REMOVE_SNAP_SUCCESS, REMOVE_SNAP_ERROR ],
+        path: '/api/launchpad/snaps/delete',
+        options: {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({ repository_url: repositoryUrl }),
+          credentials: 'same-origin'
+        }
       }
-      dispatch(removeSnapSuccess(repositoryUrl));
-    } catch (error) {
-      dispatch(removeSnapError(repositoryUrl, error));
-    }
-  };
-}
-
-export function removeSnapSuccess(repositoryUrl) {
-  return {
-    type: REMOVE_SNAP_SUCCESS,
-    payload: { repository_url: repositoryUrl }
-  };
-}
-
-export function removeSnapError(repositoryUrl, error) {
-  return {
-    type: REMOVE_SNAP_ERROR,
-    payload: {
-      repository_url: repositoryUrl,
-      error
-    },
-    error: true
+    });
   };
 }
--- a/src/common/actions/snaps.js
+++ b/src/common/actions/snaps.js
@@ -10,38 +10,32 @@ export const REMOVE_SNAP_SUCCESS = 'REMOVE_SNAP_SUCCESS';
 export const REMOVE_SNAP_ERROR = 'REMOVE_SNAP_ERROR';
 
 export function fetchUserSnaps(owner) {
-  return async (dispatch) => {
-    const query = `owner=${encodeURIComponent(owner)}`;
+  const query = `owner=${encodeURIComponent(owner)}`;
 
-    dispatch({
-      type: FETCH_SNAPS,
-      [CALL_API]: {
-        types: [ FETCH_SNAPS_SUCCESS, FETCH_SNAPS_ERROR ],
-        path: `/api/launchpad/snaps/list?${query}`,
-        options: {
-          headers: { 'Content-Type': 'application/json' },
-          credentials: 'same-origin'
-        }
+  return {
+    [CALL_API]: {
+      types: [ FETCH_SNAPS, FETCH_SNAPS_SUCCESS, FETCH_SNAPS_ERROR ],
+      path: `/api/launchpad/snaps/list?${query}`,
+      options: {
+        headers: { 'Content-Type': 'application/json' },
+        credentials: 'same-origin'
       }
-    });
+    }
   };
 }
 
 export function removeSnap(repositoryUrl) {
-  return async (dispatch) => {
-    dispatch({
-      type: REMOVE_SNAP,
-      payload: { repository_url: repositoryUrl },
-      [ CALL_API ]: {
-        types: [ REMOVE_SNAP_SUCCESS, REMOVE_SNAP_ERROR ],
-        path: '/api/launchpad/snaps/delete',
-        options: {
-          method: 'POST',
-          headers: { 'Content-Type': 'application/json' },
-          body: JSON.stringify({ repository_url: repositoryUrl }),
-          credentials: 'same-origin'
-        }
+  return {
+    payload: { repository_url: repositoryUrl },
+    [ CALL_API ]: {
+      types: [ REMOVE_SNAP, REMOVE_SNAP_SUCCESS, REMOVE_SNAP_ERROR ],
+      path: '/api/launchpad/snaps/delete',
+      options: {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ repository_url: repositoryUrl }),
+        credentials: 'same-origin'
       }
-    });
+    }
   };
 }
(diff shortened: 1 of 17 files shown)


src/common/actions/snaps.js
```diff
@@ -14,85 +11,37 @@ export const REMOVE_SNAP_ERROR = 'REMOVE_SNAP_ERROR';
...
-        headers: { 'Content-Type': 'application/json' },
-        credentials: 'same-origin'
-      });
-      await checkStatus(response);
-      const result = await response.json();
-      if (result.status !== 'success') {
-        throw getError(response, result);
+      type: FETCH_SNAPS,
+      [CALL_API]: {
+        types: [ FETCH_SNAPS_SUCCESS, FETCH_SNAPS_ERROR ],
```
katiefenn: The main difference between my implementation and the [Redux real world example](https://github.com/reactjs/redux/blob/master/examples/real-world/src/actions/index.js) is that my example expects two action types, not three.

The reason for this is that Redux requires actions to always have a type:
> Actions must have a type property that indicates the type of action being performed.
([Redux docs: actions](http://redux.js.org/docs/basics/Actions.html)

Some Redux tools break if actions don't have a type (redux-mock-store). My solution is to use the action dispatched by the action creator to dispatch the "request type" action, instead of it [being passed as an action type](https://github.com/reactjs/redux/blob/master/examples/real-world/src/actions/index.js#L11).

I think this makes sense, because the role of middleware should be to augment other actions, not to be used in lieu of actions.

src/common/middleware/call-api.js
```diff
@@ -0,0 +1,101 @@
+import { checkStatus, getError } from '../helpers/api';
+
+// Action key that carries API call info interpreted by this Redux middleware.
+export const CALL_API = 'CALL_API';
+
+// A Redux middleware that interprets actions with CALL_API info specified.
+// Performs the call and promises when such actions are dispatched.
+export default defaults => () => next => async action => {
```
katiefenn: A small variation on the Redux real-world example is an extra function around the middleware to memoise default settings, such as the API basepath (`defaults.endpoint`).

src/server/middleware/csrf-token.js
```diff
@@ -0,0 +1,67 @@
...
+  if (typeof req.session.csrfTokens === 'undefined') {
+    req.session.csrfTokens = [];
+  }
+
+  req.session.csrfTokens.push(csrfToken);
+
+  // Only 5 valid CSRF tokens at a time. Safe to increase number.
+  while (req.session.csrfTokens.length > 5) {
+    req.session.csrfTokens.shift();
+  }
```
katiefenn: We keep the last 5 generated CSRF tokens for each user. A new token is generated on every page view that renders the application. This should be enough for the user to use the site in more than one browser tab.

Interested in some debate about whether this is five tokens is enough to keep.

src/server/middleware/csrf-token.js
```diff
@@ -0,0 +1,67 @@
+import uuid from 'node-uuid';
+
+export const generateToken = (req, res, next) => {
+  // Don't generate new tokens for API calls
+  if (req.path.match(/^\/api/)) {
+    return next();
+  }
+
+  // TODO: Remove debug
```
katiefenn: >_<

src/common/middleware/call-api.js
```diff
@@ -0,0 +1,101 @@
...
+    await checkStatus(response);
+    const result = await response.json();
+
+    if (result.status !== 'success') {
+      throw getError(response, result);
+    }
+
+    return next(createAction({
+      type: successType,
+      response: result.payload
```
bartaz: I think we should keep using Flux Standard Action format, so the response should be sent in action `payload` property.

src/common/middleware/call-api.js
```diff
@@ -0,0 +1,101 @@
...
+    }));
+  }
+  catch(error) {
+    return next(createAction({
+      type: failureType,
+      error: error
```
bartaz: Similar comment to above, in Flux Standard Action error should be sent in `payload` and `error: true` added to action object.

src/server/middleware/csrf-token.js
bartaz: Some unnecessary TODO comment?

src/common/actions/snaps.js
bartaz: If I read this correctly, we are not doing the fetch in the action creator anymore, but we dispatch an action with API call meta data that is ignored by reducers, but middleware uses it to make the call and dispatch another actions based on the results?

That's quite a confusing abstraction layer when I first look at it. And I wonder what benefits does it have over some helper/wrapper around fetch that would add common headers and token to the request.

As far as I can see also there is no way to do anything more with API result other then dispatching an action with it?

test/helpers.js
```diff
@@ -31,6 +33,18 @@ expect.extend({
...
+    expect.assert(
+      this.actual.filter(predicate).length > 0,
+      `Expected dispatched actions to have action matching supplied predicate.${EOL}
+       Actual: ${inspect(this.actual, { depth: 3 })}`
+    );
+
+    return this;
+  }
+});
+
```
katiefenn: Added new matcher to compare actions using a comparator function to tidy up some unit tests.

src/common/actions/snaps.js
katiefenn: It's the [same abstraction](https://github.com/reactjs/redux/blob/master/examples/real-world/src/actions/index.js) we discussed in IRC last week, so I'm not sure why it's confusing.

The principle benefits I see are that we can memoise settings such as the API base URL and the CSRF token when the middleware is added to Redux. A helper function would have to accept the settings by another route:

- By parameter, on every call.
- From a global.
- From a singleton pattern.

These trade-offs are undesirable, in my opinion. I think the middleware is a great option that follows an example set by the Redux team.

earnubs: @katiefenn can you explain a little more why the request action isn't included here?

katiefenn: @earnubs The request type is defined on the action instead of in [CALL_API]. If the action doesn't have a type, it breaks some Redux tools like redux-mock-store. Not a lot has changed here, only how the "request" action type is defined.

src/common/middleware/call-api.js
katiefenn: This follows the [Redux real-world example](https://github.com/reactjs/redux/blob/master/examples/real-world/src/middleware/api.js#L117). The reason it does this is not to conflict with the payload property from [the action it extends](https://github.com/reactjs/redux/blob/master/examples/real-world/src/middleware/api.js#L107).

I'm using this to allow the payload from the base action to be passed with the success and error actions too. This is important for actions such as `REMOVE_SNAP_SUCCESS`, because it needs the repo URL from the payload to remove the snap from the store.

src/common/middleware/call-api.js
katiefenn: We need to choose which standard we are going to implement here. Redux Standard Actions defines one standard which mandates:

> An action MUST NOT include properties other than type, payload, error, and meta.

...where `error` must be a bool. The React real-world example and thunks use a different standard. Do we need to use FSA everywhere?

bartaz: Yep, it's quite unfortunate that with such a freedom redux gives it doesn't really enforce any conventions, so every example use it in different way.

I thought it was nice for us to try to use FSA as much as we could as they give us this convention that redux doesn't on its own.
Unfortunately - the redux 'real world' example - doesn't use it, and if we want to follow its patters we have to leave FSA behind.

That being said, I'm not sure if we need to tread redux real-world example as an official standard documentation. It's just an example of possible implementation. Copying it one by one doesn't make it better then anything else. Doesn't make it more 'redux standard'.
So maybe it would be possible to make our CALL_API middleware compatible with FSA in some way?

src/server/middleware/csrf-token.js
cjwatson: The other thing you could consider is generating CSRF tokens using HMAC based on a server-side secret and the session ID.  That way the server doesn't need to store the tokens, and they're still unforgeable as long as the secret stays secret and the underlying hash algorithm is sound.  You can include additional criteria like an expiration timestamp if you like.  It's the same kind of idea as macaroons.

Web-searching for "CSRF HMAC" finds a few things about this, e.g. [hmac-csrf](https://www.npmjs.com/package/hmac-csrf), though I haven't looked them over in any detail.

src/common/middleware/call-api.js
katiefenn: @bartaz: How about changing CALL_API to dispatch an action like this?

```
{
  type: 'ACTION_TYPE',
  payload: {
    callApiPayload: { /* response or error from callApi */ },
    actionPayload: { /* payload from original action */ }
  }
}
```

katiefenn: In my opinion, the standard defined by FSA takes precedence over the standard defined by the Redux real world example.

bartaz: @katiefenn that could work, but actual payload will get quite ugly nested.

Would it work as something like this?
```
{
  type: 'ACTION_TYPE',
  payload: {
    ...actionPayload: { /* payload from original action */ }
    ...callApiPayload: { /* response or error from callApi just added to original payload */ },
  }
}
```

Of course there would be a risk of callApi payload overriding the action payload - but it overrides action type anyway when calling SUCCESS/ERROR.

The other way maybe:
```
{
  type: 'ACTION_TYPE',
  payload: {
    ...actionPayload: { /* payload from original action */ }
    response: callApiPayload { /* response or error from callApi nested */ },
  }
}
```

So that only response object would be nested? And this would be API response object (whatever is returned by API) in case of success or error object in case of error?

These are just quick proposals, I'm not sure exactly how it would match specific bits of callApi middleware.

src/server/middleware/csrf-token.js
katiefenn: @cjwatson: I think this is a good idea, but would be better introduced by a feature further down the road.

src/common/middleware/call-api.js
katiefenn: I don't see why nesting both the actionPayload and the callApiPayload would be ugly, but I would open to nesting only the callApiPayload as in your second example.

src/common/middleware/call-api.js
```diff
@@ -0,0 +1,105 @@
...
+    const result = await response.json();
+
+    if (result.status !== 'success') {
+      throw getError(response, result);
+    }
+
+    return next(createAction({
+      type: successType,
+      payload: {
+        response: result.payload
```
katiefenn: `response` is now nested in the action payload.

src/common/middleware/call-api.js
```diff
@@ -0,0 +1,105 @@
...
+      }
+    }));
+  }
+  catch(error) {
+    return next(createAction({
+      type: failureType,
+      payload: {
+        error: error
```
katiefenn: `error` is now nested in the action payload.

src/common/middleware/call-api.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+};
+
+// Creates new action with request, success or
+// failure types, corresponding with the stages
+// of the request cycle. Bundles properties set
+// on the original action.
+function createActionWith(action, data) {
+  const finalPayload = Object.assign({}, action.payload, data.payload);
+  const finalAction = Object.assign({}, action, data, { payload: finalPayload });
```
katiefenn: Merge action payload with middleware payload, then merge action with payload action.

src/server/server.js
```diff
@@ -57,6 +58,7 @@ app.use(metricsBundle);
 // routes
 app.use('/metrics', trustedNetworks, metricsBundle.metricsMiddleware);
 app.use('/', routes.login);
+app.use(generateToken);
```
earnubs: Should this not be before login route?

src/server/server.js
```diff
@@ -56,12 +57,12 @@ app.use(metricsBundle);
 
 // routes
 app.use('/metrics', trustedNetworks, metricsBundle.metricsMiddleware);
+app.use(generateToken);
 app.use('/', routes.login);
 app.use('/api', routes.github);
 app.use('/api', routes.launchpad);
 app.use('/api', routes.store);
 app.use('/', routes.webhook);
-app.use(routes.login);
```
katiefenn: Removed duplicate route.

src/server/server.js
```diff
@@ -56,12 +57,12 @@ app.use(metricsBundle);
 
 // routes
 app.use('/metrics', trustedNetworks, metricsBundle.metricsMiddleware);
+app.use(generateToken);
 app.use('/', routes.login);
 app.use('/api', routes.github);
 app.use('/api', routes.launchpad);
 app.use('/api', routes.store);
 app.use('/', routes.webhook);
```
cjwatson: Maybe move this above `generateToken`?  We don't need or want to generate a new CSRF token for webhook notifications.

src/server/server.js
```diff
@@ -56,12 +57,12 @@ app.use(metricsBundle);
...
-app.use(routes.login);
 app.use(routes.githubAuth);
```
cjwatson: It might also be worth considering where this goes in the route order, as with `routes.webhook` above.

src/server/routes/launchpad.js
```diff
@@ -20,6 +21,7 @@ router.get('/launchpad/snaps', findSnap);
 router.use('/launchpad/snaps/list', json());
 router.get('/launchpad/snaps/list', findSnaps);
```
cjwatson: Shouldn't you add `verifyToken` here too?

src/common/middleware/call-api.js
bartaz: Personally I find such nested function one-liners a bit hard to read especially when params are not wrapped in parens. I really don't know where one function starts and where other ends.

I know that using `return` for each single nested function would be a bit of boilerplate and indent, so it's not perfect too.

I'm not strongly against current state, but maybe it could be made more readable.

katiefenn: For the record, I dislike this too. The only reason I went with it is because it's a common pattern in middleware.

PR #566, opened 2017-03-30
Add metric for builds released to edge

### LINKED ISSUE #517: Add metric for developers with snaps pushed to edge [2]
Track the number of developers with snaps pushed to edge for the KPI dashboard.

I've split this out from #446 because it involves a somewhat different kind of work from the other metrics.  Pushing to edge is done by Launchpad, and the only way BSI can currently find out about that is when fetching build statuses, which I don't think would be a very good point at which to update this metric.

The cleanest approach I've thought of is to make a small extension to Launchpad's snap:build webhook so that it also pushes a webhook event when the store upload status changes.  BSI can then subscribe to that webhook and use it to update a metric.

cjwatson:
We do this by configuring a webhook on Launchpad snaps and counting
events where `store_upload_status` is "Uploaded".  (We could use the
information provided by this webhook for some more things in future to
reduce polling, although that would also require putting some more
thought into ordering guarantees; since "Uploaded" is a terminal status
we don't need to think about that too hard in this case.)

Of course there are lots of existing snaps that need to have the webhook
added, and it's also possible that we might create the snap and then
fail to create the webhook.  We don't currently have any infrastructure
for running scripts or periodic jobs, which would be a better way to
handle this; but as a reasonably lightweight workaround for these
problems, `/api/launchpad/snaps/list` also ensures that the necessary
webhook exists for each of the snaps it returns, and we make a note in
memcached that we've done so to avoid hitting the Launchpad API too
often.

Fixes #517.

### DIFF
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -1,15 +1,16 @@
 import { createHash } from 'crypto';
 
 import yaml from 'js-yaml';
-import parseGitHubUrl from 'parse-github-url';
 
+import { parseGitHubRepoUrl } from '../../common/helpers/github-url';
 import db from '../db';
 import { conf } from '../helpers/config';
 import { getMemcached } from '../helpers/memcached';
 import requestGitHub from '../helpers/github';
 import getLaunchpad from '../launchpad';
-import { getSnapcraftData } from './github';
 import logging from '../logging';
+import { getSnapcraftData } from './github';
+import { getLaunchpadRootSecret, makeWebhookSecret } from './webhook';
 
 const logger = logging.getLogger('express');
 
@@ -99,6 +100,7 @@ export const getRepoUrlPrefix = (owner) => `https://github.com/${owner}/`;
 // memcached cache id helpers
 export const getUrlPrefixCacheId = (urlPrefix) => `url_prefix:${urlPrefix}`;
 export const getRepositoryUrlCacheId = (repositoryUrl) => `url:${repositoryUrl}`;
+export const getHasWebhookCacheId = (snapUrl) => `has_webhook:${snapUrl}`;
 
 // Wrap errors in a promise chain so that they always end up as a
 // PreparedError.
@@ -167,8 +169,8 @@ const checkAdminPermissions = async (session, repositoryUrl) => {
   }
   const token = session.token;
 
-  const parsed = parseGitHubUrl(repositoryUrl);
-  if (parsed === null || parsed.owner === null || parsed.name === null) {
+  const parsed = parseGitHubRepoUrl(repositoryUrl);
+  if (parsed === null) {
     logger.info(`Cannot parse "${repositoryUrl}"`);
     throw new PreparedError(400, RESPONSE_GITHUB_BAD_URL);
   }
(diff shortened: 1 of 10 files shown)


src/server/handlers/webhook.js
```diff
@@ -83,5 +115,23 @@ export const notify = async (req, res) => {
       logger.error(`Failed to request builds of ${repositoryUrl}: ${error}.`);
       return res.status(500).send();
     }
+  } else if (req.headers['x-launchpad-event-type'] === 'snap:build:0.1') {
```
bartaz: Maybe it would be clearer to have separate function to handle github and launchpad webhooks.

It's not terrible now, as there are only 2 of them. But if we get to the point where this `if else` grows even more it will be quite hard to comprehend.

src/server/handlers/launchpad.js
```diff
@@ -99,6 +100,7 @@ export const getRepoUrlPrefix = (owner) => `https://github.com/${owner}/`;
 // memcached cache id helpers
 export const getUrlPrefixCacheId = (urlPrefix) => `url_prefix:${urlPrefix}`;
 export const getRepositoryUrlCacheId = (repositoryUrl) => `url:${repositoryUrl}`;
+export const getHasWebhookCacheId = (snapUrl) => `has_webhook:${snapUrl}`;
```
bartaz: Please add to memcached.md in docs.

src/server/handlers/webhook.js
bartaz: What does `0.1` in `snap:build:0.1` stand for? Is this some kind of webhook event type versioning?
If it changes on LP side we want to stop recognizing the event and need to reimplement?

test/routes/src/server/routes/t_webhook.js
```diff
@@ -393,5 +404,148 @@ describe('The WebHook API endpoint', () => {
...
+        await db.model('GitHubUser')
+          .forge({
+            github_id: 1,
+            login: 'anowner',
+            last_login_at: new Date()
+          })
+          .save();
+      });
+
+      describe('if store_upload_status is not Uploaded', () => {
```
bartaz: As a convention we tend to be using `describe` only on top level (when we say we 'describe' a 'thing' that is being tested) and for different variants/scenarios we use `context` (when we say how a thing should work in specific case).

src/server/handlers/webhook.js
cjwatson: Refactored, thanks.

As for `snap:build:0.1`, yes, it's a versioning thing: see [LP's webhooks documentation](https://help.launchpad.net/API/Webhooks).  If that part of the event name changes then by definition it's a backward-incompatible change and we'll need to make some changes on our side, but since we subscribe specifically to `snap:build:0.1` we shouldn't be caught by surprise.

src/server/handlers/launchpad.js
cjwatson: Done, thanks.

test/routes/src/server/routes/t_webhook.js
cjwatson: Yes, this was a typo (well, two typoes).  Corrected.

PR #577, opened 2017-04-04
Replace node uuid

katiefenn:


package.json
```diff
@@ -77,6 +76,7 @@
     "reselect": "^2.5.4",
     "sanitize-html": "^1.14.1",
     "serve-favicon": "^2.4.0",
+    "uuid": "^3.0.0",
```
cjwatson: This needs a corresponding shrinkwrap change.

cjwatson: Never mind, my mistake - I was misled by GitHub hiding that bit of the diff.

PR #580, opened 2017-04-04
Updated action names for adding repo

bartaz:
Reimplements #571 

Updating add repository action names. 
Took a chance and did some repository/repo renaming on the way.

Part of #541

### DIFF
--- a/src/common/actions/repository.js
+++ b/src/common/actions/repository.js
@@ -36,7 +36,7 @@ export function addRepos(repositories, owner) {
   };
 }
 
-// add a repository to launchpad's build queue, returns a snap representation
+// add a repository to launchpad's build queue
 export function addRepo(repository) {
   const { id, url, owner, name } = repository;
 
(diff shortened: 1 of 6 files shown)


src/common/actions/repository.js
```diff
@@ -36,12 +36,13 @@ export function buildRepositories(repositories, owner) {
   };
 }
 
-export function buildRepository(repository) {
+// add a repository to launchpad's build queue, returns a snap representation
```
earnubs: the fetch might, but we're not handling it here are we? it's just updating the original repository state i think

bartaz: yep, ok - let's not make the comment more confusing that it needs to be :)

test/unit/src/common/actions/t_repositories.js
```diff
@@ -132,7 +132,7 @@ describe('repositories actions', () => {
       it('should store no pageLinks', async () => {
         await store.dispatch(fetchUserRepositories());
         expect(store.getActions()).notToHaveActionOfType(
-          ActionTypes.SET_REPOSITORY_PAGE_LINKS
+          ActionTypes.SET_REPO_PAGE_LINKS
```
earnubs: not sure this exists anymore, but the test hasn't been removed or updated so it's a valid change here

bartaz: It probably doesn't exist, because I think I only seen it in this one place in test when renaming.

PR #581, opened 2017-04-04
Better select-repository-list component test coverage

earnubs:


src/common/components/select-repository-list/index.js
```diff
@@ -102,37 +102,32 @@ export class SelectRepositoryListComponent extends Component {
     this.props.dispatch(fetchUserRepositories(pageNumber));
   }
 
-  pageSlice(array, pageLinks) {
+  pageSlice(repositoriesIndex, pageLinks) {
```
bartaz: This method seems quite independent from SelectRepositoryList (it doesn't seem to have any reference to `this`), so maybe it could actually be external helper function?

Would that help in any way? Make it clearer? Easier to test? - if not, never mind ;)

test/unit/src/common/components/select-repository-list/t_select-repository-list.js
```diff
@@ -17,41 +25,207 @@ describe('<SelectRepositoryListComponent /> instance', function() {
...
+    it('should render disabled Add button', function() {
       expect(wrapper.find('Button').prop('disabled')).toBe(true);
     });
 
     it('should show message about 0 selected repos', function() {
       expect(wrapper.html()).toInclude('0 selected');
     });
+
+    it('should render same number of rows as repos in state', function() {
+      expect(wrapper.find(SelectRepositoryRow).length).toBe(3);
```
bartaz: Wouldn't `props.repositories.ids.length` instead of `3` be more descriptive?

src/common/components/select-repository-list/index.js
earnubs: For me the problems with this method lies in the state it has to deal with, pageLinks needs some work I think, but it could absolutely be a helper or static, but I would do it later as part of dealing with pageLinks I think.

PR #588, opened 2017-04-05
fix select-repository-row tests

earnubs:
select-repository-row has the repository as a prop and can figure out if it should be selected or not from that, it can't know if it's enabled, so that has to be passed from the parent where the comparison of snap and repo list (selector) works that out.

Use simulate('change') to trigger the onChange handler.

FWIW I find the verbose naming of these tests a little hard to parse on a test by test basis.

### DIFF
--- a/test/unit/src/common/components/select-repository-row/t_select-repository-row.js
+++ b/test/unit/src/common/components/select-repository-row/t_select-repository-row.js
@@ -47,20 +47,20 @@ describe('The SelectRepositoryRow component', () => {
 
         beforeEach(() => {
           component = shallow(<SelectRepositoryRow { ...props } />);
-          component.instance().onChange();
         });
 
-        xit('should call the onChange callback', () => {
+        it('should call the onChange callback', () => {
+          component.find('input').simulate('change');
           expect(onChange).toHaveBeenCalled();
         });
       });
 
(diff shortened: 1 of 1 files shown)


test/unit/src/common/components/select-repository-row/t_select-repository-row.js
```diff
@@ -79,10 +79,21 @@ describe('The SelectRepositoryRow component', () => {
...
         });
 
-        xit('should contain a disabled checkbox', () => {
+        it('should contain a disabled checkbox', () => {
+          const component = shallow(<SelectRepositoryRow { ...props } />);
+          expect(component.find('input[disabled=true]').length).toBe(1);
+        });
+      });
+
+      context('and the reposistory is fetching', () => {
```
bartaz: Agreed about a bit confusing naming.

Tests (and contexts) should be independent - therefore quite understandable on case by case basis.
Contexts that start with `and ...` suggest dependancy on something that happened before that. If there is a dependency on previous tests - then the tests are not well implemented, if there is no dependency, we shouldn't suggest it in naming.

bartaz: But anyway - let's not worry about it now.

PR #596, opened 2017-04-06
Show store-assigned revisions of builds

cjwatson:
Launchpad exports this information now, so we can restore the
`--revision=` part of `snap install` instructions.

### DIFF
--- a/src/common/containers/build-details.js
+++ b/src/common/containers/build-details.js
@@ -38,12 +38,13 @@ class BuildDetails extends Component {
           snapcraft cleanbuild --debug
         </HelpInstallSnap>
       );
-    } else if (snap && snap.store_name) {
-      // otherwise if we have snap name show install instructions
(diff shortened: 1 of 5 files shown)


src/common/containers/build-details.js
```diff
@@ -38,12 +38,13 @@ class BuildDetails extends Component {
           snapcraft cleanbuild --debug
         </HelpInstallSnap>
       );
-    } else if (snap && snap.store_name) {
-      // otherwise if we have snap name show install instructions
+    } else if (snap && snap.store_name && build.storeRevision) {
```
bartaz: We don't need to support the case where we have snap name but not the revision?
In such case it means build failed (at least failed to publish), so it cannot be installed?

cjwatson: Right, in that case it's not possible to show install instructions.  (It might not be a failure - perhaps it's still in progress.)

PR #597, opened 2017-04-06
More test coverage for repository actions and reducers

earnubs:


src/common/reducers/entities/index.js
```diff
@@ -1,26 +1,22 @@
...
-  if (action.payload && action.payload.id && RepoActionTypes[action.type]) {
-    return {
-      ...state,
-      repos: {
-        ...state.repos,
-        [action.payload.id]: repository(state.repos[action.payload.id], action)
-      }
-    };
+  if (RepoActionTypes[action.type]) {
+    return repository(state, action);
```
bartaz: I don't like that `repository` actually returns whole entities object.

Maybe renaming it to `repositoryEntities` would make it more clear, but still, it's a bit confusing.

I think repository helper/reducer shouldn't know about entities structure, so maybe something like:

```js
return {
   ...state,
   repos: reposEntities(state.repos, action)
}
```

This way `reposEntities` doesn't need to know about structure of `entities`, only about repos object inside.
And probably it may help us move away from checking action types in `entities` at all, as it's `reposEntities` that may check it and decide weather repos need to be touched at all.

earnubs: 11:48 AM <beowulf> bartaz: ok, i see your point now
11:48 AM <beowulf> bartaz: so ... right, repository as it currently is should be back in entities ... yeah, i get you
11:49 AM <beowulf> and _repository is what stays (because it deals with a repo, not an entity collection)
11:49 AM <beowulf> bartaz: have i understood that correctly?
11:50 AM <bartaz> yes - and actually if 'reposEntities' (or howerer we call it - 'repos' maybe) would be implemented in a way that don't knows about entities, I'd be ok with having it in separate file
11:51 AM <bartaz> like 'repos.js' that would have repos and repository
11:51 AM <beowulf> bartaz: in a future where snaps are also involved, the repository func (that returns and entity collect) would be a helper function that you would pass 'snap' or 'repo' too
11:51 AM <bartaz> right, the only difference would be the prop name
11:51 AM <beowulf> bartaz: i think it has to know about entities, no? it's really just the mapping ... yeah
11:52 AM <beowulf> so it should be in entities, and then the naming isn't so much of a problem either
11:52 AM <bartaz> yes
11:52 AM <beowulf> it really has nothign to do with repos
11:52 AM <beowulf> it's just a path finder

src/common/reducers/entities/index.js
```diff
@@ -1,27 +1,43 @@
...
+  }
+
+  return state;
+}
+
+function reduceRepoEntity(state, action) {
+  return reduceEntityHelper(state, action, repository, 'repos');
+}
+
+function reduceEntityHelper(state, action, reducer, name) {
```
earnubs: This could maybe be made more magical if name were also the key to an object containing reducers.

test/unit/src/common/reducers/entities/fixtures.js
```diff
@@ -0,0 +1,64 @@
...
+};
+
+export const repoSuccessState = {
+  isSelected: false,
+  isFetching: false,
+  error: null
+};
+
+export const repoFailureState = {
+  isSelected: true,
```
earnubs: This should arguably be false, allowing the user to submit the remaining repos without further interaction.

src/common/reducers/entities/index.js
```diff
@@ -1,27 +1,43 @@
...
+  repos: {},
+  owners: {}
+}, action) {
   if (action.payload && action.payload.entities) {
     return merge({}, state, action.payload.entities);
   }
 
   // only modify repos if action is one of REPO_ types
-  if (action.payload && action.payload.id && RepoActionTypes[action.type]) {
+  if (RepoActionTypes[action.type]) {
```
bartaz: Maybe we could now move `if (RepoActionTypes[action.type]) {` into `reduceRepoEntity`.

Like:

```js
export function entities(state, action) {
   // do the merge payload.entities merge here

   return {
      ...state,
      ...reduceRepoEntity(state, action)
      // ...reduceSnapEntity(state, action) in future
   }
}

function reduceRepoEntity(state, action) {
   if (RepoActionTypes[action.type]) {
      return reduceEntityHelper(state, action, repository, 'repos');
   }
   return state;
}
```

Maybe even `reduceEntityHelper` could do the conditional thing, but it would make it even more magical.

earnubs: I'm inclined to keep this as it is for now, and come back to it on the snaps refactor. The return  from your example in the case where state is undefined would be a new state object, I'm not 100% sure if that's OK to do.

I'm also not sure we even need the check on RepoActionType, it seems like a sane thing to do to me but if it's not there state just passes through untouched ... and maybe that's better.

But, given deadlines, I suggest merging and dealing with this later?

test/unit/src/common/reducers/entities/t_repository.js
```diff
@@ -0,0 +1,66 @@
...
+
+  context('update entity reducer', function() {
+
+    let state;
+
+    beforeEach(function() {
+      state = repository(undefined, {
+        type: 'REPO_ADD',
+        payload: {
+          id: 1001
```
bartaz: I don't like `1001` being hardcoded all over the place instead of imported from fixtures or similar, but it doesn't matter that much now.

But we should consider making it into variable or constant imported from fixture later.

earnubs: d'oh! I created the fixture but didn't use it.

Issue #465, opened 2017-03-13
Speed up future development & fixes: Refactor snaps in launchpad state in store [4]


earnubs:
Refactoring of the structure of snaps related data in application store. Meant to improve decoupling of the data and making it easier and quicker to fix bugs and implement new snaps related features in future.

~~Currently blocks progress on #300~~

----

Refactor the `snapBuilds`, `snaps` and `registerName` state to work off a single list  of snaps-in-launchpad, based off whatever best practice established in #464, with the investigation into potential benefits of also merging into the source repositories structure established in #464.

This should be split up further if possible in light of work done in #464.

- [x] ~~refactor `snaps` store - use only `id`s in snaps list, keep snap objects (with their state) in `entities` [1-2 days, should unblock #300]~~ DONE in #624
- [x] ~~refactor snaps related API calls to use new CALL_API middleware (possibly split more into several tasks for each API call or action)~~
  - find snaps API calls already refactored
  - register name API calls are to the store so no need for them to go through middleware
  - snap builds API calls to be refactored separately as part of #639
- [x] ~~refactor snap related API responses to return normalized and camelCased props (just like repo APIs do now)~~ DONE #650
- [x] ~~try moving register name state from `registerName` reducer to snaps objects in entities~~ DONE in #637
  - ~~update ids used by register name actions to be compatible with snap ids (url) and not repo full names as they are now~~
- [x] ~~refactor snap builds - move them to entities? keep builds ids list in snaps in entities? (to be further discussed)~~ MOVED TO #639

bartaz:
This depends on #464, so is blocked until we have more established practices for these refactoring tasks.

PR #598, opened 2017-04-06
Merge master into refactor redux part i

bartaz:


src/common/reducers/entities/index.js
```diff
@@ -12,6 +12,11 @@ export function entities(state = { snaps: {}, repos: {} }, action) {
     return merge({}, state, action.payload.entities);
   }
 
+  // TODO merge master - duplication of above, do we need to support both formats?
+  if (action.payload && action.payload.response && action.payload.response.entities) {
```
bartaz: One of the issues after merge is that `entities` from API may be directly in action `payload` (if fetch was done directly) or in `response` prop (if it went through CALL_API middleware).

src/common/reducers/entities/index.js
```diff
@@ -12,7 +12,9 @@ export function entities(state = { snaps: {}, repos: {} }, action) {
     return merge({}, state, action.payload.entities);
   }
 
-  // TODO merge master - duplication of above, do we need to support both formats?
+  // XXX
+  // some action that don't use CALL_API middleware pass entities in payload directly
+  // while CALL_API middleware does it in response prop
   if (action.payload && action.payload.response && action.payload.response.entities) {
```
bartaz: We need to read `entities` both from action payload directly and response object to support current state of 2 ways of calling APIs (direct or via CALL_API).

src/common/reducers/snaps.js
```diff
@@ -51,17 +51,17 @@ export function snaps(state = {
         isFetching: false,
         success: true,
         snaps: [
-          ...action.payload.snaps
+          ...action.payload.response.payload.snaps
```
bartaz: Nested in a bit ugly way, but let's look into that later when we refactor snaps store.

src/common/reducers/entities/index.js
```diff
@@ -12,6 +12,13 @@ export function entities(state = { snaps: {}, repos: {} }, action) {
     return merge({}, state, action.payload.entities);
   }
 
+  // XXX
+  // some action that don't use CALL_API middleware pass entities in payload directly
```
earnubs: ref #465

src/common/store/configureStore.js
```diff
@@ -2,18 +2,25 @@ import { createStore, applyMiddleware, compose } from 'redux';
...
 import { browserHistory } from 'react-router';
 import thunk from 'redux-thunk';
+
 import rootReducer from '../reducers';
 import { analytics } from '../middleware';
+import callApi from '../middleware/call-api';
+import { conf } from '../helpers/config';
 
-export default function configureStore(preloadedState) {
+export default function configureStore(preloadedState, csrfToken) {
```
earnubs: it's just occurred to me that the token could have been in preloadedstate

PR #603, opened 2017-04-06
keep selected state on success

earnubs:


src/common/reducers/entities/repository.js
```diff
@@ -23,7 +23,14 @@ export default function repository(state={
         isSelected: true
       };
     }
-    case ActionTypes.REPO_SUCCESS:
+    case ActionTypes.REPO_SUCCESS: {
+      return {
+        ...state,
+        isSelected: true,
```
bartaz: I would say that `isSelected` doesn't need to be set here, as this action doesn't have anything to do with selection.

earnubs: I've been thinking through possible use cases and on balance I agree.

PR #609, opened 2017-04-07
Refactor-redux: CALL_API

katiefenn:
- Added support for normalised store to FETCH_SNAPS_SUCCESS and
  REMOVE_SNAP_SUCCESS

### DIFF
--- a/src/common/reducers/snaps.js
+++ b/src/common/reducers/snaps.js
@@ -1,6 +1,7 @@
 import * as ActionTypes from '../actions/snaps';
 import * as RegisterNameActionTypes from '../actions/register-name';
 import { getGitHubRepoUrl } from '../helpers/github-url';
+import union from 'lodash/union';
 
 // TODO move to selector
 function findSnapByFullName(snaps, fullName) {
@@ -51,7 +52,7 @@ export function snaps(state = {
         snaps: [
           ...action.payload.response.payload.snaps
         ],
-        ids: action.payload.response.result,
+        ids: union(state.ids, action.payload.response.result),
         error: null
       };
     case ActionTypes.FETCH_SNAPS_ERROR:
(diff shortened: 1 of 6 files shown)


src/common/reducers/snaps.js
```diff
@@ -51,7 +52,7 @@ export function snaps(state = {
         snaps: [
           ...action.payload.response.payload.snaps
         ],
-        ids: action.payload.response.result,
+        ids: union(state.ids, action.payload.response.result),
```
earnubs: With `union` and without an action to delete a snap, and subsequently remove it from the snap this list, the snap index will never update on remove, nor dependent selectors.

bartaz: `REMOVE_SNAP_SUCCESS` is handled properly below, so that shouldn't be a case.

src/server/handlers/schema.js
```diff
@@ -18,6 +18,6 @@ export const repo = new schema.Entity('repos', {
 export const repoList = new schema.Array(repo);
 
 export const snap = new schema.Entity('snaps', {}, {
-  idAttribute: 'name'
+  idAttribute: 'git_repository_url'
```
bartaz: I wonder if it is safe to assume that there is going to be only one snap with given repository URL in LP.

Especially in case of organisation repos, that can accessed and possibly built by multiple users.

@cjwatson Is it ok to use GH repository URL as a snap identifier on BSI? Is there only one snap for given repo or can there be multiple in some cases?

src/common/reducers/snaps.js
earnubs: Ah, didn't realise that work was already done :)

src/server/handlers/schema.js
katiefenn: @bartaz: This was a dilemma I faced creating this PR. The repository URL is already used as a unique identifier by the Launchpad API endpoint, so it makes sense to maintain this consistency in the frontend. For now.

I think `name` would make a suitable identifier for snaps, but it affects a much greater part of the application than the Redux store and should be out-of-scope for this task.

Issue #464, opened 2017-03-13
refactor source repositories state in store [11+]


earnubs:
Refactor the source repositories data in the store to include the `selectRepositoriesForm`, `repository` and `repositoriesStatus` in one list based on `repositories` (which is essentially just the response from gh repo api). 

Consider use of `normalizr` module or similar to bring control to api responses according to schema with the ultimate goal of the source repo list being the main table against which other data is either indexed or selected.

This card can be broken down further along the lines of:

- [x] - investigate normalizr for benefits to design of store [2]
- [x] - apply schema to source repos list api response [1]
- [x] - add selector for `selectedRepositories` [2]
- [x] - merge `selectRepositoriesForm` to `repositories` [2]
- [x] - merge `repositoriesStatus` into `repositories` [2]
- [x] - consistent action naming [1]
<s>- merge `repository` into `repositories` [1]</s> See #562

earnubs:
See also #465 #466 #467

PR #613, opened 2017-04-10
612: Fix select repositories page pagination

katiefenn:
- Added pageLinks to action payload
- Improved unit tests with stub data better reflecting a real API
  response

### DIFF
--- a/src/common/actions/repositories.js
+++ b/src/common/actions/repositories.js
@@ -11,10 +11,13 @@ export const REPOSITORIES_SUCCESS = 'REPOSITORIES_SUCCESS';
 export const REPOSITORIES_FAILURE = 'REPOSITORIES_FAILURE';
 
 
-export function fetchRepositoriesSuccess(repos) {
+export function fetchRepositoriesSuccess(result) {
   return {
     type: REPOSITORIES_SUCCESS,
-    payload: repos
+    payload: {
+      ...result.payload,
+      pageLinks: result.pageLinks
+    }
   };
 }
 
@@ -41,7 +44,7 @@ export function fetchUserRepositories(pageNumber) {
         throw getError(response, result);
       }
 
-      dispatch(fetchRepositoriesSuccess(result.payload));
+      dispatch(fetchRepositoriesSuccess(result));
     } catch (error) {
       // TODO: Replace with logging helper
       console.warn(error); // eslint-disable-line no-console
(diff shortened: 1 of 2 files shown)


src/common/actions/repositories.js
```diff
@@ -11,10 +11,13 @@ export const REPOSITORIES_SUCCESS = 'REPOSITORIES_SUCCESS';
...
 
 
-export function fetchRepositoriesSuccess(repos) {
+export function fetchRepositoriesSuccess(result) {
   return {
     type: REPOSITORIES_SUCCESS,
-    payload: repos
+    payload: {
+      ...result.payload,
+      pageLinks: result.pageLinks
```
katiefenn: Added `pageLinks` to action payload, because this is where the reducer expects them to be.

test/unit/src/common/actions/t_repositories.js
```diff
@@ -32,20 +32,42 @@ describe('repositories actions', () => {
...
+          && action.payload.result == result.payload.result;
+      });
+    });
 
+    it('should supply pagelinks in action payload', () => {
       store.dispatch(action);
-      expect(store.getActions()).toInclude(expectedAction);
+      expect(store.getActions()).toHaveActionsMatching((action) => {
+        return action.payload.pageLinks == result.pageLinks;
+      });
```
katiefenn: Improved unit tests to make sure page links are supplied to reducer, and made the stub API response more consistent with a real-world response.

src/common/actions/repositories.js
bartaz: Following conventions from #464 actions should know as little as possible about structure of the response and just pass response payload to reducer if possible.

Maybe refactoring it to use callApi middleware would help with that.

test/unit/src/common/actions/t_repositories.js
```diff
@@ -32,20 +32,42 @@ describe('repositories actions', () => {
...
+            123: { fullName: 'test1' },
+            456: { fullName: 'test2' }
+          }
+        },
+        result: {
+          0: 123,
+          1: 456
+        }
+      },
+      pageLinks: {
```
bartaz: It's a bit weird that pageLinks are outside of payload - it also makes action more dependant on structure of the response.

I think that during refactoring pageLinks should be moved into payload together with rest of normalized response.

src/common/actions/repositories.js
katiefenn: I anticipated this would be a short-term measure before a refactor to use the callApi middleware.

test/unit/src/common/actions/t_repositories.js
katiefenn: 👍

PR #616, opened 2017-04-11
614: Refactor repositories API calls

katiefenn:
- Replaced fetch repositories API calls with call-api middleware

### DIFF
--- a/src/common/actions/repositories.js
+++ b/src/common/actions/repositories.js
@@ -6,16 +6,16 @@ export const REPOSITORIES_SUCCESS = 'REPOSITORIES_SUCCESS';
 export const REPOSITORIES_FAILURE = 'REPOSITORIES_FAILURE';
 
 export function fetchUserRepositories(pageNumber) {
-  let urlParts = ['/api/github/repos'];
+  let path = '/api/github/repos';
 
   if (pageNumber) {
-    urlParts.push('page/' + pageNumber);
+    path += ('?page=' + pageNumber);
   }
 
   return {
     [CALL_API]: {
       types: [REPOSITORIES_REQUEST, REPOSITORIES_SUCCESS, REPOSITORIES_FAILURE],
-      path: urlParts.join('/'),
+      path: path,
       options: {
         headers: { 'Content-Type': 'application/json' },
         credentials: 'same-origin'
--- a/src/common/actions/repositories.js
+++ b/src/common/actions/repositories.js
@@ -1,63 +1,26 @@
-import 'isomorphic-fetch';
-
-import { checkStatus, getError } from '../helpers/api';
-import { conf } from '../helpers/config';
+import { CALL_API } from '../middleware/call-api';
 import { fetchUserSnaps } from './snaps.js';
-
-const BASE_URL = conf.get('BASE_URL');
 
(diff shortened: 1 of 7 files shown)


src/common/actions/repositories.js
```diff
@@ -1,66 +1,29 @@
...
 export function fetchUserRepositories(pageNumber) {
-  return async (dispatch) => {
-    let urlParts = [BASE_URL, 'api/github/repos'];
+  let urlParts = ['/api/github/repos'];
 
-    dispatch({
-      type: REPOSITORIES_REQUEST
-    });
+  if (pageNumber) {
+    urlParts.push('page/' + pageNumber);
```
bartaz: It's not really part of this PR, but I wonder why we send params to this one API via url parts where in every other API we use query.

test/routes/src/server/routes/t_github.js
```diff
@@ -640,12 +641,21 @@ const hasStatus = (expected) => {
   };
 };
 
-const hasMessage = (expected) => {
+const hasPayloadMessage = (expected) => {
```
bartaz: Why these are called `hasPayloadMessage` `hasBodyMessage` when they seem to be checking `code` property?

src/common/reducers/repositories.js
```diff
@@ -20,14 +20,14 @@ export function repositories(state = {
...
-        pageLinks: action.payload.pageLinks
+        ids: union(state.ids, action.payload.response.result),
+        pageLinks: action.payload.response.pageLinks
       };
     case ActionTypes.REPOSITORIES_FAILURE:
       return {
         ...state,
         isFetching: false,
-        error: action.payload,
+        error: action.payload.error,
```
bartaz: With `callApi` middleware we need to nest error in payload instead of error just being the payload?

I guess it's needed so we can make sure we also pass any other payload items if needed, right?

It breaks a little bit Flux Standard Actions pattern, but I guess it's a compromise we need to make. I think we were already breaking FSA in this regard to send both `id` and `error` in payload. So I guess it needs to stay like that.

So I guess in our FSA 'variant' we should assume that if `action.error` is true then `action.payload.error` should be an error object.

src/common/actions/repositories.js
bartaz: Or maybe - since there are small updated to serverside of this API in this PR, maybe it can be updated to query be consistent with rest of API calls?

katiefenn: Yeah, I think it's worth a bit of time.

src/common/reducers/repositories.js
katiefenn: Yes, this is the compromise that I made.

FSA _requires_ that `action.error` is a bool, but passing the error as the payload itself is only a _convention_.

test/routes/src/server/routes/t_github.js
katiefenn: `hasPayloadMessage` is checking for a code property inside `action.payload` whereas `hasBodyMessage` is checking for a code property at the top level of `action`. This is part of the gradual removal of the `payload` property from API responses, and when it has been removed across the board `hasPayloadMessage` can be removed.

I think it's worth doing this work in stages, but open to discussion about whether a different approach should be taken.

PR #624, opened 2017-04-11
Refactor snaps redux store

bartaz:
First part of #465

- removes snaps objects from `snaps` reducer and only keeps list of `ids` there
- snaps data is now fully kept in `entities`
- updated relevant components, selectors, etc to accommodate the change

This should be a good first step refactoring, further work is needed around actions, API calls etc.
As described in TODO items in code and in description of #465

### DIFF
--- a/src/common/components/repositories-home/index.js
+++ b/src/common/components/repositories-home/index.js
@@ -64,6 +64,8 @@ class RepositoriesHome extends Component {
   }
 
   renderRepositoriesList() {
+    // TODO: bartaz refactor
+    // does it need snaps (ids) or entities?
     const { snaps, snapBuilds } = this.props;
 
     return (
--- a/src/common/components/repositories-home/index.js
+++ b/src/common/components/repositories-home/index.js
@@ -2,6 +2,7 @@ import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router';
 
+import { hasSnaps } from '../../selectors';
 import { fetchUserSnaps } from '../../actions/snaps';
 import { fetchBuilds } from '../../actions/snap-builds';
 import { LinkButton } from '../vanilla/button';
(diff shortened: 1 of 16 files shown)


src/common/components/repositories-home/index.js
```diff
@@ -19,16 +20,19 @@ const SNAP_POLL_PERIOD = (15 * 1000);
 
 class RepositoriesHome extends Component {
   fetchData(props) {
-    const { snaps } = props;
+    const { hasSnaps, snaps, entities } = props;
 
-    if (snaps.success) {
+    // check both success and fetching to avoid redirecting
+    // based on previous success
+    if (snaps.success && !snaps.isFetching) {
```
bartaz: We had issues with this before and still in some places we have different corner cases that require confusing checks on `success` and/or `isFetching`.

Probably not worth handling in this PR, but we should add it as discussion point.

src/common/components/repositories-home/index.js
```diff
@@ -64,11 +68,13 @@ class RepositoriesHome extends Component {
   }
 
   renderRepositoriesList() {
-    const { snaps, snapBuilds } = this.props;
+    // TODO: bartaz refactor
+    // should it fetch data for first time heading (it already does need it anyway probably)
```
bartaz: FirstTimeHeading needs data about snaps and builds (to display progress) this data is currently fetched by parent component even if it doesn't directly need it.

It even doesn't pass it to FTH, because it uses connect on its own and gets data from selectors...

This is a bit strange case. We have it in 2 places (my repos and select repositories - where first time heading is shown).

src/common/components/select-repositories-page/index.js
```diff
@@ -30,11 +30,14 @@ class SelectRepositoriesPage extends Component {
   }
 
   render() {
+    // TODO: bartaz refactor
     // XXX this should fetch snaps and repos and pass to children ?
-    const { snaps, snapBuilds } = this.props;
+    // should it fetch data for first time heading (or others?)
```
bartaz: Same case of fetching for FirstTimeHeading as described in other comment.

src/common/reducers/entities/index.js
```diff
@@ -26,9 +28,28 @@ export function entities(state = {
...
 
+  // update snaps on register name actions
+  if (RegisterNameActionTypes[action.type]) {
+    // TODO refactor
+    // register name actions use fullName as id instead of repo url
+    const snapAction = {
+      ...action,
+      payload: {
+        ...action.payload,
+        id: getGitHubRepoUrl(action.payload.id)
```
bartaz: Actions from `register-name` use repo full name as an id, while we use url as snap in store.
So before it's changed, we need to transform one id into another.

src/common/components/repositories-home/index.js
earnubs: "Success is the resolution of a Promise, not a matter of state." -- Abraham Lincoln

katiefenn: What do you think the solution would look like? A greater range of selectors to handle nuances in snaps state?

bartaz: Honestly I'm not sure right now. We would need to identify the different cases where we use these statuses and see how we can solve it.

I opened #625 to deal with that later.

Issue #300, opened 2017-03-03
[name not matching] snapcraft.yaml name is registered by someone else [2]


nottrobin:
---

Copied from Trello card: https://trello.com/c/HsTMP9Ni/328-name-not-matching-snapcraft-yaml-name-is-registered-by-someone-else
Labels:

---

TODO:

- [x] add API call to check if name is registered in the store (by anyone)
  - try to request a throwaway package_upload macaroon from the store's /dev/api/acl/ endpoint restricted to the name in question; that will fail if the name is not registered
- [x] add API call to try registering name and interpret 409 conflict to know if name is registered by current user or someone else:
  - try to request a throwaway package_upload macaroon from the store's /dev/api/acl/ endpoint restricted to the name in question; that will fail if the name is not registered
- [x] use API requests to set status of the snap (who owns name from snapcraft.yaml) after user signs in to SSO
- [x] display proper message in UI based on snap status

bartaz:
@cjwatson do we currently have the means to check if name was registered by someone else in the store (without trying to register it and checking error response)?

cjwatson:
@bartaz What's wrong with trying to register it and checking the error response?  That seems like a perfectly reasonable way to do it.  The `register-name` handler returns fine-grained error codes that distinguish between this case and other cases.

bartaz:
@cjwatson It may have a side-effect of registering the name user didn't want to register.

The case is they have some name registered for their snap and different name in snapcraft.yaml. Maybe they just made a typo in the snapcraft.yaml they didn't intent.
We just want to check if the name is registered by a different user so we can suggest them to use different one.

If automatically we will try to register the name just to check if it is already registered we may end up registering the name that was maybe just a typo.

cjwatson:
@bartaz So, it only makes a difference if the name is not registered at all.  In that case BSI can tell the difference as follows:

 - try to request a throwaway `package_upload` macaroon from the store's `/dev/api/acl/` endpoint restricted to the name in question; that will fail if the name is not registered
 - if that succeeds, try `/dev/api/register-name/`, which will return 409 with an `already_registered` code if the name belongs to somebody else

bartaz:
UI for that requires #297 (knowing the location of snapcraft.yaml in repo)

matthewpaulthomas:
(@magdalena-mirowicz Please don’t mark an issue as “Design required” with no explanation why. This has been in the spec since February 7th.)

bartaz:
[...]

Issue #272, opened 2017-03-03
[API] Allow more SCA requests to be made cross-origin [2]


nottrobin:
We need to get rid of our proxy endpoints so that the `package_upload_request` macaroon doesn't need to go through our servers.

---

Copied from Trello card: https://trello.com/c/2Gh7N9rJ/191-api-allow-more-sca-requests-to-be-made-cross-origin
Labels:

nottrobin:
Priority: 5

A security fix to the store API. It prevents an attacker who gains access to build.snapcraft.io from being able to get further.

katiefenn:
C: 5
K: 5
B: 5

sparkiegeek:
https://code.launchpad.net/~adam-collard/software-center-agent/add-cors-headers/+merge/322856 seeks to add CORS headers to the store so we can address this

sparkiegeek:
This was reverted in #698

PR #633, opened 2017-04-13
[PART OF #300] Added API route to check snap name ownership in store

bartaz:
First part of #300 

Action creator to check name ownership in store, by requesting package_upload macaroon and then trying to register the name (expecting conflict error to determine who owns the name).

Actions are dispatched but currently not used by any reducer yet. To be added in upcoming PRs.

For discussion on the implementation see comments on #300


environments/test.env
```diff
@@ -13,8 +13,8 @@ LP_API_TOKEN=token key
 LP_API_TOKEN_SECRET=token secret
 LP_WEBHOOK_SECRET=dummy-lp-webhook-secret
 
-STORE_API_URL=https://myapps.developer.ubuntu.com/dev/api
-STORE_DEVPORTAL_URL=https://myapps.developer.ubuntu.com/dev
+STORE_API_URL=http://myapps.developer.ubuntu.com/dev/api
+STORE_DEVPORTAL_URL=http://myapps.developer.ubuntu.com/dev
```
bartaz: For some reason nock mocks were failing (not being properly satisfied in tests) when it was `https`.
It seemed like nock was expecting port :443 in URL but it was not there...

Changing test URLs to http resolved the issue. But it's obviously more like a workaround rather that solution.

src/server/handlers/store.js
```diff
@@ -101,3 +101,90 @@ export const signAgreement = async (req, res) => {
   const json = await response.json();
   return res.status(response.status).send(json);
 };
+
+export const nameOwnership = async (req, res) => {
+  const snapName = req.query.snap_name;
+  const root = req.query.root;
+  const discharge = req.query.discharge;
```
cjwatson: I know that various things in this file are already receiving root and discharge macaroons, but they're all thin shims over store APIs rather than containing much logic of their own, and they can be replaced with direct calls to the store from the client side once the store sets the appropriate CORS headers.  This method is different since it contains non-trivial logic of its own.

So, could you move all this to the client side to avoid building up more technical debt that we'll need to rearrange once #272 is done?  All the server-side facilities are already in place: `/dev/api/acl/` can be called directly using CORS, and `/dev/api/register-name/` already has a shim in this file.

bartaz: OK. Makes sense. WIll do.

src/common/helpers/github-url.js
```diff
@@ -2,6 +2,12 @@ import parseGitHubUrl from 'parse-github-url';
 
 export const getGitHubRepoUrl = (owner, name) => {
   let repository = !name ? owner : `${owner}/${name}`;
+
+  // if already a valid GH repository
+  if (repository.indexOf('https://github.com/') === 0) {
```
bartaz: If we pass already valid GH URL as only argument, lets just return it.

src/common/actions/register-name.js
```diff
@@ -9,16 +9,26 @@ import { requestBuilds } from './snap-builds';
 
 const BASE_URL = conf.get('BASE_URL');
 
+// action types
 export const REGISTER_NAME = 'REGISTER_NAME';
 export const REGISTER_NAME_SUCCESS = 'REGISTER_NAME_SUCCESS';
 export const REGISTER_NAME_ERROR = 'REGISTER_NAME_ERROR';
 export const REGISTER_NAME_CLEAR = 'REGISTER_NAME_CLEAR';
 
+export const CHECK_NAME_OWNERSHIP = 'CHECK_NAME_OWNERSHIP';
```
katiefenn: I've been following a convention of using `REQUEST`, `SUCCESS` and `ERROR` suffixes, matching the [Redux real world example](https://github.com/reactjs/redux/blob/master/examples/real-world/src/actions/index.js#L36).

src/common/actions/register-name.js
```diff
@@ -75,21 +85,24 @@ async function signAgreement(root, discharge) {
   }
 }
 
-export async function internalRegisterName(root, discharge, snapName,
-                                           repositoryUrl) {
-  const response = await fetch(`${BASE_URL}/api/store/register-name`, {
+async function requestRegisterName(root, discharge, snapName) {
+  return await await fetch(`${BASE_URL}/api/store/register-name`, {
```
katiefenn: Double `await`?

src/common/actions/register-name.js
```diff
@@ -198,3 +211,79 @@ export function registerNameClear(id) {
     }
   };
 }
+
+export async function internalNameOwnership(root, discharge, snapName) {
+  if (!snapName) {
+    throw new APICompatibleError({
```
katiefenn: Should this be a APICompatibleError given the source of this error is from the application and not the API? A standard `Error` would probably do.

src/common/actions/register-name.js
bartaz: Oh... that's interesting that it doesn't throw syntax error and actually works...

src/common/actions/register-name.js
bartaz: I guess you're right. Especially that it's most likely to be an error caused by developer mistake rather then application state.

src/common/helpers/github-url.js
katiefenn: When does `owner` include the full URL?

bartaz: Sometimes we pass the redux store 'id' to this function. In some legacy places (like register name actions) this id is repo fullName (so the function turns it into URL).

With our new convention the id in store (for the snap) is full repo URL, so the parts of code that pass the id here now pass the full URL.

In theory it may be a temporary thing until we have ids as full URLs everywhere for snaps in store, so the transformation would not be needed.

bartaz: Related for example to this:
https://github.com/canonical-ols/build.snapcraft.io/pull/624/files#r110907873

src/common/actions/register-name.js
katiefenn: If this error is caused by developer mistake, is it really something we want to handle by keeping it in the store?

src/common/helpers/github-url.js
katiefenn: Okay 👍

PR #635, opened 2017-04-13
614: Add CSRF tokens to single repo API calls

katiefenn:
- Refactored actions to use CALL_API middleware
- Rewrote CALL_API middleware to use promises instead of async / await
- Renamed single repo action type identifiers
- Changed webhook-already-created response to return 200
- Added verify token middleware to add repo endpoint
- Added verify token middleware to webhook endpoint

### DIFF
--- a/src/common/actions/repository.js
+++ b/src/common/actions/repository.js
@@ -1,9 +1,7 @@
 import { push } from 'react-router-redux';
-import { checkStatus, getError } from '../helpers/api';
 
-import { conf } from '../helpers/config';
-
-const BASE_URL = conf.get('BASE_URL');
+import { APICompatibleError } from '../helpers/api';
+import { CALL_API } from '../middleware/call-api';
 
 export const REPO_TOGGLE_SELECT = 'REPO_TOGGLE_SELECT';
 
@@ -16,58 +14,58 @@ export function toggleRepositorySelection(id) {
   };
 }
 
-export const REPO_ADD = 'REPO_ADD';
-export const REPO_SUCCESS = 'REPO_SUCCESS';
-export const REPO_FAILURE = 'REPO_FAILURE';
+export const REPO_ADD_REQUEST = 'REPO_ADD_REQUEST';
+export const REPO_ADD_SUCCESS = 'REPO_ADD_SUCCESS';
+export const REPO_ADD_FAILURE = 'REPO_ADD_FAILURE';
 export const REPO_RESET = 'REPO_RESET';
 
-export function addRepos(repositories, owner) {
-  return (dispatch) => {
-    const promises = repositories.map(
-      (repository) => {
-        return dispatch(addRepo(repository));
-      }
-    );
-    return Promise.all(promises).then(() => {
+export function addRepos(repos, owner) {
+  return async (dispatch) => {
+    try {
+      await Promise.all(repos.map(async (repo) => {
+        await dispatch(createWebhook(repo));
+        return dispatch(addRepo(repo));
+      }));
+
       if (owner) {
         dispatch(push(`/user/${owner}`));
       }
-    });
+    } catch(error) {
+      dispatch(addRepoError(error.action.id, error));
+    }
   };
 }
 
-// add a repository to launchpad's build queue
-export function addRepo(repository) {
-  const { id, url, owner, name } = repository;
+function addRepo(repository) {
+  const { id, url } = repository;
 
-  return async (dispatch) => {
-    dispatch({
-      type: REPO_ADD,
-      payload: {
-        id
-      }
-    });
-
-    try {
-      await createWebhook(owner, name);
-
-      // XXX
-      // actual LP API call that we use on server side returns a snap representation
-      // that we could use to immediatelly update client side state with newly
-      // added snap
-      const response = await fetch(`${BASE_URL}/api/launchpad/snaps`, {
+  return {
+    id,
+    [CALL_API]: {
+      path: '/api/launchpad/snaps',
+      types: [REPO_ADD_REQUEST, REPO_ADD_SUCCESS],
+      options: {
         method: 'POST',
         headers: { 'Content-Type': 'application/json' },
         body: JSON.stringify({ repository_url: url }),
         credentials: 'same-origin'
-      });
-
-      await checkStatus(response);
-      dispatch(addRepoSuccess(id));
-    } catch (error) {
-      dispatch(addRepoError(id, error));
-      return Promise.reject(error);
+      }
     }
+  };
+}
+
+function addRepoError(id, error) {
+  return {
+    type: REPO_ADD_FAILURE,
+    payload: {
+      id,
+      error: new APICompatibleError({
+        code: error.json.body.code,
+        message: 'Repo cannot be added at this time',
+        detail: `Failed to add repo: ${error}`
+      })
+    },
+    error: true
   };
 }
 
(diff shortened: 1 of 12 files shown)


src/common/actions/repository.js
```diff
@@ -16,104 +14,80 @@ export function toggleRepositorySelection(id) {
   };
 }
 
-export const REPO_ADD = 'REPO_ADD';
-export const REPO_SUCCESS = 'REPO_SUCCESS';
-export const REPO_FAILURE = 'REPO_FAILURE';
+export const REPO_ADD_REQUEST = 'REPO_ADD_REQUEST';
+export const REPO_ADD_SUCCESS = 'REPO_ADD_SUCCESS';
+export const REPO_ADD_FAILURE = 'REPO_ADD_FAILURE';
```
katiefenn: Renamed action types to make them easier to identify.

src/common/actions/repository.js
```diff
@@ -16,104 +14,80 @@ export function toggleRepositorySelection(id) {
...
-      (repository) => {
-        return dispatch(addRepo(repository));
-      }
-    );
-    return Promise.all(promises).then(() => {
+export function addRepos(repos, owner) {
+  return async (dispatch) => {
+    try {
+      await Promise.all(repos.map(async (repo) => {
+        await dispatch(createWebhook(repo));
```
katiefenn: Extracted dispatch of `createWebhook` from `addRepo` so that each can be rewritten to use the CALL_API middleware.

src/common/actions/repository.js
```diff
@@ -16,104 +14,80 @@ export function toggleRepositorySelection(id) {
...
   return {
-    type: REPO_RESET,
+    type: REPO_ADD_FAILURE,
     payload: {
-      id
-    }
+      id,
+      error: new APICompatibleError({
+        code: error.json.body.code,
+        message: 'Repo cannot be added at this time',
```
katiefenn: Generic error message - end user should not need to know about CSRF tokens.

src/common/actions/repository.js
```diff
@@ -16,104 +14,80 @@ export function toggleRepositorySelection(id) {
...
-    error: true
+    [CALL_API]: {
+      path: '/api/github/webhook',
+      options: {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ owner, name }),
+        credentials: 'same-origin'
+      }
+    }
```
katiefenn: No action types supplied. We don't currently handle problems installing single webhooks, but when a call results in an error it returns a rejected promise that is caught in [addRepos](https://github.com/canonical-ols/build.snapcraft.io/pull/635/files#diff-3d0e5cd44d7df7fc7a2fecb29c4a6deaR33)

src/common/components/select-repository-list/index.js
```diff
@@ -55,18 +53,6 @@ export class SelectRepositoryListComponent extends Component {
...
-    const { payload = undefined } = error;
-
-    if (payload) {
-      if (payload.code === SNAP_NAME_NOT_REGISTERED_ERROR_CODE) {
-        return this.getSnapNotRegisteredMessage(payload.snap_name);
-      } else {
-        return payload.message;
-      }
-    }
-  }
```
katiefenn: Some baggage left over from when the file was copied from `components/repositories-list/index.js`. The snap-name-not-registered error is never shown in this component, so this method is largely redundant.

src/common/middleware/call-api.js
```diff
@@ -17,39 +17,67 @@ export default defaults => () => next => async action => {
...
+            },
+            error: true
+          }));
+        }
+
+        error.action = action;
+
+        return Promise.reject(error);
+      });
+    });
```
katiefenn: I've rewritten the call-api middleware to use promises instead of async/await. Source mapping for async/await code seems to be a bit flakey, I found it hard to step through this code to debug error responses handled by Fetch and problems in reducers.

I think this is a reasonable tradeoff within the scope of the middleware, I would still be happy to use async/await elsewhere in the application.

It is definitely something we should keep an eye on.

src/common/middleware/call-api.js
```diff
@@ -5,7 +5,7 @@ export const CALL_API = 'CALL_API';
```
katiefenn: I've rewritten the call-api middleware to use promises instead of async/await. Source mapping for async/await code seems to be a bit flakey, I found it hard to step through this code to debug error responses handled by Fetch and problems in reducers.

I think this is a reasonable tradeoff within the scope of the middleware, I would still be happy to use async/await elsewhere in the application.

It is definitely something we should keep an eye on.

src/common/middleware/call-api.js
```diff
@@ -17,39 +17,67 @@ export default defaults => () => next => async action => {
...
+            payload: {
+              error: error
+            },
+            error: true
+          }));
+        }
+
+        error.action = action;
+
+        return Promise.reject(error);
```
katiefenn: In various parts of our codebase we have a few API calls that do not result in actions being dispatched, but instead the resolution of the call is used to influence other actions.

A good example is `createWebhook`, which is now a fully-fledged action creator in this pull request. When this API call returns an error result, the outcome is handled by the `addRepos` action creator and stops the application redirecting the user back to the dashboard. There is currently no need for `createWebhook` to dispatch more actions.

My proposal is to allow the call-api middleware to return resolved or rejected promises so that `dispatch({ [CALL_API]: ... })` to be used with `Promise.all`, allowing API calls to be batched. There is [some precedence](https://github.com/acdlite/redux-promise/blob/master/src/index.js#L20) for middleware doing this, although I admit it feels like "making it up while I'm going along".

Let's see if we can come up with a better solution together, if there is one.

src/server/handlers/github.js
```diff
@@ -196,7 +196,7 @@ export const createWebhook = async (req, res) => {
           return res.status(401).send(RESPONSE_AUTHENTICATION_FAILED);
         case 'Validation Failed':
           // Webhook already created
-          return res.status(422).send(RESPONSE_ALREADY_CREATED);
+          return res.status(200).send(RESPONSE_ALREADY_CREATED);
```
katiefenn: Because this API endpoint is called by the call-api middleware now, it cannot tell the difference between a 422 "already created" error response and any other error. The call-api middleware is designed to treat all non-2XX errors the same and treat it as a failure.

This suggested to me that surfacing a 4XX error to the client was not the correct approach in the first place, because it does not result in a failure outcome for the application.

src/common/components/select-repository-list/index.js
bartaz: OK. But the `//TODO` comment above the method also seems redundant.

src/common/middleware/call-api.js
bartaz: I don't mind having it implemented with promises for sake of debugging.

Maybe it would be worth putting a comment about it into the code itself so if we end up wondering why we don't use async/await here we will have some explanation.

src/common/middleware/call-api.js
bartaz: It would be good to have this discussion in bigger group to find a good compromise.

Creating a webhook is just a one quite simple example, because at this point we don't even do much with the result, we just want to wait till request is finished before proceeding.

We have much more complicated examples around store auth and name registering, where actions involve calling multiple APIs for authorization, exchanging tokens, etc. Trying to make them all fit into CALL_API middleware will likely make middleware full of special case 'workarounds' - like the one with returning a Promise here.

Returning a Promise from the middleware seems a bit inconsistent to me, especially that we can't rely on dispatch ALWAYS returning a promise, because promise will only be returned for if CALL_API finds settings object, not for any other actions.
If we end up having middleware returning promises from dispatch, I would expect that any dispatch call on any action would return a promise, so we would be able to consistently rely on it, not thinking if this particular action goes through the middleware or not.

At this point - before we have the discussion and find a better solution for it, I would suggest we only use CALL_API middleware for API calls that result in redux actions.
Yes, it means that not all of our API calls will go through middleware, but as far as I can tell, the ones that currently don't fit into CALL_API scheme don't really need CSRF tokens (for example store API calls that we intend to do directly to store once it allows cross origin requests).

src/common/components/select-repository-list/index.js
katiefenn: @bartaz Thanks 👍

katiefenn: @bartaz: Fixed.

src/common/middleware/call-api.js
```diff
@@ -17,39 +19,71 @@ export default defaults => () => next => async action => {
...
-      payload: {
-        response: result
+          error.action = action;
+
+          return Promise.reject(error);
+        });
+    })
+    .catch((error) => {
+      if (failureType) {
+        next(createAction({
```
katiefenn: Stopped early-return here if `failureType` is defined. This means a promise is always returns as long as `[CALL_API]` settings are set on action.

This should make it a little easier to understand.

src/common/middleware/call-api.js
```diff
@@ -4,8 +4,10 @@ import { checkStatus, getError } from '../helpers/api';
 export const CALL_API = 'CALL_API';
 
 // A Redux middleware that interprets actions with CALL_API info specified.
-// Performs the call and promises when such actions are dispatched.
-export default defaults => () => next => async action => {
+// Performs the call and promises when such actions are dispatched. This
+// has been implemented using promises because of suspected problems with
+// sourcemapping async / await code.
+export default defaults => () => next => action => {
```
katiefenn: Added comment explaining use of promises instead of async / await.

PR #637, opened 2017-04-18
465 move register name to entities

bartaz:
Part of #465 refactoring.

Moved register name reducer functionality into snap entity reducer.


src/common/reducers/entities/snap.js
```diff
@@ -1,11 +1,50 @@
...
+    error: null
+  };
+
   switch(action.type) {
+    case RegisterNameActionTypes.REGISTER_NAME:
+      return {
+        ...state,
+        registerNameStatus: {
+          ...registerNameInitialStatus,
+          isFetching: true
```
bartaz: I wasn't sure if this should be flat in snap object or better nested in custom registerNameStatus, but success and error seem quite custom here (error representing error of register name action or validation of register name), so I think it's safer to keep it in custom namespace.

PR #650, opened 2017-04-21
Normalize snap responses

bartaz:
Part of #465 

Normalizing responses that contain snap data, whitelist only needed props and make them camelCase.

### DIFF
--- a/src/server/handlers/schema.js
+++ b/src/server/handlers/schema.js
@@ -22,7 +22,7 @@ export const snap = new schema.Entity('snaps', {}, {
   processStrategy: (entity) => {
     return {
       id: entity.git_repository_url,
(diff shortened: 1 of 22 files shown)


src/server/handlers/schema.js
```diff
@@ -18,6 +18,15 @@ export const repo = new schema.Entity('repos', {
...
+  idAttribute: 'git_repository_url',
+  processStrategy: (entity) => {
+    return {
+      id: entity.git_repository_url,
+      gitRepoUrl: entity.git_repository_url,
+      selfLink: entity.self_link,
+      snapcraftData: entity.snapcraft_data,
+      storeName: entity.store_name
+    };
+  }
```
katiefenn: Thanks for this, it makes reading this PR a lot easier.

PR #681, opened 2017-04-24
Suggest snap name based on repo name

### LINKED ISSUE #669: Suggest the repository name as the snap name to register
If there's no snapcraft.yaml to infer the snap name from, then we should pre-fill a suggested name which is created from the repository name - as [mentioned in the spec](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.sf9bdlnfvbkg):

> Otherwise, the default value should be the repo name, lower-cased, with any invalid characters converted to hyphens, and then any leading/trailing hyphens ignored.

bartaz:
Fixes #669 

If snapcraft.yaml is not available we are now suggesting snap name to register based on repo name.

### DIFF
--- a/src/common/components/repository-row/index.js
+++ b/src/common/components/repository-row/index.js
@@ -36,7 +36,10 @@ export class RepositoryRowView extends Component {
     if (props.snap.snapcraftData && props.snap.snapcraftData.name) {
       snapName = props.snap.snapcraftData.name;
     } else {
-      snapName = '';
+      // suggest name based on git repository name
+      snapName = parseGitHubRepoUrl(props.snap.gitRepoUrl).name.toLowerCase();
+      // replace invalid characters with dash are trim leading and trailing dashes
+      snapName = snapName.replace(/[^a-z0-9-]/g, '-').replace(/^-+|-+$/g, '');
     }
 
     this.state = {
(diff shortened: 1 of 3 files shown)


src/common/components/repository-row/index.js
```diff
@@ -407,6 +410,7 @@ export class RepositoryRowView extends Component {
       return (
         <form onSubmit={this.onRegisterSubmit.bind(this, snap.gitRepoUrl)}>
           <input
+            autoFocus={true}
```
bartaz: Added `autofocus` to register name inputs, so they are focused when dropdown is opened and it is more clear that user can edit them and not just blindly register suggested name.

src/common/components/repository-row/index.js
```diff
@@ -36,7 +36,10 @@ export class RepositoryRowView extends Component {
     if (props.snap.snapcraftData && props.snap.snapcraftData.name) {
       snapName = props.snap.snapcraftData.name;
     } else {
-      snapName = '';
+      // suggest name based on git repository name
+      snapName = parseGitHubRepoUrl(props.snap.gitRepoUrl).name.toLowerCase();
```
earnubs: I forget, but isn't github repo url available from state now?

PR #678, opened 2017-04-21
Point browser to SCA for registerName

sparkiegeek:
Note this depends on [a SCA MP](https://code.launchpad.net/~adam-collard/software-center-agent/add-cors-headers/+merge/322856) before it can be deployed. I figure it's worth getting :eyes: on this first :man_shrugging: 

Instead of talking to the backend proxy for `registerName` POST directly to SCA. This removes one case of the security issue highlighted in #272

PR #682, opened 2017-04-24
Point browser to SCA for registerName

sparkiegeek:

Instead of talking to the backend proxy for `registerName` POST directly to SCA. This removes one case of the security issue highlighted in #272

(note this is a re-run of #678 but with the code living in the canonical-ols namespace)


src/common/actions/auth-store.js
```diff
@@ -270,7 +272,13 @@ async function setShortNamespace(root, discharge, userName) {
     }
     return data;
   } else {
-    const json = await response.json();
+    const text = await response.text();
```
earnubs: Out of interest, why doesn't SCA always return valid JSON?

src/common/actions/auth-store.js
```diff
@@ -223,10 +224,11 @@ function checkSignedIntoStoreError(error) {
 }
 
 async function fetchAccountInfo(root, discharge) {
-  const query = { root, discharge };
-  const accountUrl = `${BASE_URL}/api/store/account?${qs.stringify(query)}`;
-  const response = await fetch(accountUrl, {
-    headers: { 'Accept': 'application/json' }
+  const response = await fetch(`${STORE_API_URL}/account`, {
+    headers: {
+      'Authorization': `Macaroon root="${root}", discharge="${discharge}"`,
```
earnubs: This template appears in a number of places, would it be better extracted to a helper function?

src/common/actions/auth-store.js
sparkiegeek: Heh, same point was raised in the PR that got approved for landing here (lack of support for stacked branches in GitHub lead to this approach of C -> B -> A -> master).

Please see https://github.com/canonical-ols/build.snapcraft.io/pull/683  tl;dr: unknown, but being conservative.

src/common/actions/auth-store.js
sparkiegeek: Hehe, same point raised again as in https://github.com/canonical-ols/build.snapcraft.io/pull/683 - same answer tl;dr: I don't think it's worth it.

src/common/actions/auth-store.js
cjwatson: One reason why any service might not return valid JSON would be if it's down such that haproxy returns some kind of dumb text-only response.  We should always handle that (not saying that we do at the moment).

cjwatson: OK, the specific case that the previous code was handling was a 204 response from PATCH with an empty response body (as befits 204).  That's HTTPishly correct since PATCH has no content to send in its response, but we have to take care to cope with it.

However, now that this code is on the client side, the 204 case is already handled by the other branch of this `if` which entirely ignores the response body, so there's no need to handle the empty-body case specially.

I would suggest:

 * go back to the previous `const json = await response.json();` code here
 * add an XXX comment somewhere with a bug reference reminding us to do something about the general case of better handling of non-JSON error responses; I agree at this point that that doesn't need to be done in this branch, but let's not forget about it

sparkiegeek:
:unamused: OK, I stand by my argument that this is unnecessary cognitive load on future code readers, but yield to the reviewer.

src/common/helpers/api.js
```diff
@@ -27,3 +27,8 @@ export class APICompatibleError extends Error {
     this.json = { status: 'error', payload };
   }
 }
+
+// Format for Authorization header using macaroons
+export function getAuthHeader(root, discharge) {
```
earnubs: Probably clearer and less cognitive load if it were getMacaroonAuthHeader, but that's nit picking :)

PR #683, opened 2017-04-24
Point browser to SCA for accountInfo

sparkiegeek:

Instead of talking to the backend proxy for `accountInfo` GET/PATCH directly to SCA. This removes one case of the security issue highlighted in #272


src/common/actions/auth-store.js
```diff
@@ -251,15 +253,15 @@ async function fetchAccountInfo(root, discharge) {
...
     method: 'PATCH',
-    headers: { 'Content-Type': 'application/json' },
-    body: JSON.stringify({
-      short_namespace: userName,
-      root,
-      discharge
-    })
+    headers: {
+      'Authorization': `Macaroon root="${root}", discharge="${discharge}"`,
+      'Content-Type': 'application/json'
```
bartaz: It seems most (if not all) store API calls use the same headers - would it be worth extracting them to one place (possibly function due to dependency on macaroon values)?

src/common/actions/auth-store.js
```diff
@@ -270,7 +272,13 @@ async function setShortNamespace(root, discharge, userName) {
     }
     return data;
   } else {
-    const json = await response.json();
+    const text = await response.text();
```
bartaz: Why not just read the JSON with `.json()`? Some responses are not formatted as JSON?

If not `.json()` will throw an error we can catch.

src/server/handlers/store.js
```diff
@@ -15,29 +15,6 @@ export const getAccount = async (req, res) => {
   return res.status(response.status).send(json);
 };
 
-export const patchAccount = async (req, res) => {
```
bartaz: Shouldn't `getAccount` be removed too? It seems to be removed from router.

src/common/actions/auth-store.js
bartaz: Sure, no problem.

earnubs: 1 reason to do it might be that Macaroon auth bugs are pretty awful to debug, and mitigating a typo here would defend against someone having that journey.

PR #684, opened 2017-04-24
Point browser to SCA for signAgreement

sparkiegeek:

Instead of talking to the backend proxy for `signAgreement` POST directly to SCA. This removes one case of the security issue highlighted in #272


src/common/actions/register-name.js
```diff
@@ -65,16 +65,15 @@ export async function getPackageUploadRequestMacaroon() {
 }
 
 async function signAgreement(root, discharge) {
-  const response = await fetch(`${BASE_URL}/api/store/agreement`, {
+  const response = await fetch(`${conf.get('STORE_API_URL')}/agreement`, {
```
bartaz: We have `STORE_API_URL` as const already.

src/server/routes/store.js
```diff
@@ -1,13 +0,0 @@
...
-import {
-  signAgreement
-} from '../handlers/store';
-
-const router = Router();
-
-router.use('/store/agreement', json());
-router.post('/store/agreement', signAgreement);
-
-export default router;
```
bartaz: If router is now fully removed shouldn't `handlers/store.js` be removed as well?

sparkiegeek: Yes! Killed, thanks.

src/common/actions/register-name.js
sparkiegeek: Oops, re-used.

PR #690, opened 2017-04-27
Add baseline support for organisation repositories

### LINKED ISSUE #620: Organisations: List organisation repositories on the "add repositories" page, so they can be added to the dashboard [4]
- [ ] Ensure organisation repository information is being pulled from GitHub on sign-up
- [ ] Ensure already added organisations are greyed out
- [ ] List organisation repositories in /select-repositories
- [ ] Ensure organisation repositories can be added to the dashboard
- [ ] We need to ensure that only repositories for which the user is an "admin" are shown

### LINKED ISSUE #648: Organisations: Also show organisation repositories on "my repos" dashboard even if you're not admin [2]

cjwatson:
This shows organisation repositories for which the user has admin
permissions in the add-repositories dialog, and shows organisation
repositories that the user can write to on the my-repositories page.
Various aspects of the UI will need to be tidied up in future PRs, but
this at least gets us started.

We now request the `read:org` scope from GitHub, which allows us to tell
which organisations a user is in.

I also included a change to sort repository list rows alphabetically,
since given the rest of this change I was seeing odd results without
that.

Fixes #620; fixes #648.

### DIFF
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -424,18 +424,26 @@ const internalFindSnaps = async (owner, token) => {
   let remainingPrefixes;
 
   try {
-    snaps = await Promise.all(
+    // For each prefix, check for snaps in memcached.
+    const snapsByPrefix = await Promise.all(
       cacheIds.map((cacheId) => memcached.get(cacheId))
     );
-    remainingPrefixes = zip(urlPrefixes, snaps)
-      .filter(([, snap]) => snap === undefined)
+    // Flatten the results from memcached into a single list of all the
+    // snaps we know about so far.
+    snaps = [].concat(
+      ...(snapsByPrefix.filter((snaps) => snaps !== undefined))
+    );
+    // Work out which prefixes have no entries in memcached and thus need to
+    // be checked with Launchpad.
+    remainingPrefixes = zip(urlPrefixes, snapsByPrefix)
+      .filter(([, snaps]) => snaps === undefined)
       .map(([urlPrefix, ]) => urlPrefix);
+
     if (!remainingPrefixes.length) {
-      return [].concat(...snaps).map((entry) => {
+      return snaps.map((entry) => {
         return lpClient.wrap_resource(entry.self_link, entry);
       });
     }
-    snaps = snaps.filter((snap) => snap !== undefined);
   } catch (error) {
     logger.error(`Error getting one of ${cacheIds} from memcached: ${error}`);
     snaps = [];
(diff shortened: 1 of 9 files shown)


src/server/handlers/launchpad.js
```diff
@@ -411,34 +412,60 @@ export const internalFindSnap = async (repositoryUrl) => {
...
   try {
-    const result = await getMemcached().get(cacheId);
-    if (result !== undefined) {
-      return result.map((entry) => {
+    snaps = await Promise.all(
+      cacheIds.map((cacheId) => memcached.get(cacheId))
+    );
+    remainingPrefixes = zip(urlPrefixes, snaps)
+      .filter(([, snap]) => snap === undefined)
+      .map(([urlPrefix, ]) => urlPrefix);
```
earnubs: Consider a new line here, so I can catch my breath :)

(I think the old style guide said new line after assignments)

src/server/handlers/launchpad.js
```diff
@@ -411,34 +412,60 @@ export const internalFindSnap = async (repositoryUrl) => {
...
+    if (!remainingPrefixes.length) {
+      return [].concat(...snaps).map((entry) => {
```
earnubs: is this the same as `[...snaps]`?

test/routes/src/server/routes/t_github-auth.js
```diff
@@ -4,9 +4,17 @@ import nock from 'nock';
 import expect from 'expect';
 import url from 'url';
 
-import auth from '../../../../../src/server/routes/github-auth';
 import db from '../../../../../src/server/db';
+import {
+  listOrganizationsCacheId
+} from '../../../../../src/server/handlers/github';
```
earnubs: FYI We talked recently in the fortnightly JS discussion about moving tests along side src as a way to resolve this `../../../../../` line noise, something for the future maybe, but maybe we should open that meeting up to members of OLS and further afield.

src/server/handlers/launchpad.js
cjwatson: No, I'm using this as a one-level flatten operator:

    > var foo = [[1, 2], [3, 4]];
    undefined
    > [].concat(...foo)
    [ 1, 2, 3, 4 ]
    > [...foo]
    [ [ 1, 2 ], [ 3, 4 ] ]

However, you've caused me to realise that this code is confusingly written (each element of `snaps` here is in fact an array of snaps) and possibly wrong in one place, so I'll tidy that up.

test/routes/src/server/routes/t_github-auth.js
cjwatson: I would be overjoyed to see that change.

PR #692, opened 2017-04-27
Updated visual style of My Repos table

### LINKED ISSUE #321: Distinguish table cell that was clicked on from others in the row [2]
In the “Repos to build and publish” table, when you click a cell that expands the row, there is no indication of which cell you clicked on.

This does not matter for Maas, where clicking any part of the row reveals the same stuff.

But it does matter for us, because clicking a row’s “Configured” cell reveals different stuff from the “Registered for publishing” cell.

_One_ possibility is as illustrated: highlighting just the clicked cell in the same color as the revealed panel, while leaving the row’s other cells in their original color.

---

Copied from Trello card: https://trello.com/c/emsmNfMS/336-distinguish-table-cell-that-was-clicked-on-from-others-in-the-row
Labels:

---

Scope of this task based on new visual design and as discussed in [comment below](#issuecomment-296981651):

Related to table cell highlighting - should be handled in this task:
- [x] removed grey background from the table
- [ ] added arrows icons in expandable cells
- [x] hover effects on cells
- [x] updated style of expanded row (borders, background)

Other table changes (can be covered in this issue):
- [x] updated link style (not blue until hovered)
- [x] updated spacing of the rows
- [x] possibly updated font size and weights

bartaz:
Fixes #321
See there for [visual spec](https://github.com/canonical-ols/build.snapcraft.io/issues/321#issuecomment-293893096).

screen shot 2017-04-27 at 15 12 52

### DIFF
--- a/src/common/components/build-status/buildStatus.css
+++ b/src/common/components/build-status/buildStatus.css
@@ -1,7 +1,3 @@
-.buildStatus {
-  padding-left: 5px;
-}
-
 .buildStatus.green,
 .buildStatus.green a {
   color: $success;
--- a/src/common/components/build-status/buildStatus.css
+++ b/src/common/components/build-status/buildStatus.css
@@ -2,32 +2,6 @@
   padding-left: 5px;
 }
 
-.buildStatus::before {
-  content: '';
-  display: block;
-  width: 5px;
-  position: absolute;
-  top: 0;
-  left: 0;
-  height: 100%;
-}
-
-.buildStatus.green::before {
-  background-color: $success;
-}
-
-.buildStatus.red::before {
-  background-color: $error;
-}
-
-.buildStatus.yellow::before {
-  background-color: $warning;
-}
-
-.buildStatus.grey::before {
-  background-color: $dark-grey;
-}
-
 .buildStatus.green,
 .buildStatus.green a {
   color: $success;
--- a/src/common/components/repository-row/index.js
+++ b/src/common/components/repository-row/index.js
@@ -1,10 +1,9 @@
 import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
-import { Link } from 'react-router';
 import localforage from 'localforage';
 import { bindActionCreators } from 'redux';
 
-import { Row, Data } from '../vanilla/table-interactive';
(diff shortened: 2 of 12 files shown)


src/common/components/build-status/buildStatus.css
```diff
@@ -1,33 +1,3 @@
...
-}
-
-.buildStatus.yellow::before {
-  background-color: $warning;
-}
-
-.buildStatus.grey::before {
-  background-color: $dark-grey;
-}
-
```
bartaz: Build status doesn't need a fancy 'border' anymore.

src/common/components/repository-row/index.js
```diff
@@ -276,27 +275,15 @@ export class RepositoryRowView extends Component {
...
-        <Data col="25">
-          { this.renderSnapName(registeredName, showRegisterNameInput, snap) }
-        </Data>
-        <Data col="30">
+        <DataLink col="28" to={ hasBuilt ? `/user/${fullName}` : null }>
+          <strong>{ fullName }</strong>
+        </DataLink>
+        { this.renderConfiguredStatus(snap) }
+        { this.renderSnapName(registeredName, showRegisterNameInput, snap) }
+        <DataLink col="24" to={ hasBuilt ? `/user/${fullName}` : null }>
```
bartaz: This could use some cleaning up (moving all cells to separate functions - or even small components).
Not sure if it's worth touching now...

Thoughts?

earnubs: It would certainly be easier to read and more concise as small components. If it's for this PR or another is up to you :)

src/common/components/vanilla/table-interactive/table.css
```diff
@@ -37,6 +39,27 @@ $table-row-border-color: $dark-grey;
...
+}
+
+.dataLink {
+  composes: data;
+  color: $cool-grey;
+}
+
+a.dataLink:hover,
+a.dataLink:focus,
+a.dataLink.active {
```
bartaz: I don't like too much putting tag name in CSS here, but it's the easiest thing to do to distinguish cell that is a link from the one that is not without introducing another class name (that will serve exactly the same purpose).

PR #693, opened 2017-04-27
remap api strings to bsi equivalents

### LINKED ISSUE #406: Update build statuses as per the spec [1]
Currently we show build statuses as returned by Launchpad.

We need to map those to values defined in design doc.

https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/

Comments:

@magdalena_mir Yes.
It's partly blocked on https://trello.com/c/lBs9NcL8/186-build-detail-launchpad-api-for-publishing-state-of-build 

But we are still showing build statuses directly from LP, while we should map them to messages (and colours) from the doc.,@bartoszopka  is it still valid? If not please move to done

---
Creator: Robin Winslow
Assignees: 
Copied from Trello card: https://trello.com/c/SS7S6Q3t/149-update-build-status-names-and-colours
Labels: copy to github

earnubs:
Fixes #406


bartaz:
I'm afraid #406 is a bit more than that.

We currently only show build status of a snap ignoring publishing status (if snap was published to the store already).
LP API now has store_upload_status field that gives us this information (https://launchpad.net/+apidoc/devel.html#snap_build).

So to fully fix #406 we need to map values from LP API `buildstate` and `store_upload_status` fields to BSI list of "UI" statuses as listed in spec.

src/common/helpers/snap-builds.js
```diff
@@ -25,43 +25,116 @@
...
+      case LaunchpadStoreUploadStates.FAILED_UPLOAD:
+      case LaunchpadStoreUploadStates.FAILED_RELEASE:
+        return BuildAndPubState.PUBLISH_FAILED;
+
+      case LaunchpadStoreUploadStates.UPLOADED:
+        return BuildAndPubState.BUILT;
+    }
+  }
+
+  if (buildState === LaunchpadBuildStates.NEEDS) {
```
bartaz: This list of `if`s could be a second `switch` statement I guess, as it's simple 1:1 mapping now.

src/common/helpers/snap-builds.js
```diff
@@ -25,43 +25,116 @@
...
+  'PUBLISH_NOW':     createState('Built, publishing now', BuildStatusColours.BLUE, 'tick', 3),
+  'PUBLISH_FAILED':  createState('Built, failed to publish', BuildStatusColours.RED, 'tick', 4),
+  'PUBLISHED':       createState('Built and published', BuildStatusColours.GREEN, 'tick_filled', 5)
+};
+
+function createState(message, colour, icon, priority) {
+  return {
+    message,
+    colour,
+    icon,
```
bartaz: Not sure if in the end we want to have purely UI stuff like colour and icon here, but on the other hand we already have colour here and it being on client side it can be treated as UI code, so maybe no need to worry about this now.

Especially that we don't have icons yet, so we can postpone working on that to the point when we will be adding the icons to UI.

src/common/helpers/snap-builds.js
```diff
@@ -25,43 +25,116 @@
...
 };
 
+const BuildAndPubState = {
+  'NEVER_BUILT':     createState('Never built', BuildStatusColours.GREY, false, 8),
+  'BUILDING_SOON':   createState('Building soon', BuildStatusColours.GREY, 'ellipsis', 7),
+  'IN_PROGRESS':     createState('In progress', BuildStatusColours.BLUE, 'spinner', 2),
+  'FAILED_TO_BUILD': createState('Failed to build', BuildStatusColours.RED, 'cross', 1),
+  'WONT_PUBLISH':    createState('Built, won\'t be published', BuildStatusColours.GREEN, 'tick_outlined', 6),
+  'PUBLISH_SOON':    createState('Built, publishing soon', BuildStatusColours.GREY, 'tick', 2),
+  'PUBLISH_NOW':     createState('Built, publishing now', BuildStatusColours.BLUE, 'tick', 3),
```
bartaz: Nit picking, but `PUBLISHING_NOW` `_SOON` would be easier to understand when reading out of context.

src/common/helpers/snap-builds.js
```diff
@@ -25,43 +25,116 @@
...
-  'Uploading build': BuildStatusColours.YELLOW,
-  'Cancelling build': BuildStatusColours.RED,
-  'Cancelled build': BuildStatusColours.RED
+const LaunchpadBuildStates = {
+  'NEEDS': 'Needs building',
+  'SUCCESS': 'Successfully built',
+  'FAILED': 'Failed to build',
+  'WAIT': 'Dependency wait',
+  'CHROOT': 'Chroot problem',
+  'SUPERCEDED': 'Build for superseded Source',
```
cjwatson: It's "SUPERSEDED" in British English too.

src/common/helpers/snap-builds.js
```diff
@@ -25,43 +25,116 @@
...
+  'BUILDING': 'Currently building',
+  'UPLOAD_FAIL': 'Failed to upload',
+  'UPLOADING': 'Uploading build',
+  'CANCELLING': 'Cancelling build',
+  'CANCELLED': 'Cancelled build'
```
cjwatson: I think it'd be helpful to match the [short names used internally in Launchpad](https://git.launchpad.net/launchpad/tree/lib/lp/buildmaster/enums.py#n22), for the benefit of people reading across both codebases: `NEEDSBUILD`, `FULLYBUILT`, etc.

src/common/helpers/snap-builds.js
```diff
@@ -25,43 +25,116 @@
...
+};
+
+const LaunchpadStoreUploadStates = {
+  'UNSCHEDULED': 'Unscheduled',
+  'PENDING': 'Pending',
+  'FAILED_UPLOAD': 'Failed to upload',
+  'FAILED_RELEASE': 'Failed to release to channels',
+  'UPLOADED': 'Uploaded'
```
cjwatson: [Ditto](http://bazaar.launchpad.net/+branch/launchpad/view/head:/lib/lp/snappy/interfaces/snapbuild.py#L81).  (The Git import is a bit out of date so you get a slow Bazaar link instead.)

src/common/helpers/snap-builds.js
```diff
@@ -25,43 +25,116 @@
...
+  if (buildState === LaunchpadBuildStates.NEEDS) {
+    return BuildAndPubState.NEVER_BUILT;
+  }
+
+  if (buildState === LaunchpadBuildStates.FAILED) {
+    return BuildAndPubState.FAILED;
+  }
+
+  if (buildState === LaunchpadBuildStates.WAIT) {
+    return BuildAndPubState.SOON;
```
cjwatson: I think that's a typo for `BUILDING_SOON`.  But anyway, the dependency-wait state should just be `BuildAndPubState.FAILED`; if that state ever arises for a snap build (which it normally doesn't) then nothing is going to clear it automatically, so it might as well just be considered a failure.

src/common/helpers/snap-builds.js
```diff
@@ -25,43 +25,116 @@
...
+  }
+
+  if (buildState === LaunchpadBuildStates.BUILDING) {
+    return BuildAndPubState.IN_PROGRESS;
+  }
+
+  if (buildState === LaunchpadBuildStates.UPLOADING) {
+    return BuildAndPubState.PUBLISH_NOW;
+  }
```
cjwatson: There seem to be several other states not covered here yet.  As @bartaz said, it'd probably be clearer if it were `switch (buildState)` and then it could just cover all the possible states in their order of declaration, so that it's easy to see that none are left out.

src/common/helpers/snap-builds.js
```diff
@@ -25,43 +25,116 @@
...
+  if (buildState === LaunchpadBuildStates.WAIT) {
+    return BuildAndPubState.SOON;
+  }
+
+  if (buildState === LaunchpadBuildStates.BUILDING) {
+    return BuildAndPubState.IN_PROGRESS;
+  }
+
+  if (buildState === LaunchpadBuildStates.UPLOADING) {
+    return BuildAndPubState.PUBLISH_NOW;
```
cjwatson: Should be `IN_PROGRESS`.  (This state means that the build is still working its way through Launchpad's internal infrastructure; we haven't started to upload it to the store yet.)

src/common/helpers/snap-builds.js
cjwatson: This is either a design bug or a Launchpad bug depending on which way you look at it, but as far as I can tell there's no way for BSI to detect any state that we might describe as "publishing now": it's either pending ("publishing soon") or it's failed or it's done, and LP doesn't really expose the in-progress states between those.  We may or may not regard this as a problem, and at any rate I guess it doesn't need to be dealt with in this PR.

src/common/helpers/snap-builds.js
katiefenn: 🇬🇧🇬🇧🇬🇧

src/common/helpers/snap-builds.js
```diff
@@ -25,43 +25,116 @@
...
   GREY: 'grey'
 };
 
+const BuildAndPubState = {
+  'NEVER_BUILT':     createState('Never built', BuildStatusColours.GREY, false, 8),
+  'BUILDING_SOON':   createState('Building soon', BuildStatusColours.GREY, 'ellipsis', 7),
+  'IN_PROGRESS':     createState('In progress', BuildStatusColours.BLUE, 'spinner', 2),
+  'FAILED_TO_BUILD': createState('Failed to build', BuildStatusColours.RED, 'cross', 1),
+  'WONT_PUBLISH':    createState('Built, won\'t be published', BuildStatusColours.GREEN, 'tick_outlined', 6),
+  'PUBLISH_SOON':    createState('Built, publishing soon', BuildStatusColours.GREY, 'tick', 2),
```
katiefenn: Is there supposed to be two items with a priority of `2`?

earnubs: Good point, that is how the spec defines it however... @matthewpaulthomas

src/common/helpers/snap-builds.js
earnubs: http://i2.kym-cdn.com/photos/images/original/000/117/009/soon.jpg

earnubs: image

src/common/helpers/snap-builds.js
earnubs: Paging @matthewpaulthomas ...

src/common/helpers/snap-builds.js
```diff
@@ -25,43 +25,110 @@
...
-  'Uploading build': BuildStatusColours.YELLOW,
-  'Cancelling build': BuildStatusColours.RED,
-  'Cancelled build': BuildStatusColours.RED
+const LaunchpadBuildStates = {
+  'NEEDSBUILT': 'Needs building',
+  'FULLYBUILT': 'Successfully built',
+  'FAILEDTOBUILD': 'Failed to build',
+  'MANUALDEPWAIT': 'Dependency wait',
+  'CHROOTWAIT': 'Chroot problem',
+  'SUPERSEDED': 'Build for superseded Source',
```
earnubs: Not sure what state this should result in, can't see anything obvious in the spec.

src/common/helpers/snap-builds.js
```diff
@@ -25,43 +25,110 @@
...
+  'FAILEDTORELEASE': 'Failed to release to channels',
+  'UPLOADED': 'Uploaded'
+};
+
+function mapBuildAndPublishedStates(buildState, publishState) {
+  switch (buildState) {
+    case LaunchpadBuildStates.NEEDSBUILT:
+      return BuildAndPubState.NEVER_BUILT;
+    case LaunchpadBuildStates.FULLYBUILT:
+    case LaunchpadBuildStates.FAILEDTOUPLOAD:
```
earnubs: @cjwatson I doubt what I'm doing here. Is it the case that if we are FAILEDTOUPLOAD in buildState we'll also have it in publishState?

src/common/helpers/snap-builds.js
earnubs: Yeah, I thought I'd pencil it in as it seems like an OK place for it to go, it's a reducer so it's figuring out the state for your UI. The icon names would be better as constants or codes, but no point in finessing it now since we have no icons.

src/common/helpers/snap-builds.js
matthewpaulthomas: What does it mean? (“Build for superseded Source” makes me none the wiser.)

src/common/helpers/snap-builds.js
matthewpaulthomas: Ok, I’ve removed “Publishing now” from the spec.

src/common/helpers/snap-builds.js
cjwatson: I don't think it can happen in practice for snap builds (some of the build statuses are there because they make sense for other build types in LP), but for completeness just call it `FAILED_TO_BUILD`.

src/common/helpers/snap-builds.js
cjwatson: No.  `FAILEDTOUPLOAD` in `buildState` means that Launchpad failed to fetch the results of a build from the builder, and in that case it will never get as far as attempting to publish it to the store.  You should just put this in the "failed to build" category.

src/common/helpers/snap-builds.js
cjwatson: @matthewpaulthomas It's for cases such as "tried to build a particular source package version in a PPA but when the build got round to running it turned out that a newer source package version had already been uploaded".  Not the sort of thing that actually happens for snap builds, so it might as well just be considered a build failure if it does show up for some weird reason.

src/common/helpers/snap-builds.js
```diff
@@ -25,49 +25,190 @@
...
 
 export const BuildStatusColours = {
+  BLUE: 'blue',
   GREEN: 'green',
   YELLOW: 'yellow',
   RED: 'red',
   GREY: 'grey'
 };
 
+export const BuildStatusIcons = {
```
earnubs: Not currently used, marked as TODO in main export.

src/common/helpers/snap-builds.js
```diff
@@ -25,49 +25,190 @@
...
+  SPINNER: 'spinner',
+  CROSS: 'cross'
+};
+
+export const UserFacingState = {
+  'NEVER_BUILT': createState(
+    'Never built',
+    BuildStatusColours.GREY,
+    false,
+    8
```
earnubs: Also not currently used, marker as TODO in main export.

earnubs:
@matthewpaulthomas how would you like this to look, the longer build strings are overflowing the layout making it look broken, I don't see anything specific in the spec 

longer-build-state-strings

matthewpaulthomas:
@earnubs Currently the table column widths are all hard-coded, as 27%, 15%, 25%, 30%, and 3%. I don’t know why that is, and it seems a bit arbitrary: most obviously, 3% of the width at a given window size might be much less — or much more —than the width of a trash can button.

I think the “Latest build” column on the “My repos” page, and the equivalent “Result” column on the repo page, should be just wide enough that, in any likely font, the widest plausible combination of status+age will not collide with each other. So long as we don’t have localized UI, that is probably “Built, waiting to publish” “59 minutes ago”. And to cope with font size changes, I guess that width should be specified in em rather than %. Do you need help from a visual designer to measure it?

earnubs:
@matthewpaulthomas thanks for the clarification, I can handle the measurements.

src/common/helpers/snap-builds.js
bartaz: Good. Let's keep these TODOs around, they need to be added eventually anyway.

PR #699, opened 2017-04-28
Properly propagate store authenticated initial state

bartaz:
Even though we store macaroons in browser storage app did not properly check if user is already signed in after full page reload.

It was caused by `authenticated` flag being undefined in initial state rather then being `null` (as app code expected).

### DIFF
--- a/src/common/components/repositories-list/index.js
+++ b/src/common/components/repositories-list/index.js
@@ -23,9 +23,11 @@ const SNAP_NAME_NOT_REGISTERED_ERROR_CODE = 'snap-name-not-registered';
 export class RepositoriesListView extends Component {
 
   fetchAuthData(authStore) {
-    if (authStore.authenticated === null) {
+    // check if authenticated and hasShortNamespace haven't been set yet,
+    // so are null or undefined
+    if (authStore.authenticated == null) {
       this.props.dispatch(checkSignedIntoStore());
-    } else if (authStore.hasShortNamespace === null) {
+    } else if (authStore.hasShortNamespace == null) {
       this.props.dispatch(getAccountInfo(authStore.userName));
     }
   }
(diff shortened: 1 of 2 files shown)


src/server/handlers/universal.js
```diff
@@ -34,7 +34,7 @@ export const handleMatch = (req, res, error, redirectLocation, renderProps) => {
 
     const initialState = {
       auth: { authenticated: false },
-      authStore: {}
```
bartaz: Properly initialize `authenticated` as `null`.

src/common/components/repositories-list/index.js
```diff
@@ -23,9 +23,11 @@ const SNAP_NAME_NOT_REGISTERED_ERROR_CODE = 'snap-name-not-registered';
 export class RepositoriesListView extends Component {
 
   fetchAuthData(authStore) {
-    if (authStore.authenticated === null) {
+    // check if authenticated and hasShortNamespace haven't been set yet,
+    // so are null or undefined
+    if (authStore.authenticated == null) {
```
bartaz: Just to be safe updated the conditional checks to also accept `undefined`.

src/common/components/repositories-list/index.js
```diff
@@ -23,9 +23,13 @@ const SNAP_NAME_NOT_REGISTERED_ERROR_CODE = 'snap-name-not-registered';
 export class RepositoriesListView extends Component {
 
   fetchAuthData(authStore) {
-    if (authStore.authenticated === null) {
+    // check if authenticated and hasShortNamespace haven't been set yet,
+    // so are null or undefined
+    if (authStore.authenticated === null ||
+        typeof authStore.authenticated === 'undefined') {
```
cjwatson: This seems like an oddly verbose way to write `authStore.authenticated === undefined` (which as far as I can tell works fine).

bartaz: Ah, it's a habit from old days when direct comparison to `undefined` as not safe (value of undefined could be changed by external script). But it seems it's not the case anymore with modern browsers and strict mode. Fixed.

Issue #680, opened 2017-04-24
Organisations: Link to allow users to review their organisation permissions [3]


cjwatson:
The way the GitHub API works for showing organisation repositories is that you ask for the `read:org` scope, and then the OAuth authorisation page asks you which organisations you want the application to be able to see.  If you want to review that list later - and this is quite likely - then you need to go to your [authorised applications](https://github.com/settings/applications) page and configure it there.

The GitHub documentation [recommends](https://developer.github.com/v3/oauth/#directing-users-to-review-their-access-for-an-application) that applications provide a direct link to the GitHub settings page corresponding to their application, so that users can review this sort of thing easily.  We need to work out where to put that.

For comparison, see the [Travis profile page](https://travis-ci.org/profile), which has a bit in the bottom left saying "Is an organization missing?  Review and add your authorized organizations."

matthewpaulthomas:
Most obviously, this will be an issue when you choose “Add repos” and don’t see the repo you’re looking for. That has two variations:
1\. Your only public repos belong to this organization. So you don’t see any repos at all, so you get the empty state.
2\. You have public repos that are personal, or that belong to organizations for which you’ve authorized Snapcraft. So you see some of the repos you expect, but not all of them.

@cjwatson, what would happen if you add a repo, then on GitHub you (accidentally, perhaps) revoke build.snapcraft.io’s access to that repo’s organization, then you return to the “My repos” page? Would the repo just disappear from the list altogether? Or would/could it appear as a ghost, where you could (for example) click on it to get an explanation directing you to the GitHub settings page?

cjwatson:
I think the only realistic option would be for it to disappear altogether.  We can only know which repositories to display by way of knowing your organisation membership, and if the permission has been revoked then GitHub won't even mention that organisation to us.  The only other available data store that could tell us what repositories a particular user has added would be the snap store, but (a) that requires signing in to SSO and (b) it can only tell us about repositories where the user has registered a name, so I don't think that approach is viable here.

It may be worth considering the Travis-style option of showing you a list of organisations that BSI knows you're a member of, which would be a useful cue.

matthewpaulthomas:
[...]

PR #702, opened 2017-05-02
Add list of orgs to user info

bartaz:
Part of #680

To give user some detail help text we need details on their organisations.

This PR adds user orgs list to user object in session (and redux store).


test/routes/src/server/routes/t_github-auth.js
```diff
@@ -157,16 +158,6 @@ describe('The login route', () => {
           );
         });
 
-        it('should clear cached organization information', async () => {
```
bartaz: Cache is cleared, but because we fetch list of orgs immediately (to add to user object) cache is re-populated immediately with new data.

PR #707, opened 2017-05-02
API to refresh user orgs

bartaz:
Part of #680

API to refresh list of user orgs (in cache, session and redux store) in case access was changed on GitHub side.

### DIFF
--- a/src/server/routes/github.js
+++ b/src/server/routes/github.js
@@ -20,6 +20,7 @@ router.use('/github/user', json());
 router.get('/github/user', getUser);
 
 router.use('/github/orgs', json());
+router.get('/github/orgs', verifyToken);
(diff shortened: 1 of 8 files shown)


src/server/routes/github.js
```diff
@@ -18,6 +19,9 @@ router.post('/github/webhook', createWebhook);
 router.use('/github/user', json());
 router.get('/github/user', getUser);
 
+router.use('/github/orgs', json());
```
cjwatson: Don't you need `verifyToken` middleware here as well, for CSRF protection?

bartaz: It's just a GET, so I'm not sure if we need it, but it does clear the cache, so it wouldn't hurt to verify.

PR #713, opened 2017-05-04
Popover with info about missing repos

### LINKED ISSUE #680: Organisations: Link to allow users to review their organisation permissions [3]

bartaz:
Fixes #680

Reusing current popover as discussed [in comments](https://github.com/canonical-ols/build.snapcraft.io/issues/680#issuecomment-297953437) of #680.

screen shot 2017-05-04 at 12 25 24


src/server/helpers/config.js
```diff
@@ -12,6 +12,7 @@ const CLIENT_SIDE_WHITELIST = [
   'NODE_ENV',
   'BASE_URL',
   'GITHUB_API_ENDPOINT',
+  'GITHUB_AUTH_CLIENT_ID',
```
bartaz: CLIEND_ID is not secret (it's part of app setting URL), so we can safely expose it to client.

src/common/components/private-repos-info/index.js
```diff
@@ -1,10 +1,11 @@
-import React, { Component } from 'react';
+import React, { Component, PropTypes } from 'react';
```
earnubs: may be worth starting to use `https://www.npmjs.com/package/prop-types`

matthewpaulthomas:
That is too much stuff for a popover, but I understand the need to do it as a short-term fix.

I specced that the comma-separated list of organizations be in brackets, rather than following a colon, but I guess it doesn’t really matter which.

Apart from that, the screenshot looks fine to me.

src/common/components/private-repos-info/index.js
bartaz: We are not yet on React 15.5 that deprecates that, but that is something to consider sooner then later I guess.

bartaz:
Added #715 as follow up to not loose track of visual update for this feature.

PR #724, opened 2017-05-05
Fixes for build status not updating correctly on my repos page

### LINKED ISSUE #716: "Never built" text never updates
The never built state doesn't automatically change to building.

screen shot 2017-05-04 at 13 39 42
screen shot 2017-05-04 at 13 44 47

bartaz:
Fixes #716

### DIFF
--- a/src/common/components/build-row/index.js
+++ b/src/common/components/build-row/index.js
@@ -3,6 +3,7 @@ import { Link } from 'react-router';
 import moment from 'moment';
 
 import { Row, Data } from '../vanilla/table-interactive';
+import { BuildStatusColours } from '../../helpers/snap-builds.js';
 import BuildStatus from '../build-status';
 
 const BuildRow = (props) => {
@@ -66,7 +67,7 @@ BuildRow.propTypes = {
   buildId:  PropTypes.string,
   buildLogUrl: PropTypes.string,
   architecture: PropTypes.string,
-  colour:  PropTypes.oneOf(['green', 'yellow', 'blue', 'red', 'grey']),
+  colour: PropTypes.oneOf(Object.values(BuildStatusColours)),
   statusMessage: PropTypes.string,
   dateStarted: PropTypes.string,
   duration: PropTypes.string
--- a/src/common/components/build-row/index.js
+++ b/src/common/components/build-row/index.js
@@ -66,7 +66,7 @@ BuildRow.propTypes = {
   buildId:  PropTypes.string,
   buildLogUrl: PropTypes.string,
   architecture: PropTypes.string,
-  colour:  PropTypes.oneOf(['green', 'yellow', 'red', 'grey']),
+  colour:  PropTypes.oneOf(['green', 'yellow', 'blue', 'red', 'grey']),
   statusMessage: PropTypes.string,
   dateStarted: PropTypes.string,
   duration: PropTypes.string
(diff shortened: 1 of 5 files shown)


src/common/components/build-row/index.js
```diff
@@ -66,7 +66,7 @@ BuildRow.propTypes = {
   buildId:  PropTypes.string,
   buildLogUrl: PropTypes.string,
   architecture: PropTypes.string,
-  colour:  PropTypes.oneOf(['green', 'yellow', 'red', 'grey']),
+  colour:  PropTypes.oneOf(['green', 'yellow', 'blue', 'red', 'grey']),
```
bartaz: Blue was missing from props validation.

src/common/components/build-status/buildStatus.css
```diff
@@ -3,6 +3,11 @@
   color: $success;
 }
 
+.buildStatus.blue,
+.buildStatus.blue a {
+  color: $information;
```
bartaz: ... and from CSS.

src/common/helpers/snap-builds.js
```diff
@@ -136,7 +138,7 @@ const LaunchpadStoreUploadStates = {
 function mapBuildAndUploadStates(buildState, uploadState) {
   switch (buildState) {
     case LaunchpadBuildStates.NEEDSBUILT:
-      return UserFacingState.NEVER_BUILT;
+      return UserFacingState.BUILDING_SOON;
```
bartaz: When there is build object on LP it means that it will be building soon for BSI UI.

src/common/components/repositories-home/index.js
```diff
@@ -59,11 +49,16 @@ class RepositoriesHome extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const snapsLength = nextProps.snaps.ids.length;
-
-    if ((this.props.snaps.success !== nextProps.snaps.success)
```
bartaz: Main source of the bug.

We were relying on `success` being changed, but after refactoring it stayed `true` after first success even during subsequent fetches.

src/common/components/repositories-home/index.js
```diff
@@ -59,11 +49,16 @@ class RepositoriesHome extends Component {
...
-        || (snapsLength === 0)) {
-      this.fetchData(nextProps);
+    const { hasSnaps, snaps } = nextProps;
+    // if snaps stopped fetching
+    if ((this.props.snaps.isFetching !== snaps.isFetching) && !snaps.isFetching) {
```
bartaz: Rewritten the condition not to rely on `success` flag anymore, and always reload builds when snaps stopped fetching.

src/common/components/build-row/index.js
earnubs: What about using the exported values where they're defined? https://github.com/canonical-ols/build.snapcraft.io/blob/master/src/common/components/build-status/index.js#L43

src/common/components/build-status/buildStatus.css
earnubs: Oops.

src/common/components/repositories-home/index.js
earnubs: This seems like another reason not to like success as a feature of state per se?

bartaz: Yup, especially that we don't handle it consistently - sometimes `success` is cleared with every new fetch, sometimes not...

PR #730, opened 2017-05-08
update "install snapd" link

earnubs:
The spec has the copy linked and with no external link icon.

### DIFF
--- a/src/common/components/help/help-install-snap.js
+++ b/src/common/components/help/help-install-snap.js
@@ -25,15 +25,15 @@ export default class HelpInstallSnap extends Component {
           </p>
         }
         <p className={ styles.p }>
-          Don’t have snapd installed?  {' '}
+          (
           <a
-            className={ styles.external }
             href={ HELP_INSTALL_URL }
             rel="noreferrer noopener"
             target="_blank"
           >
-            Install it now
-          </a>.
+            Don’t have snapd installed?
+          </a>
+          )
         </p>
       </div>
     );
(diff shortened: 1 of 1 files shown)


src/common/components/help/help-install-snap.js
```diff
@@ -25,15 +25,15 @@ export default class HelpInstallSnap extends Component {
           </p>
         }
         <p className={ styles.p }>
-          Don’t have snapd installed?  {' '}
+          (
```
bartaz: Is this needed?

src/common/components/help/help-install-snap.js
```diff
@@ -25,15 +25,15 @@ export default class HelpInstallSnap extends Component {
...
           <a
-            className={ styles.external }
```
bartaz: Spec is not a visual design, so I think we should keep external link icon - this is a link outside of BSI (and opens in new tab), to it makes sense to keep the icon here.

src/common/components/help/help-install-snap.js
earnubs: I don't understand, the spec has `(Don’t have snapd installed?)` with the text linked and the parens unlinked.

src/common/components/help/help-install-snap.js
earnubs: OK.

src/common/components/help/help-install-snap.js
bartaz: My parser got it wrong ;)

Issue #655, opened 2017-04-21
Next steps: update your install instructions [1]


evandandrea:
In addition to guiding the developer to promote to a stable channel we should also suggest they add the something like the following to their install instructions:

```
Install snap_name in seconds on Linux (Ubuntu and others) with:
  sudo snap install snap_name

Don’t have snapd installed? [Install it now](https://snapcraft.io/docs/core/install).
```

We could also give buttons to post this message on Facebook, Twitter, and Google+.

(It'd be nice to give them options to submit to Reddit and Hacker News, but we'd need a URL for the story and I don't think the builds page is what we want to present end users.)

- [x] update text to match spec ( #730 )
- [x] update layout to match spec (2 columns of copy, one for testing, one for install)
- [x] add copy to clipboard icon and behaviour (#732)
- [x] add Twitter share button
- [ ] add share URL buttons (facebook, g+, etc)

matthewpaulthomas:
The instructions for (reposting elsewhere for) end users to install the stable build are very similar to the instructions for testers to install the latest build. So I think I’ll put them alongside each other, to help visitors understand how+why they differ.

matthewpaulthomas:
[Spec updated](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.dwvfeg899834).

bartaz:
@matthewpaulthomas @nottrobin 
Do we have social icons we want to use there (with hover states)? I see some on ubuntu.com footer, but not sure if there are the ones (in these sizes) we want to have here?

bartaz:
@matthewpaulthomas @nottrobin 
Do we have social icons we want to use there (with hover states)? I see some on ubuntu.com footer, but not sure if there are the ones (in these sizes) we want to have here?

evandandrea:
@bartaz I can't speak to sizes, but there's the svg in https://github.com/canonical-ols/build.snapcraft.io/issues/654#issuecomment-298949804

matthewpaulthomas:
I think those same icons would work at the same sizes.

Feel free to split out “Copy to clipboard” to a separate issue report.

bartaz:
@evandandrea @matthewpaulthomas 
There is still an issue that was mentioned by @earnubs  on stand-up that it seems FB and G+ don't allow sharing/posting from API without link to share. And main descriptions suggests we don't want to share repo builds page URL.

[...]

PR #732, opened 2017-05-08
add copy to clipboard behaviour

earnubs:
spec stipulates that a copy-to-clipboard icon exist below cli instructions, part of #655


magdalena-mirowicz:
please pause this work for now

magdalena-mirowicz:
@earnubs please ignore me, it's ok to work on

src/common/components/share/index.js
```diff
@@ -0,0 +1,28 @@
+import React, { Component, PropTypes } from 'react';
+import Clipboard from 'clipboard';
+
+import styles from './styles.css';
+
+const COPY_TO_CLIPBOARD_CLASS = styles.clipboard;
+
+export class CopyToClipboard extends Component {
+  componentDidMount() {
+    new Clipboard(`.${COPY_TO_CLIPBOARD_CLASS}`);
```
bartaz: Will it work just fine if there are two instances of that on a page?

So I suppose `new Clipboard` would be called 2 times with same class name?

earnubs: Good point, I was looking at the spec and expecting only one of these per page.

earnubs:
image

src/common/components/share/index.js
```diff
@@ -0,0 +1,50 @@
...
+      this.initClipboard();
+    }
+  }
+
+  componentWillUnmount() {
+    this.clipboard && this.clipboard.destroy();
+  }
+
+  initClipboard() {
+    this.clipboard = new Clipboard(this.copyBtn);
```
earnubs: Only add the eventListener to a single component instance, so multiple instances of this component can live with each other.

earnubs:
@bartaz have another look, addressing your comment by using a ref to the component as the target of the clipboard lib rather than a selector, and cleaning up on unmount.

src/common/components/share/index.js
bartaz: Ok, thanks.

That looks much more encapsulated right now.

Issue #675, opened 2017-04-21
Adding snap build status badge to your README [2]


nottrobin:
Create a [GitHub badge](https://shields.io/) to show the build status of a repository, and provide helper text to help add this to the README of a repository. This helper text should probably be shown on the build status page, next to the instructions for installing the snap from the CLI.

We should also consider telling the user how to add "snap install" instructions to their README at the same time.

---

TODO:
- [x] generate badge images for all the build statues [#727]
- [x] handler for checking latest status for given repo [#734]
- [x] handler for returning proper badge image for given repo [#734]
- [x] display badge on repo page ([spec](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.vhi6nalioput))
- [ ] generate Markdown for badge image with link
- [ ] copy Markdown to clipboard on button click

nottrobin:
The second part of this is covered by #655

matthewpaulthomas:
It might be a little annoying if build.snapcraft.io continued giving you this suggestion, when you had already followed it days or years ago. Would it make sense for build.snapcraft.io to scan the README, and suggest adding the badge only if you hadn’t already? (That might not make sense if, for example, it’s plausible for you to put the graphic in places other than the README too.)

matthewpaulthomas:
[Spec updated](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.vhi6nalioput). Instead of making the instructions dependent on whether you’ve already added the badge, I’ve taken a less finicky approach by putting an icon-only Copy button next to the badge itself.

bartaz:
For implementation: https://github.com/badges/gh-badges

bartaz:
@matthewpaulthomas Spec defines two separate links for two parts of a badge. I've never seen any badge do that and it would be pretty hard to achieve that in Markdown as we are simply dealing with a single image wrapped in a link.

Most badges seem to link to a "project" page listing all the builds (with latest on top) - so that would be our Repo page.

matthewpaulthomas:
@bartaz https://github.com/badges/gh-badges was last modified 3 years ago, and links to http://img.shields.io/, which redirects to http://shields.io/, which says “`?link=http://left&link=http://right` Specify what clicking on the left/right of a badge should do”.

[...]

PR #734, opened 2017-05-08
Handler to serve badge image based on repo build status

bartaz:
Part of #675


src/server/server.js
```diff
@@ -48,6 +48,10 @@ app.use(expressWinston.logger({
   }
 }));
 app.use(helmet());
+
+// keep it before setting up session, as it doesn't need that
+app.use('/', routes.badge);
```
bartaz: Badge route doesn't need session cookies or anything...

test/routes/src/server/routes/t_badge.js
```diff
@@ -0,0 +1,131 @@
...
+    });
+
+    afterEach(() => {
+      lpApi.done();
+      nock.cleanAll();
+    });
+
+    context('when there are builds for given repo', () => {
+      beforeEach(() => {
+        lpApi = lpApi.get(lp_builds_path)
```
bartaz: It's a bit painful to mock all those API requests even when they are not directly used (only accessed via some exported `internal-` functions. Especially that LP API responses have very specific structure...

I don't know if there is any easier way to mock those. Other option would be to proxyquire methods that call these APIs, but proxyquire has a troublesome setup on its own...

src/server/handlers/launchpad.js
```diff
@@ -653,6 +653,12 @@ export const authorizeSnap = async (req, res) => {
   }
 };
 
+export async function internalGetSnapBuilds(snap, start = 0, size = 10) {
```
katiefenn: Do the parameter defaults make the default values on lines 665 and 666 redundant?

bartaz: I guess so, good catch. Thanks.

test/routes/src/server/routes/t_badge.js
katiefenn: I wonder whether we're actually getting the full value from these tests because they are a weird compromise of unit test and integration test. If we tested this functionality using unit tests, we wouldn't have to mock as many services. If we tested this functionality using integration tests, we would be able to run these tests against a staging or production instance. As it stands, we're getting the worst of both worlds.

bartaz: Node module imports/requires makes it hard to test module as an independent 'unit', because in its source we define very specific dependencies. 

The simplest way around that in our current stack would probably getting used to using proxyguire more and simply mock/stub as much external dependencies as possible.

Issue #646, opened 2017-04-21
Organisations: Prevent other organisation members from disabling snap auto building [2]


nottrobin:
When the user wasn't the one that setup an organisation snap, they can't disable it.

matthewpaulthomas:
Straw person proposal: If you aren’t the person who “set up” the repo, we could still show you the Delete icon. But when you click it, instead of getting the Cancel and Delete buttons, we’d tell you who set it up, and suggest that you go talk to them. Does that seem reasonable?

Question: What do we mean by “set up”, exactly? Is this something we should do for every organization repo, _or_ just for those repos that have a registered name?

evandandrea:
Your proposal seems reasonable to me.

I think to keep things simple, for now let's make it for every organisation repo. If Robin adds CanonicalLtd/snappy-docs, then it's read-only on Carla's dashboard. She cannot register a name for it even if Robin hasn't. She can click the delete icon, but doing so pops up your suggested text.

cjwatson:
An important consideration that I discovered after talking to Natalia last week is that, in the general case, the store won't tell us who is the publisher of a particular snap, for privacy reasons.  (The new design for organisations in BSI per Mark is that the store doesn't (yet?) have any idea about the organisational relationships, so we can't say that the store will expose this information only to other admins; and I don't think there's any way that we can work around this by remembering the publisher in BSI, because it might be reassigned later on the store side.)

In the case where a snap has reached the stable channel, we can get this information in the same way that `snap info` does.  So it sounds like we'll be able to show the publisher in some cases, but not in every case.

Can we accommodate this in the design?

cjwatson:
@evandandrea Perhaps counterintuitively, your "to keep it simple" proposal actually makes things more complicated.  There's currently no difference in the data we store for a repository until the point when the name is registered (and no very obvious place to store data that would differ).  It would be significantly easier to implement if we decided that this would just be for repositories with a registered name.

nottrobin:
@cjwatson if we retrieve information in the same way that `snap info` does, won't we only get the publisher's namespace, rather than their name? (e.g. the `upload-assets` snap's "publisher" is `canonicalwebteam`, but the SSO username is actually `cwebteam`) Is that of any use to us?

Are we not even able to tell if you own a particular snap if you are logged in to SSO?

evandandrea:
[...]

PR #735, opened 2017-05-09
Name ownership reducer

bartaz:
Part of #646

Currently we only check name ownership for name in snapcraft.yaml - and we keep this status together with snapcraft.yaml data in redux store.

But for #646 we also need to check name ownership for registered names of snaps, so it would be better to have a separate store for keeping name ownership status per name.

This PR refactors current state of the app (name ownership of snapcraft.yaml name) into new name ownership reducer, where in future also register names ownership may be kept.

### DIFF
--- a/src/common/components/repository-row/dropdowns/remove-repo-dropdown.js
+++ b/src/common/components/repository-row/dropdowns/remove-repo-dropdown.js
@@ -38,6 +38,9 @@ const RemoveRepoDropdown = (props) => {
 };
 
 RemoveRepoDropdown.propTypes = {
+  authStore: PropTypes.shape({
+    authenticated: PropTypes.bool
+  }).isRequired,
   message: PropTypes.string.isRequired,
   onRemoveClick: PropTypes.func.isRequired,
   onCancelClick: PropTypes.func.isRequired
--- a/src/common/components/repository-row/dropdowns/remove-repo-dropdown.js
+++ b/src/common/components/repository-row/dropdowns/remove-repo-dropdown.js
@@ -38,9 +38,6 @@ const RemoveRepoDropdown = (props) => {
 };
 
 RemoveRepoDropdown.propTypes = {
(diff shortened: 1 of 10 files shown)


src/common/components/repository-row/dropdowns/remove-repo-dropdown.js
```diff
@@ -38,6 +38,9 @@ const RemoveRepoDropdown = (props) => {
 };
 
 RemoveRepoDropdown.propTypes = {
+  authStore: PropTypes.shape({
+    authenticated: PropTypes.bool
+  }).isRequired,
```
cjwatson: Leftover from debugging or something?  Seems to be unused in this file.

src/common/reducers/entities/snap.js
```diff
@@ -10,32 +10,32 @@ export default function snap(state={}, action) {
...
+    //   };
+    // case RegisterNameActionTypes.CHECK_NAME_OWNERSHIP_ERROR:
+    //   return {
+    //     ...state,
+    //     snapcraftData: {
+    //       ...state.snapcraftData,
+    //       isFetching: false,
+    //       nameOwnershipStatus: null
+    //     }
+    //   };
```
cjwatson: Either fix or remove (I'm guessing remove since it's now in `name-ownership.js`) - let's not leave commented-out code littered over the codebase.

src/common/components/repository-row/dropdowns/remove-repo-dropdown.js
bartaz: Ah... 'leftover' from not yet implemented check in RemoveRepoDropdown ;)

src/common/reducers/entities/snap.js
bartaz: Ah... Rushed up with a commit, will remove of course as that's the part that was moved.

Issue #706, opened 2017-05-02
Metric of repos with a snapcraft.yaml


evandandrea:
We have registered name, but we don't yet have a metric of repos with a snapcraft.yaml. Can we please add this to bsi_developer_uptake?

cjwatson:
Any new metric has to have both a way to work out the current value as an initialiser, and a way to keep it up to date when things change.  The update procedure is fairly easy in this case (we can do it from the GitHub webhook, plus a bit of code to handle newly-added repositories), but initialising is harder since we have to loop over all the GitHub repositories we know about while being careful not to run into the rate limit.  We may therefore have to loop over them in smaller chunks until we have a complete record in the database.  (At the moment we only store this information in memcached, which is good enough for the web UI but not good enough for metrics.)  We then need to attribute that to the users who set up the repositories in question in BSI, and combined with the chunking requirement that means that we need to start tracking repositories individually for metrics purposes in a way that we don't currently do.

I think it may make sense to do this after #621.  Fixing that is probably going to require adding a `Repository` table and then separately keeping track of which user added which repository in a link table.  Having done that work, it will be much easier to do this.

solazio:
Not relevant for snapcraft.io, closing

PR #737, opened 2017-05-09
Attribute organization repo activity to registrant

### LINKED ISSUE #621: Organisations: Tidy up attribution metrics to support organisation repositories [2]
Currently the GitHub username is inferred from the repository name (<user>/<repository>) in some places. This will no longer work once we are dealing with organisation repositoriess. This should be fixed so attribution metrics continue to reflect the current user.

cjwatson:
Keep track of who registered each repository with build.snapcraft.io in
a new Repository table, and attribute `builds_requested` and
`builds_released` activity to that user.  Catch up with existing
repositories via the `/api/launchpad/snaps/list` handler.

This isn't necessarily perfect since the publisher can be reassigned in
the store without us noticing, but this is tolerable since it's only for
metrics.

The `Repository.snapcraft_name` and `Repository.store_name` columns
aren't immediately needed for this, but will be useful for future work
such as #706.

Fixes #621.

### DIFF
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -350,19 +350,9 @@ export const newSnap = async (req, res) => {
           { transacting: trx }
         );
       }
-      const repositoryModel = db.model('Repository');
-      let dbRepository = await repositoryModel
-        .where({ owner, name })
-        .fetch({ transacting: trx });
-      if (dbRepository === null) {
-        dbRepository = repositoryModel.forge({ owner, name });
-      }
-      dbRepository.set({
-        registrant_id: dbUser && dbUser.get('id')
-      });
-      if (dbRepository.hasChanged()) {
-        await dbRepository.save({}, { transacting: trx });
-      }
+      await db.model('Repository').addOrUpdate(
+        { owner, name }, { registrant: dbUser }, { transacting: trx }
+      );
 
       const result = await requestNewSnap(repositoryUrl);
       // as new snap is created we need to clear list of snaps from cache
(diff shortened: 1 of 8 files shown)


src/server/handlers/launchpad.js
```diff
@@ -545,9 +561,46 @@ export const findSnaps = async (req, res) => {
       );
       return { ...snap, snapcraft_data: snapcraftData };
     }));
-    if (req.session.user) {
-      await initializeMetrics(req.session.user.id, snaps);
-    }
+    await db.transaction(async (trx) => {
```
bartaz: I wonder if it would make sense to extract metrics/db related code our of API response handlers.

In this particular example the 'keep the database up to date' code, which is not directly related to 'findSnap' functionality takes half of the code of this method making it quite hard to read.

It seems that likely we will be adding more and more metrics related code in different places, so it may be a good point to start thinking of some good convention to do that.

src/server/handlers/launchpad.js
```diff
@@ -612,19 +665,37 @@ export const authorizeSnap = async (req, res) => {
...
     await db.transaction(async (trx) => {
+      let dbUser = null;
       if (req.session && req.session.user) {
-        await db.model('GitHubUser').incrementMetric(
+        dbUser = await db.model('GitHubUser').incrementMetric(
           { github_id: req.session.user.id }, 'names_registered', 1,
           { transacting: trx }
         );
       }
+      const repositoryModel = db.model('Repository');
```
bartaz: Similar example here. 

As long as the metrics related code was just single call of `await db.model('GitHubUser').incrementMetric(...)` it was pretty short and didn't confuse when reading `authorizeSnap` handler.
Now metrics part grows to ~20 lines of code making it harder to spot what part is really about authorizing snap and which is keeping db up to date.

src/server/handlers/launchpad.js
cjwatson: Thanks.  I've pushed a change to refactor this into what should be a more readable form.

PR #741, opened 2017-05-10
Simulate a missing snap/snapcraft.yaml

sparkiegeek:
The mock GitHub service fails to respond sensibly to a request for `snap/snapcraft.yaml` when it can just 404 instead. This makes the fake simulate a `snapcraft.yaml` but a missing `snap/snapcraft.yaml`.


earnubs:
FYI I think we're going to drop the mock service from the code base.

PR #742, opened 2017-05-10
Get registered names from account info response

bartaz:
As suggested by @cjwatson `/store/account` API response (that we already make after user signs in) already contains list of names registered by given user.

So we can pre-populate list of names owned by given user with this information before we attempt to verify ownership of any other name.

### DIFF
--- a/src/common/actions/auth-store.js
+++ b/src/common/actions/auth-store.js
@@ -8,7 +8,7 @@ import url from 'url';
 import { checkStatus, getError } from '../helpers/api';
 import { conf } from '../helpers/config';
 import { getCaveats } from '../helpers/macaroons';
-import { getPackageUploadRequestMacaroon } from './register-name';
+import { STORE_SERIES, getPackageUploadRequestMacaroon } from './register-name';
 
 const BASE_URL = conf.get('BASE_URL');
 
@@ -228,7 +228,14 @@ async function fetchAccountInfo(root, discharge) {
     headers: { 'Accept': 'application/json' }
   });
   if (response.status >= 200 && response.status < 300) {
-    return { signedAgreement: true, hasShortNamespace: true };
+    const json = await response.json();
+    let registeredNames = null;
+
+    if (json.snaps && json.snaps[STORE_SERIES]) {
+      registeredNames = Object.keys(json.snaps[STORE_SERIES]);
+    }
+
+    return { signedAgreement: true, hasShortNamespace: true, registeredNames };
   } else {
     const json = await response.json();
     const payload = json.error_list ? json.error_list[0] : json;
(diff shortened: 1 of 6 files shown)


src/common/components/repository-row/index.js
```diff
@@ -201,14 +201,23 @@ export class RepositoryRowView extends Component {
...
-      this.checkNameOwnership(nextProps.nameOwnership, nextSnapcraftData.name);
-    }
-
-    if (nextProps.snap.storeName) {
-      this.checkNameOwnership(nextProps.nameOwnership, nextProps.snap.storeName);
+    // only attempt checking name ownership if user is authenticated
+    // and we are not fetching any auth data (because fetched account info will
+    // contain names already registered by given user)
+    if (this.props.authStore.authenticated &&
+        !nextProps.authStore.isFetching && !this.props.authStore.isFetching) {
```
bartaz: If authStore data is fetching (possibly waiting for `/store/account` response) we shouldn't attempt any name check, because account info didn't yet pre-populate the list.

src/common/actions/auth-store.js
```diff
@@ -228,7 +228,14 @@ async function fetchAccountInfo(root, discharge) {
     headers: { 'Accept': 'application/json' }
   });
   if (response.status >= 200 && response.status < 300) {
-    return { signedAgreement: true, hasShortNamespace: true };
+    const json = await response.json();
```
cjwatson: You could save a tiny bit of duplication by moving this above the `if` block, since it's the first line in both branches.

test/unit/src/common/actions/t_auth-store.js
```diff
@@ -16,13 +16,16 @@ const getPackageUploadRequestMacaroon = proxyquire.noCallThru().load(
   '../../../../../src/common/actions/register-name',
   { 'localforage': localForageStub }
 ).getPackageUploadRequestMacaroon;
+
+const STORE_SERIES = '16';
```
cjwatson: You could import this from `src/common/actions/register-name` instead by changing the code above this slightly:

    const registerNameModule = proxyquire.noCallThru().load(
      '../../../../../src/common/actions/register-name',
      { 'localforage': localForageStub }
    );
    const { getPackageUploadRequestMacaroon, STORE_SERIES } = registerNameModule;

earnubs: Grab this from the export in `src/common/actions/register-name.js`?

src/common/components/repository-row/index.js
earnubs: this isFetching checkout could be pulled into a small function, for clarity

PR #747, opened 2017-05-11
Prevent other users from removing registered snap

### LINKED ISSUE #646: Organisations: Prevent other organisation members from disabling snap auto building [2]

bartaz:
Fixes #646

Final UI part of #646. If repo has registered name user needs to sign in before removing it.
Only user who registered the name is able to remove given repo.

screen shot 2017-05-11 at 14 26 36
screen shot 2017-05-11 at 14 27 32

### DIFF
--- a/src/common/components/repository-row/dropdowns/remove-repo-dropdown.js
+++ b/src/common/components/repository-row/dropdowns/remove-repo-dropdown.js
@@ -6,14 +6,37 @@ import { WarningIcon } from '../icons';
 
 import styles from './dropdowns.css';
 
+const getRemoveWarningMessage = (latestBuild, registeredName) => {
+  let warningText;
+  if (latestBuild) {
+    warningText = (
+      'Removing this repo will delete all its builds and build logs.'
+    );
+  } else {
+    warningText = (
+      'Are you sure you want to remove this repo from the list?'
+    );
+  }
+  if (registeredName !== null) {
+    warningText += ' The name will remain registered.';
+  }
+  // XXX cjwatson 2017-02-28: Once we can get hold of published states for
+  // builds, we should also implement this design requirement:
+  //   Separately, if any build has been published, the text should end
+  //   with:
+  //     Published builds will remain published.
+
+  return warningText;
+};
+
 const RemoveRepoDropdown = (props) => {
-  const { message, onCancelClick, onRemoveClick } = props;
+  const { latestBuild, registeredName, onCancelClick, onRemoveClick } = props;
 
   return (
     <Dropdown>
       <Row>
         <Data col="100">
-          <WarningIcon /> { message }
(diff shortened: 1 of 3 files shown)


src/common/components/repository-row/dropdowns/remove-repo-dropdown.js
```diff
@@ -6,40 +6,102 @@ import { WarningIcon } from '../icons';
 
 import styles from './dropdowns.css';
 
+const getRemoveWarningMessage = (isBuilt, registeredName) => {
```
bartaz: Moved from `RepositoryRow` as it makes much more sense now to keep all remove-related logic in one place.

src/common/components/repository-row/index.js
```diff
@@ -281,7 +281,7 @@ export class RepositoryRowView extends Component {
 
     const registeredName = snap.storeName;
 
-    const hasBuilt = !!(latestBuild && snap.snapcraftData);
+    const isBuilt = !!(latestBuild && snap.snapcraftData);
```
earnubs: I think this is my fault, but would this be clearer as:

```
(latestBuild && typeof snap.snapcraftData !== 'undefined')
```

bartaz: I think `snapcraftData` is (or may be) `null` if snapcraft.yaml is not present.
So I guess `(latestBuild && typeof snap.snapcraftData !== null)`... unless we need to support both `null` and `undefined`...

earnubs: Does it need to be `null`, could it be reset to `undefined`?

bartaz: I guess it comes all way down from backend handler that reads snapcraft.yaml from GH: https://github.com/canonical-ols/build.snapcraft.io/blob/master/src/server/handlers/github.js#L173

So it seems it's never `undefined`.

test/unit/src/common/components/repository-row/dropdowns/t_remove-repo-dropdown.js
```diff
@@ -0,0 +1,101 @@
...
+
+  let view;
+
+  beforeEach(function() {
+    view = shallow(<RemoveRepoDropdown { ...props }/>);
+  });
+
+  context('when name is not registered', function() {
+    beforeEach(() => {
+      view.setProps({ registeredName: null });
```
katiefenn: I wish I knew about `setProps` earlier... 😅

katiefenn:
+1

test/unit/src/common/components/repository-row/dropdowns/t_remove-repo-dropdown.js
bartaz: Me too ;)

Issue #686, opened 2017-04-25
Update dropdown layout My Repos table


bartaz:
Layout of My Repos dropdowns should be updated.

<img width="595" alt="screen shot 2017-04-25 at 13 26 19" src="https://cloud.githubusercontent.com/assets/83575/25383141/02c1e010-29bb-11e7-829c-3407c916be35.png">
<img width="602" alt="screen shot 2017-04-25 at 13 26 34" src="https://cloud.githubusercontent.com/assets/83575/25383140/02c1677a-29bb-11e7-8c8a-619cfe5cc069.png">
<img width="591" alt="screen shot 2017-04-25 at 13 26 42" src="https://cloud.githubusercontent.com/assets/83575/25383142/02c7ad92-29bb-11e7-9d3d-5c9a67f5dca5.png">

magdalena-mirowicz:
@bartaz could you add estimate in brackets please? thanks

matthewpaulthomas:
I’m really sorry, but this is the first time I’ve seen these mockups, and the layout is quite strange in all of them. Where did they come from?

They improve on the mockups currently in the spec in one way: they show how to visually connect the clicked cell and the revealed panel (implemented in #321).

However, unless the layout changed depending on window width, they would all work poorly on phones, tablets, and narrow windows, for various reasons:
- the first two, because they assume that the buttons can fit alongside the form;
- the third, because it assumes that the warning messages can fit alongside the form.

Least importantly, I’m not a fan of putting a divider line between the form and the buttons (#754), but these layouts are incompatible with doing that.

@bartaz, would it be okay to change just the button style for now, and leave the layout for later?

bartaz:
@matthewpaulthomas Thanks for comment.

When I started to implement the layout changes I also noticed some inconsistencies and edge cases with different text lengths, so it would actually be really helpful to only do button styles now and postpone layout changes till later once they are improved.

bartaz:
It's probably not going to be implemented in this form.

If we will update the table styles we will more likely use the vanilla pattern styles.

PR #775, opened 2017-05-16
Updated dropdown cancel button style

bartaz:
Part of #686

As discussed there in comments we focus on updating button style only here, leaving layout updates to be confirmed and implemented later.

### DIFF
--- a/src/common/components/repository-row/dropdowns/register-name-dropdown.js
+++ b/src/common/components/repository-row/dropdowns/register-name-dropdown.js
@@ -231,9 +231,13 @@ const ActionButtons = (props) => {
 
   return (
     <div className={ styles.buttonRow }>
-      <a onClick={onCancelClick} className={ styles.cancel }>
+      <Button
+        appearance="base"
+        onClick={onCancelClick}
+      >
         Cancel
-      </a>
+      </Button>
+      {' '}
       <Button
         appearance="positive"
         disabled={actionDisabled}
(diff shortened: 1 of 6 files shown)


src/common/components/vanilla/button/button.css
```diff
@@ -1,11 +1,4 @@
-$primary-button-color: $ubuntu-orange;
-$primary-button-color-hover: #d44615;
-$primary-button-text-color: #FFF;
-
-$secondary-button-color: #FFF;
```
bartaz: Primary and secondary are not used anymore and are not part of Vanilla.

src/common/components/repository-row/dropdowns/register-name-dropdown.js
```diff
@@ -231,9 +231,13 @@ const ActionButtons = (props) => {
...
     <div className={ styles.buttonRow }>
-      <a onClick={onCancelClick} className={ styles.cancel }>
+      <Button
+        appearance="base"
+        onClick={onCancelClick}
+      >
         Cancel
-      </a>
+      </Button>
+      {' '}
```
earnubs: This makes me think of the very complex button rules we had for button groups (in a pre Vanilla CSS lib), but you could obviously control this with `button + button:not(:last-child)` kind of thing.

src/common/components/vanilla/button/button.css
```diff
@@ -44,53 +38,38 @@ $neutral-border-color: $dark-grey;
...
-  border: 1px solid $secondary-border-color;
 }
 
-.secondary:active,
-.secondary:focus,
-.secondary:hover {
-  background-color: $secondary-button-color-hover;
+.base:active,
+.base:focus,
+.base:hover {
```
earnubs: In order: link, visited, hover, active?

bartaz: I should dust out the "LoVe HAte" mnemonic ;)
I think it matters when there are different styles applied (in separate selectors).

PR #776, opened 2017-05-16
add tweet button to build page

earnubs:


src/common/components/share/twitter.js
```diff
@@ -0,0 +1,23 @@
+import React, { Component, PropTypes } from 'react';
+
+import styles from './styles.css';
+
+export class Tweet extends Component {
+  render() {
+    const { name } = this.props;
+    const text = encodeURIComponent(
+      `Install ${name} in seconds on Linux OSes:\nsudo snap install ${name}\n\n(Don’t have snapd? https://snapcraft.io/docs/core/install)`
```
bartaz: I know that right now we only use this to share install instructions, but if 'CopyToClipboard' gets text as param, maybe it would be good to make `Tweet` work the same way as well?

Also... are we sure that's the message we want to tweet? With the only link being to https://snapcraft.io/docs/core/install rather than any snap/build related page?

matthewpaulthomas: It is indeed awkward that the only link is not to the actual snap. But that’s pretty much unavoidable, until the snap store has a Web directory. As soon as it does, we can replace both the command line _and_ the snapd link with a directory link. The directory page can take care of people who don’t have snapd installed.

src/common/components/help/help-install-snap.js
```diff
@@ -41,6 +41,11 @@ export default class HelpInstallSnap extends Component {
           <CopyToClipboard
             copyme={ command }
           />
+          <Tweet
+            text={ `Install ${name} in seconds on Linux OSes:\n`
```
earnubs: Per review, pass the text to the component, which makes more sense in terms of editing content in the parent etc, as well as making it more flexible.

bartaz: Nit picking, but could we move the text building out of JSX into a variable?

src/common/components/share/styles.css
```diff
@@ -7,6 +7,7 @@
   background-color: #666;
   background-position: 50% 50%;
   background-repeat: no-repeat;
+  margin-right:6px;
```
bartaz: `CopyToClipboard` is reused in different context and this additional margin unnecessarily move it a bit.

bartaz:
Looks good codewise, but testing it made me realise that we show it under 'To test this snap on your PC or cloud instance' rather than under 'To install the latest stable version' (which we cannot show because we don't know if stable version exists).

So I'm not sure if we want to have Twitter share (or even copy) here, because there is not much point in sharing install instructions that may not work.

earnubs:
2:27 PM <beowulf> bartaz: i'm not entirely clear on what share button goes where, but from the spec it looks like the twitter button has to wait until we have help for "install latest stable", is that what you meant?
2:28 PM <bartaz> yes, because we are sharing install instructions for stable snap, and we don't know if it exists right now
2:29 PM <bartaz> also, from comments on the spec it seems that it still needs some design: https://docs.google.com/a/canonical.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit?disco=AAAABKatpCM
2:30 PM <bartaz> so to my understanding this issue is blocked on a) being able to tell if snap was released in stable channel b) clarifications how the info about installing it and promoting the install should work/look like

src/common/components/help/help-install-snap.js
```diff
@@ -12,6 +12,12 @@ export default class HelpInstallSnap extends Component {
     const revOption = revision ? `--revision=${ revision }` : '';
     const command = children || `sudo snap install --edge ${name} ${revOption}`;
 
+    /**
+    const tweet = `Install ${name} in seconds on Linux OSes:\n`
```
bartaz: It would be nice to have a reference to #655 in the comment to remember why it's commented out.

matthewpaulthomas:
Is the title of this PR a typo? There should not be a tweet button on the build page, it should be on the repo page.

earnubs:
The title of the PR has no influence on what happens in the PR; but no, the button doesn't and won't appear on the build page, or any page with this PR.

Issue #780, opened 2017-05-17
Tell the difference between new user and user with no repos


bartaz:
To fully support application flow as in spec in relation to #665, #657, #755 we need to be able to tell the difference between first time user (who doesn't have any repos yet) and user who did have some repos, but now does not (because they removed it or their org repos were removed by other admins).

Specifically it applies to "...and none added in the past" part of the spec.

@matthewpaulthomas For implementation we need to clarify what does "...and none added in the past" mean in terms of organization repos.

If user didn't add any repos by himself, but did have some organization repos in "My repos" (added by someone else) that are now removed - should such user be treated as first time user (who didn't add any repos), or should they be treated as user who had some repos "added in the past"?

bartaz:
As mentioned before by @cjwatson - implementing this feature means that users will not be able to get back to 'first time user' state. Also we - for testing - will not be able to get back to 'first time user' state without creating new GH account (or deleting contents of metrics db).

Currently users who remove all of their repos from BSI are treated in the same way as if they were new users.
If we implement the check if user added any snaps in the past we persist the data about the user without being able to delete it (unless we remove it from metrics db).

We currently don't have any 'delete account' functionality in BSI, so there will be no way for any user to get back to 'first time user' state once they sign in and add any repo.

@evandandrea @matthewpaulthomas Can you confirm that that's what we want to have implemented?

matthewpaulthomas:
@bartaz Yes it is. “Get back to ‘first time user’ state” is not a goal for anyone except build.snapcraft.io developers doing QA. So the cost is that QA will be more difficult until account deletion is implemented. The benefit is that if, for example, someone was using build.snapcraft.io at a company for years but they’ve just been laid off, when they return to the site to use it for a personal project, we won’t be acting like we’ve never seen them before.

Issue #666, opened 2017-04-21
"My repos" page: Progressively load new rows [2]


nottrobin:
When a new repo is added (from the "Add repos" list), currently there is a delay before the new row is added to the table, which may confuse the user.

To avoid this, the following solution was discussed in the workshop:

- Show the row immediately, filling in at least the repository name, leaving other columns blank
- Fill in the other columns when the data arrives from the server
- If the data takes more than 2 seconds to arrive, show placeholder text (e.g. "Scanning")

The exact text is still to be designed.

solazio:
Not relevant for snapcraft.io

PR #782, opened 2017-05-17
Empty state for "My repos"

### LINKED ISSUE #657: Organisations: "My repos" dashboard: Empty state [1]
As discussed in the workshop, we should design an "empty" state for the "My repos" dashboard page (`https://build.snapcraft.io/user/{username}`) - which anyone will see if they have no added repositories, unless it's their first time.

This page will be shown when a user:

- Hits "cancel" on the "Add repositories" list, with no added reposories
- Deletes their last repo from BSI
- Has their last organisation repository deleted by someone else

Crucially, an *existing* user returning to the site should be shown this empty page rather than the "Add repositories" list. (@carlaberkers please confirm this last point).

bartaz:
Fixes #657

Empty state of My repos table:

screen shot 2017-05-24 at 14 35 47

Now that #780 is implemented we can expose to client app data if user has ever added any snap, use it to properly show first time heading above empty My repos list.

### DIFF
--- a/src/common/actions/repository.js
+++ b/src/common/actions/repository.js
@@ -23,7 +23,7 @@ export const REPO_RESET = 'REPO_RESET';
 export function addRepos(repos, owner) {
   return async (dispatch) => {
     try {
-      await Promise.all(repos.map( async (repo) => {
+      await Promise.all(repos.map(async (repo) => {
         await dispatch(createWebhook(repo));
         return dispatch(addRepo(repo));
       }));
--- a/src/common/actions/repository.js
+++ b/src/common/actions/repository.js
@@ -2,6 +2,7 @@ import { push } from 'react-router-redux';
 
 import { APICompatibleError } from '../helpers/api';
 import { CALL_API } from '../middleware/call-api';
+import { fetchUserSnaps } from './snaps.js';
 
 export const REPO_TOGGLE_SELECT = 'REPO_TOGGLE_SELECT';
 
@@ -22,10 +23,17 @@ export const REPO_RESET = 'REPO_RESET';
 export function addRepos(repos, owner) {
   return async (dispatch) => {
     try {
-      await Promise.all(repos.map(async (repo) => {
+      await Promise.all(repos.map( async (repo) => {
         await dispatch(createWebhook(repo));
         return dispatch(addRepo(repo));
       }));
+
+      // XXX
+      // Wait for updated snaps list (so we don't show My repos until we
+      // updated the list with newly added repos).
+      //
+      // This is a temporary fix before we implement #666
+      await dispatch(fetchUserSnaps(owner));
 
       if (owner) {
         dispatch(push(`/user/${owner}`));
@@ -40,7 +48,7 @@ function addRepo(repository) {
   const { id, url } = repository;
 
   return {
-    id,
+    payload: { id },
     [CALL_API]: {
       path: '/api/launchpad/snaps',
       types: [REPO_ADD_REQUEST, REPO_ADD_SUCCESS],
@@ -78,10 +86,12 @@ export function resetRepository(id) {
   };
 }
 
-function createWebhook({ owner, name }) {
+function createWebhook({ owner, name, id }) {
   return {
+    payload: { id },
     [CALL_API]: {
       path: '/api/github/webhook',
+      types: [REPO_ADD_REQUEST, REPO_ADD_SUCCESS],
       options: {
         method: 'POST',
         headers: { 'Content-Type': 'application/json' },
--- a/src/common/components/repositories-home/index.js
+++ b/src/common/components/repositories-home/index.js
@@ -24,7 +24,7 @@ class RepositoriesHome extends Component {
     });
   }
 
-  componentDidMount() {
+  componentWillMount() {
     const { authenticated } = this.props.auth;
     const { updateSnaps } = this.props;
     const owner = this.props.user.login;
--- a/src/common/components/repositories-home/index.js
+++ b/src/common/components/repositories-home/index.js
@@ -77,7 +77,7 @@ class RepositoriesHome extends Component {
         <div className={ styles['button-container'] }>
           <HeadingThree>Repos to build and publish</HeadingThree>
           <div>
-            <LinkButton appearance="neutral" to="/select-repositories">
+            <LinkButton appearance="positive" to="/select-repositories">
               Add repos…
             </LinkButton>
           </div>
--- a/src/common/components/repositories-home/index.js
+++ b/src/common/components/repositories-home/index.js
@@ -98,7 +98,7 @@ class RepositoriesHome extends Component {
   render() {
     // show spinner if user has just signed in snaps data was not yet fetched
     //
-    // when snaps are loaded and user (how just signed in) doesn't have any,
+    // when snaps are loaded and user (who just signed in) doesn't have any,
     // they will be redirected to select repositories
     return (this.props.hasJustSignedIn && this.props.snaps.isFetching)
       ? this.renderSpinner()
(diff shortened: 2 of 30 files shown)


src/server/handlers/github-auth.js
```diff
@@ -114,7 +114,7 @@ export const verify = (req, res, next) => {
     });
 
     // Redirect to logged in URL
-    res.redirect(`/user/${req.session.user.login}`);
+    res.redirect(`/user/${req.session.user.login}?sign-in`);
```
bartaz: Workaround for redirecting users to /select-repositories if they don't have any repos.

We should only do the redirect after they sign in, so to distinguish getting to "My repos" from login or from other part of the app, we can use URL param.

This should not be needed once we implement proper checks if users ever had any repos (#780)

src/common/components/repositories-home/index.js
```diff
@@ -92,18 +96,18 @@ class RepositoriesHome extends Component {
...
-    // show spinner if snaps data was not yet fetched (snaps list is empty)
-    // (to avoid spinner during polling we are not checking `success` or `isFetching`
+    // show spinner if user has just signed in snaps data was not yet fetched
     //
-    // when snaps are loaded and user doesn't have any, they will be redirected
-    // to select repositories (so spinner won't be showing endlessly)
-    return !this.props.hasSnaps
+    // when snaps are loaded and user (who just signed in) doesn't have any,
+    // they will be redirected to select repositories
+    return (this.props.hasJustSignedIn && this.props.snaps.isFetching)
```
cjwatson: Hm, what about the case where a user refreshes their browser on `/user/:user` rather than going through the landing page?  `hasSnaps` will be false until snaps load, but I don't think `hasJustSignedIn` will be true, which means we'll show them an empty repository list rather than a spinner for a while.

src/common/components/repositories-home/index.js
```diff
@@ -53,10 +53,14 @@ class RepositoriesHome extends Component {
...
     if ((this.props.snaps.isFetching !== snaps.isFetching) && !snaps.isFetching) {
       // if user doesn't have enabled repos open add repositories view
-      if (!hasSnaps) {
+      if (!hasSnaps && nextProps.hasJustSignedIn) {
         this.props.router.replace('/select-repositories');
       } else {
-      // or update builds for their snaps
+        // or remove sign-in from URL
+        if (nextProps.hasJustSignedIn) {
+          this.props.router.replace(`/user/${nextProps.user.login}`);
```
cjwatson: Maybe it would be less of a time-bomb for the future to just remove `sign-in` from the query string rather than providing an entire replacement URL here?

src/common/containers/my-repos.js
```diff
@@ -15,10 +16,20 @@ class MyRepos extends Component {
...
 }
 
-export default MyRepos;
+MyRepos.propTypes = {
+  hasJustSignedIn: PropTypes.bool
+};
+
+function mapStateToProps(state, ownProps) {
+  return {
+    hasJustSignedIn: ownProps.location.query['sign-in'] !== undefined
```
cjwatson: `'sign-in' in ownProps.location.query`?

src/common/components/repositories-home/index.js
bartaz: OK. Would need to look into router API to see how to easily achieve that.

src/common/components/repositories-home/index.js
bartaz: The point of this change is for users to be able to see empty My repos.

If user sees empty My repos on /user/:user and refreshes they should see the same page.

The automatic redirect to select-repos is only meant to work for new users when they first sign in. Once they get back to (empty) My repos it should be OK for them to stay there after the refresh.

src/common/containers/my-repos.js
bartaz: That's where React Router passes current location to a component rendered by given route.

It's only visible on container level, because that's a component [referenced in the route](https://github.com/canonical-ols/build.snapcraft.io/blob/master/src/common/routes.js#L18).

src/common/components/repositories-home/index.js
bartaz: I agree that this is sub-optimal and temporary workaround.

On the other hand it seems #780 is quite high in priority, so maybe we don't need to push this solution and just find proper solution to detect new users.

src/common/components/repositories-list/index.js
```diff
@@ -116,9 +117,21 @@ export class RepositoriesListView extends Component {
...
 
-  render() {
+  renderRows() {
     const { ids } = this.props.snaps;
 
+    if (this.props.hasSnaps) {
+      return Array.from(ids).sort().map(this.renderRow.bind(this));
+    } else if (this.props.snaps.isFetching && !this.props.snaps.success) {
+      // only show spinner during initial loading (not during polling for updates)
+      return <div className={styles.noRepos}><Spinner /></div>;
```
bartaz: We need to show the spinner in My repos on initial page load, so that we don't show "No repos" until we actually load snaps from server and know if user has any.

src/common/components/repositories-home/index.js
```diff
@@ -28,7 +24,7 @@ class RepositoriesHome extends Component {
     });
   }
 
-  componentDidMount() {
+  componentWillMount() {
```
bartaz: We used to do initial data fetch in `componentDidMount`, but it resulted in rendering 'empty' state without the spinner because data was not loading yet.

Moving it to `componentWillMount` triggers actions earlier so that redux store updates before we do initial render. Even though data is not fetched yet, we at least know that fetching started and component renders spinner properly.

src/common/actions/repository.js
```diff
@@ -22,11 +23,18 @@ export const REPO_RESET = 'REPO_RESET';
...
         await dispatch(createWebhook(repo));
         return dispatch(addRepo(repo));
       }));
 
+      // XXX
+      // Wait for updated snaps list (so we don't show My repos until we
+      // updated the list with newly added repos).
+      //
+      // This is a temporary fix before we implement #666
+      await dispatch(fetchUserSnaps(owner));
```
bartaz: `addRepos` doesn't need to fetch user snaps, but thanks to that we will not redirect back to "My repos" until snaps list is updated, so when we do redirect back the list will already be updated with newly added snaps.

It adds additional request we need to wait for, so it would be best to progressively load (as described in #666), but for now it gives better user experience by not flashing old (or empty) My repos before data is loaded.

src/common/actions/repository.js
```diff
@@ -40,7 +48,7 @@ function addRepo(repository) {
   const { id, url } = repository;
 
   return {
-    id,
+    payload: { id },
```
bartaz: It was a bug. Id was not properly nested in the payload, so we didn't know which repo is being added and spinner was not properly shown on 'Add' button.

src/common/actions/repository.js
```diff
@@ -78,10 +86,12 @@ export function resetRepository(id) {
   };
 }
 
-function createWebhook({ owner, name }) {
+function createWebhook({ owner, name, id }) {
   return {
+    payload: { id },
     [CALL_API]: {
       path: '/api/github/webhook',
+      types: [REPO_ADD_REQUEST, REPO_ADD_SUCCESS],
```
bartaz: Adding repo actions to webhook actions allows us to show spinner also when we wait for webhook API.

src/server/handlers/github-auth.js
```diff
@@ -116,6 +108,15 @@ export const verify = (req, res, next) => {
       }
     });
 
+    // Save auth token and user info to session
+    req.session.githubAuthenticated = true;
+    req.session.token = body.access_token;
+    req.session.user = {
+      ...userResponse.body,
+      orgs,
+      hasAddedSnaps
```
bartaz: Exposing `hasAddedSnaps` in session user object (which is passed to initial redux store state).

PR #788, opened 2017-05-18
Show non-admin repos as disabled on the list

### LINKED ISSUE #787: Repository paging in select-repositories is broken
Paged views are sliced incorrectly...

bartaz:
Fixes #787 
screen shot 2017-05-18 at 13 18 53

### DIFF
--- a/src/common/components/select-repository-row/index.js
+++ b/src/common/components/select-repository-row/index.js
@@ -16,16 +16,18 @@ class SelectRepositoryRow extends Component {
     // TODO tidy up when we get rid of prefixes
     const isChecked = repository.isSelected || isEnabled;
     const isFetching = repository.isFetching;
-    const isDisabled = isEnabled || isFetching;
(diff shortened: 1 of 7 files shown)


src/common/components/select-repository-row/index.js
```diff
@@ -16,32 +16,32 @@ class SelectRepositoryRow extends Component {
     // TODO tidy up when we get rid of prefixes
     const isChecked = repository.isSelected || isEnabled;
     const isFetching = repository.isFetching;
-    const isDisabled = isEnabled || isFetching;
+    const isDisabled = isEnabled || isFetching || !repository.isAdmin;
```
cjwatson: This is pre-existing, I know, but it might be nice to sort out slightly different names for either or both of `isEnabled` (is repository already enabled as a snap build) and `isDisabled` (should this repository row be unavailable for selection) so that it's less terribly confusing.

PR #801, opened 2017-05-22
Fix error handling in callApi middleware

### LINKED ISSUE #774: non 200 responses from `/api/launchpad/builds` are ultimately handled incorrectly in the reducer
In dev, error in console looks like: `snap-builds.js:55 Uncaught (in promise) TypeError: Cannot read property 'error' of undefined`

https://github.com/canonical-ols/build.snapcraft.io/blob/master/src/common/reducers/snap-builds.js#L55 should probably be `action.payload.error`

bartaz:
Fixes #774 

What seemed to be just a small issue with getting error in reducer turned out to be a bigger issue with our `callApi` middleware.

This middleware currently both dispatches an action of given type and returns a promise (resolved if fetch is successful or rejected in case of an error).
The problem was, that in most cases when we handle API responses and expect reducer to handle the error we didn't expect and catch the rejected promise. Which resulted in uncaught promise rejections almost every time API returned an error (even when we 'expected' it and passed to reducer).

So, as the main part of this PR the call API middleware needed to change a bit so it either dispatches an action (if given a type) or returns a resolved/rejected promise (if action type is not given as setup).

### DIFF
--- a/src/common/middleware/call-api.js
+++ b/src/common/middleware/call-api.js
@@ -7,7 +7,7 @@ export const CALL_API = 'CALL_API';
 // Performs the call and promises when such actions are dispatched. This
 // has been implemented using promises because of suspected problems with
 // sourcemapping async / await code.
-export default defaults => () => next => action => {
+export default (defaults) => () => (next) => (action) => {
   // If action does not invoke CALL_API,
   // ignore it
   const settings = action[CALL_API];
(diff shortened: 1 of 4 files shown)


src/common/middleware/call-api.js
```diff
@@ -50,39 +50,26 @@ export default defaults => () => next => action => {
                 }));
               }
 
-              return Promise.resolve(result);
```
bartaz: After a bit of debugging I found that we had a bit of confusing and unnecessary code in this middleware.

The middleware already **always** returns a promise (because we `return fetch(...)`). So additional return `Promise.resolve()` or `Promise.reject()` was unnecessary.

src/common/middleware/call-api.js
```diff
@@ -50,39 +50,26 @@ export default defaults => () => next => action => {
...
+              // resolve promise with response JSON
+              return result;
             });
-        })
-        .catch((error) => {
```
bartaz: Also removed duplicated error handling.

We already have a `catch` on top level fetch promise, so it will catch the errors from nested promises.

src/common/middleware/call-api.js
```diff
@@ -50,39 +50,26 @@ export default defaults => () => next => action => {
...
         next(createAction({
           type: failureType,
           payload: {
             error: error
           },
           error: true
         }));
+      } else {
+        // otherwise, reject the promise with error
+        throw error;
```
bartaz: So the updated `callApi` middleware always returns a promise.

In case of an error, if the error action type was specified it will dispatch the error action and **resolve** the promise (because error was passed in the action). If the error action is not specified, the promise will be **rejected** with the error.

PR #807, opened 2017-05-25
Snaps details API

bartaz:
Part of #655 

To know if snap was released to stable channel we are proxying [CPI snaps details API](https://search.apps.ubuntu.com/docs/#details)

This PR simply exposes `/api/snaps/details/:name` route in BSI to be able to talk to CPI API cross origin.

### QA:

To test this PR run local BSI server and then try to go to API URL in the browser, for example: http://localhost:8000/api/snaps/details/core

You can also pass `fields`, `channel` query params that will be passed to CPI API:
http://localhost:8000/api/snaps/details/core?channel=beta&fields=name,revision

API should return a JSON response with details of given snap.

You can change last part of the URL to any public snap name. If snap doesn't exist API should return 404.


environments/production.env
```diff
@@ -8,6 +8,7 @@ OPENID_VERIFY_URL=https://build.snapcraft.io/login/verify
 LP_API_URL=https://api.launchpad.net
 STORE_API_URL=https://dashboard.snapcraft.io/dev/api
 STORE_DEVPORTAL_URL=https://dashboard.snapcraft.io/dev
+STORE_CPI_API_URL=https://search.apps.ubuntu.com/api/v1
```
cjwatson: Let's call this `STORE_SEARCH_API_URL` or something.  We sometimes refer to it as CPI but that's actually an abbreviation for the previous implementation (click-package-index), which has now been replaced by a new implementation, so it's even more opaque than your typical initialism ...

src/server/handlers/store.js
```diff
@@ -0,0 +1,42 @@
...
+export const getSnapDetails = async (req, res) => {
+  const snapName = req.params.name;
+  const query = {};
+
+  // pass allowed query parameters
+  for (const key of ['fields', 'channel', 'confinement'] ) {
+    if (req.query && req.query.hasOwnProperty(key)) {
+      query[key] = req.query[key];
+    }
+  }
```
cjwatson: A more concise approach might be to construct the query string like this:

    const query = qs.stringify(req.query, {
      filter: ['fields', 'channel', 'confinement']
    });

(See `src/server/handlers/login.js`, where we do this.)

Also, do you actually need `confinement`?  It doesn't seem like something that BSI would need, and the current search.apps.ubuntu.com codebase doesn't support it yet (there's a TODO comment for it).

src/server/routes/store.js
```diff
@@ -0,0 +1,13 @@
...
+import { json } from 'body-parser';
+
+import {
+  getSnapDetails
+} from '../handlers/store';
+
+const router = Router();
+
+router.use('/github/user', json());
+router.get('/snaps/details/:name', getSnapDetails);
```
cjwatson: The `router.use` parameter doesn't match `router.get` here.  (We should possibly switch to chained calls for this kind of thing so that it's harder to get wrong.)

PR #808, opened 2017-05-25
Match snapcraft and the dashboard's use of publish and release

### LINKED ISSUE #740: Match snapcraft and the dashboard's use of publish and release
It [was raised in the forum](https://forum.snapcraft.io/t/publish-or-release/518/11) that "publish" should be used when we speak about an ISV offering their applications and "release" for the action of making a version available (i.e. releasing into a channel).

If you disagree, by all means follow up on the forum thread. Otherwise, can we update the spec?

matthewpaulthomas:
Following Snapcraft terminology, changes “publish” to “release” when it is referring specifically to making a version available, as opposed to the overall process of building+uploading+releasing. Fixes #740.


test/routes/src/server/routes/t_badge.js
```diff
@@ -71,13 +71,13 @@ describe('The badge endpoint', () => {
         await supertest(app).get('/badge/anowner/aname.svg').expect(200);
       });
 
-      it('shoud return a SVG image with correct status', async () => {
+      it('should return a SVG image with correct status', async () => {
```
bartaz: Good catch, thanks :)

PR #812, opened 2017-05-29
Show install instructions for stable release of snap

### LINKED ISSUE #655: Next steps: update your install instructions [1]

bartaz:
Fixes #655

Currently only Twitter sharing is implemented. Others to be defined and implemented in #814.

screen shot 2017-05-30 at 09 36 59

screen shot 2017-05-30 at 09 36 31

### QA

1. Build a snap and wait for it to publish in the store
2. You should see 'No stable version yet' on 'My repos' page
3. Use the dashboard.snapcraft.io UI or snapcraft command line to release snap to stable channel
4. You should see 'To install the latest stable version' on 'My repos' page

### DIFF
--- a/src/common/actions/snaps.js
+++ b/src/common/actions/snaps.js
@@ -1,4 +1,5 @@
 import 'isomorphic-fetch';
+import qs from 'qs';
 
 import { CALL_API } from '../middleware/call-api';
 
@@ -44,7 +45,10 @@ export function removeSnap(repositoryUrl) {
 }
 
 export function fetchSnapStableRelease(repositoryUrl, snapName) {
-  const query = 'channel=stable&fields=name,revision';
+  const query = qs.stringify({
+    channel: 'stable',
+    fields: 'name,revision'
+  });
 
   return {
     payload: { id: repositoryUrl },
--- a/src/common/actions/snaps.js
+++ b/src/common/actions/snaps.js
@@ -8,6 +8,9 @@ export const FETCH_SNAPS_ERROR = 'FETCH_SNAPS_ERROR';
 export const REMOVE_SNAP = 'REMOVE_SNAP';
 export const REMOVE_SNAP_SUCCESS = 'REMOVE_SNAP_SUCCESS';
 export const REMOVE_SNAP_ERROR = 'REMOVE_SNAP_ERROR';
+export const FETCH_SNAP_DETAILS = 'FETCH_SNAP_DETAILS';
+export const FETCH_SNAP_DETAILS_SUCCESS = 'FETCH_SNAP_DETAILS_SUCCESS';
+export const FETCH_SNAP_DETAILS_ERROR = 'FETCH_SNAP_DETAILS_ERROR';
 
 export function fetchUserSnaps(owner) {
   const query = `owner=${encodeURIComponent(owner)}`;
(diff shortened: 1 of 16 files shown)


src/common/components/help/help-custom.js
```diff
@@ -0,0 +1,23 @@
+import React, { Component, PropTypes } from 'react';
+
+import { HeadingThree } from '../vanilla/heading/';
+import styles from './help.css';
+
+
+export default class HelpCustom extends Component {
```
bartaz: New component to allow custom content in help box (without install instructions).

src/common/components/help/help-install-snap.js
```diff
@@ -1,58 +1,58 @@
...
-          {/*
-          <Tweet
-            text={ tweet }
-          />
-          */}
+          <div>
+            <CopyToClipboard
+              copyme={ command }
+            />
+            { stable &&
```
bartaz: Tweet sharing is only available for stable releases.

src/common/components/help/help.css
```diff
@@ -11,6 +15,16 @@
   margin-bottom: 0;
 }
 
+.helpFlexWrapper {
```
bartaz: Some flex-box magic to make the headings of help boxes have equal height regardless of text length.

test/unit/src/common/reducers/entities/t_entities.js
```diff
@@ -3,21 +3,33 @@ import proxyquire from 'proxyquire';
...
+        default: repoSpy
+      },
+      './snap': {
+        default: snapSpy
+      }
+    }).entities;
+  });
+
+  afterEach(() => {
+    expect.restoreSpies();
```
bartaz: The spies were not properly cleared before, so some of the results have been false positives.

Should be fixed now.

src/server/handlers/store.js
```diff
@@ -27,7 +27,10 @@ export const getSnapDetails = async (req, res) => {
     });
     const json = await response.json();
 
-    res.status(response.status).send(json);
+    res.status(response.status).send({
+      status: json.error_list ? 'error' : 'success',
```
bartaz: We need to return `status: 'success'` from API for `callApi` middleware to correctly treat it as a success.

src/common/actions/snaps.js
```diff
@@ -39,3 +42,19 @@ export function removeSnap(repositoryUrl) {
     }
   };
 }
+
+export function fetchSnapStableRelease(repositoryUrl, snapName) {
+  const query = 'channel=stable&fields=name,revision';
```
cjwatson: I'd prefer this to be built from an object with `qs.stringify` or similar; not because the current code is wrong, but in case somebody modifies or copies-and-pastes this in future and introduces unsafe parameter encoding.

PR #816, opened 2017-05-30
Making sure Cancel button is always visible on Add repos page

### LINKED ISSUE #755: “Cancel” button missing from initial “Add repos” dialog
Currently, if you sign up for BSI and nobody has added an organization repo you can access, you see a variant of the “Add repos” dialog that has no “Cancel” button.

The reason for omitting “Cancel” was that there is no point in cancelling; there is literally nothing else you can do, and nothing else that will happen, in the app. (Also, our first-time flow logic would have immediately reopened the dialog as soon as you closed it.)

This is no longer the case: it’s possible that one of your colleagues adds an organization repo, so the dialog is not the only way you can end up with useful things to do.

So, once the empty state is implemented (#657), we should make the “Cancel” button always show up in the Add dialog.

bartaz:
Fixes #755 

Even if there is no repos user should be able to cancel 'Add repos' and go back to 'My repos'.

### QA

1. Clean-up or remove app db (or make sure to sign in with fresh user).
2. Sign in and get redirected to 'Add repos' page
3. Cancel button should be visible, click it to get to empty 'My repos' page

### DIFF
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -116,7 +116,7 @@ export class SelectRepositoryListComponent extends Component {
           <HeadingThree>
             { selectedRepositories.length } selected
           </HeadingThree>
-          <div className={ styles.buttons }>
+          <div>
             <LinkButton appearance="neutral" to={`/user/${user.login}`}>
               Cancel
             </LinkButton>
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -116,24 +116,19 @@ export class SelectRepositoryListComponent extends Component {
           <HeadingThree>
             { selectedRepositories.length } selected
           </HeadingThree>
-          <div className={ styles['footer-right'] }>
-            <div className={ styles['button-wrapper'] }>
-              { ids && ids.length > 0 &&
-                <LinkButton appearance="neutral" to={`/user/${user.login}`}>
-                  Cancel
-                </LinkButton>
-              }
-            </div>
-            <div className={ styles['button-wrapper'] }>
-              <Button
-                appearance={ 'positive' }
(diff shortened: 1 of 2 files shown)


src/common/components/select-repository-list/index.js
```diff
@@ -116,24 +116,19 @@ export class SelectRepositoryListComponent extends Component {
...
-                isSpinner={buttonSpinner}
-              >
-                Add
-              </Button>
-            </div>
+          <div className={ styles.buttons }>
+            <LinkButton appearance="neutral" to={`/user/${user.login}`}>
+              Cancel
+            </LinkButton>
+            {' '}
```
bartaz: Putting space between inline buttons (to avoid additional wrappers for spacing).

src/common/components/select-repository-list/index.js
```diff
@@ -143,7 +138,7 @@ export class SelectRepositoryListComponent extends Component {
   renderPageLinks(pageLinks) {
     if (pageLinks) {
       return (
-        <div className={ styles['page-links-container'] }>
+        <div className={ styles.pageLinksContainer }>
```
bartaz: Just a code style fix.

src/common/components/select-repository-list/index.js
```diff
@@ -116,24 +116,19 @@ export class SelectRepositoryListComponent extends Component {
...
-              <Button
-                appearance={ 'positive' }
-                disabled={ !selectedRepositories.length || buttonSpinner }
-                onClick={ this.handleAddRepositories.bind(this) }
-                isSpinner={buttonSpinner}
-              >
-                Add
-              </Button>
-            </div>
+          <div className={ styles.buttons }>
```
cjwatson: As mentioned on IRC, this seem like a leftover from something since `styles.buttons` isn't defined.

PR #821, opened 2017-06-01
Show spinner on buttons without affecting their width

### LINKED ISSUE #622: Give user permission to collaborate on organisation snaps
So that the user doesn't need to be directly signed in to the organisation's SSO account, either alongside or instead of their existing SSO account, we should add collaboration information to the store so the user has direct access to build and publish the snap.

This cannot start until changes have been made to the snap store to support collaboration.

More details can be found in [this discussion](https://forum.snapcraft.io/t/github-organisations-proposal/157).

### LINKED ISSUE #662: "Add repos" list: Prevent "cancel" button from jumping arround
After you click the "Add" button, a spinner appears within the button:

add spinner

This causes the cancel button next to it to "jump" left. This should be solved somehow.

bartaz:
Fixes #662 

Making sure spinner during loading shows on buttons instead of the text (rather then next to it), not changing its width.

### QA

1. Go to add repos, select some repos, click Add button
2. Spinner should show on Add button in place of the text (not changing buttons width)
3. On my repos open 'Register name' dropdown, Sign in
4. Sign in button should show spinner instead of the text (not changing width)
5. Click 'Register name' button
6. Register button should show spinner without changing buttons width


src/common/components/spinner/index.js
```diff
@@ -1,6 +1,8 @@
 import React, { Component, PropTypes } from 'react';
 
-import spinner from './spinner.svg';
+import spinnerDark from './spinner-dark.svg';
+import spinnerLight from './spinner-light.svg';
```
bartaz: Added new white spinner to use it on dark button backgrounds.

src/common/components/vanilla/button/button.css
```diff
@@ -107,15 +107,23 @@ $neutral-border-color: $dark-grey;
   box-sizing: content-box;
 }
 
+.hasSpinner {
+  position: relative;
+}
+
+.hasSpinner .text {
+  /* hide text without chaning button width */
+  opacity: 0;
```
bartaz: Hiding text with opacity (so the width of the button doesn't change).

src/common/components/vanilla/button/button.css
```diff
@@ -107,15 +107,23 @@ $neutral-border-color: $dark-grey;
...
-  margin-right: 0.5em;
-  height: 21px;
-  width: 21px;
-  float: left;
-  box-sizing: content-box;
+  position: absolute;
+  height: 20px;
+  width: 20px;
+  left: 50%;
+  top: 50%;
```
bartaz: Positioning spinner in the middle of the button.

src/common/components/vanilla/button/index.js
```diff
@@ -1,53 +1,58 @@
...
-export default function Button(props) {
-  const { appearance='primary', flavour='normal', isSpinner=false, ...rest } = props;
-  return (
-    <button {...rest} className={ `${style[appearance]} ${style[flavour]}` }>
-      { isSpinner && <span className={ style.spinner }><Spinner /></span> }
-      { props.children }
-    </button>
-  );
+function createButtonComponent(Component) {
+  function ButtonComponent(props) {
```
bartaz: We have 3 button classes (`Button`, `Anchor` and `LinkButton` - for `<button>`, `<a>`, and React router `Link`). Previously they had separate implementations that were meant to share functionality, props, etc, but in the end each of them was different (supporting different subset of props).

This refactoring fixes that. `createButtonComponent` is a factory function that builds a button component for given component (or tag name).

src/common/components/vanilla/button/index.js
```diff
@@ -1,53 +1,58 @@
...
-  const { appearance='primary', flavour='normal', icon, ...rest } = props;
-  return (
-    <a {...rest} className={ `${style[appearance]} ${style[flavour]}` }>
-      { props.children }
-      { icon && <img className= { style.icon } src={ icon } /> }
-    </a>
-  );
-}
+const Button = createButtonComponent('button');
+Button.displayName = 'Button';
```
bartaz: Using `displayName` to make sure generated components have user friendly (and backwards compatible) names for developer tools and testing.

src/common/components/vanilla/button/button.css
bartaz: Thanks for suggestion, I tend to forget we already have all major browsers supporting stuff like this.

Issue #715, opened 2017-05-04
Visual update of 'Missing repos' info in Add repos


bartaz:
Follow-up to #680 (implemented in #713).

Popover/dialog about missing repos needs visual design and proper implementation in place of quick reuse of popover 'tooltip'.

Spec defines UI for Add repos as scrollable dialog: https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.43owh6tjv12q

But we don't have visual design for that and it needs to be implemented before 'Missing repos' info can be moved into dialog popover as defined in spec.

kwm14:
## Updated iteration

- adding new popover/dialog view

### 001 Missing repos

![updates-02-add-repos-some-selected-dialog](https://cloud.githubusercontent.com/assets/17748020/25799744/29fd0484-33de-11e7-80f8-6205bfb973e0.png)

magdalena-mirowicz:
@kwm14 @matthewpaulthomas please chat that through

matthewpaulthomas:
@kwm14 As we just discussed, that is rather different from the wireframe, where the list of suggestions _replaces_ the list of repos inside the dialog. I proposed that for four reasons:
- So that the suggestions have exactly the same layout for “Any missing?” as they do in the empty state.
- So that on small screens, we don’t need to use horizontal space for a border/shadow/margin conveying that there’s a popover on top of the dialog (since there is no popover).
- So that on small screens, it’s obvious that the list of suggestions scrolls (since it’s replacing a list that scrolls).
- Least of all, because all else being equal, two layers is simpler than three.

A benefit of your design, though, is that if someone clicked “Any missing?” by accident, it’s clearer how to get from the suggestions back to the list of repos. Clicking an X is more obvious than clicking anywhere outside the suggestions.

Ideally we can find a design that combines these advantages. For example, maybe “Any missing?” when you’re viewing the list could change to “Return to list” when you’re viewing the suggestions.

matthewpaulthomas:
Work in progress:
![add repos at different screen sizes](https://cloud.githubusercontent.com/assets/19801137/26113411/5fb9fbca-3a52-11e7-8f87-45da126e05f1.png)

kwm14:
## Updated visual

- referencing @matthewpaulthomas wireframes above
- some visual styling different to wires to fit inline with Vanilla/current patterns

### Desktop 
[...]

PR #838, opened 2017-06-12
Moved private repo info out of popover

bartaz:
Part of preparations for #715 

This is a refactoring change that takes private repo info contents and moves it out of current popover to separate component, so it can be reused in different context (scrollable area of Add repos) in #715

### DIFF
--- a/src/common/components/private-repos-info/index.js
+++ b/src/common/components/private-repos-info/index.js
@@ -1,15 +1,11 @@
 import React, { Component, PropTypes } from 'react';
 
-import { conf } from '../../helpers/config';
-const BASE_URL = conf.get('BASE_URL');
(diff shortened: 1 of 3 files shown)


nottrobin:
@bartaz would you mind going into a little more detail here about exactly what change I should be seeing and why? I'm sure that information is buried in #715 if you understand it thoroughly enough, but I'm afraid I'm struggling to work it out.

bartaz:
@nottrobin With this PR you shouldn't see any change in the app - it's just moving code out from popover to separate component, so it can be used in next PR.

nottrobin:
@richmccartney you feel? You have faith? :stuck_out_tongue: 

This looks good to me. Let's get it merged. :+1:

Issue #400, opened 2017-03-06
[Bug] “Add repos” > “Cancel” > needless table reloading


nottrobin:
1\. Go to the Dashboard.
2\. Choose “Add repos”.
3\. Choose “Cancel”.

What happens:

1\. There is a spinner while the table loads.
3\. There is a spinner while the table reloads.

What should happen:

3\. Since you haven’t changed the contents of the table, it shouldn’t need to reload.

Comments:

---
Creator: Robin Winslow
Assignees: 
Copied from Trello card: https://trello.com/c/Ja8ZN2Z3/306-bug-add-repos-cancel-needless-table-reloading
Labels: copy to github

solazio:
No relevant for snapcraft.io, closing

Issue #853, opened 2017-06-23
“Anything missing?”/“Return to repos list” jumps to the top of the page


matthewpaulthomas:
1\. Choose “Add repos…”.
2\. Scroll away from the top of the page (for example, to see all of the dialog on a small screen).
3\. Choose “Anything missing?”.
4\. Scroll away from the top of the page.
5\. Choose “Return to repos list”.

What happens:
3\. The page scrolls to the top.
5\. The page scrolls to the top.

What should happen:
3\. The page stays at the same scroll position.
5\. The page stays at the same scroll position.

richmccartney:
This should be resolved by #856 PR :)

Issue #829, opened 2017-06-05
'Add repos' dialog should adapt height to viewport


bartaz:
Follow up to #827 

We made 'Add repos' dialog scrollable in #827, but using a fixed height for it.

As per [spec](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.43owh6tjv12q), the dialog height should adapt to viewport height, show or hide first time heading, etc.
This functionality needs to be implemented here.

bartaz:
@mpt With header, beta notification, user avatar and first time heading there is quite a lot of space taken already, leaving not much space for 'Add repos' dialog.

As far as I can see user avatar is removed from newer visual designs, but even without it it may be hard to fit everything on smaller screen.

How should we define the rules how to show 'Add repos' and other parts of the screen?
If screen is too small for first time heading should it be simply hidden (not displayed at all) or should user be able to scroll up to it?
Should we make space to make footer visible or should dialog just expands and keep footer off screen?

Screenshot below taken on screen with heigh of 1200px (which leaves over 1000px for whole page) and is likely more than usual laptop screen would have. Still, footer didn't fit the screen and significant amount of screen is taken by elements other than 'Add repos' dialog:

<img width="1099" alt="screen shot 2017-06-05 at 15 08 31" src="https://cloud.githubusercontent.com/assets/83575/26820330/65d389e2-4a9a-11e7-8d6f-0d2f36443ff5.png">

bartaz:
@matthewpaulthomas Thanks for the update.

Vanilla has some breakpoints built in:
```
$breakpoint-x-small: 460px !default;
$breakpoint-small: 620px !default;
$breakpoint-medium: 768px !default;
$breakpoint-large: 1030px !default;
```
Where widths below `small` trigger mobile layout.

I guess we could start with these values as well to stay consistent.

From spec point of view it seems that everything needed for implementation is ready.

solazio:
Not relevant for snapcraft.io, closing

PR #856, opened 2017-06-26
Removed the onDocument click events for closing dialog

### LINKED ISSUE #854: Clicking outside Add dialog switches back to repo listing
1\. Choose “Add repos…”.
2\. Choose “Anything missing?”.
3\. Click anywhere outside the dialog.

What happens: The dialog switches back to the repo listing.

What should happen: Nothing.

richmccartney:
Removed the onDocument click events for closing the missing / private repo dialog. Updated toggle link with aria details to help accessibility.

## Done

- Removed JS from ```select-repository-list/index.js``` file to stop clicking outside of dialog
- Updated markup on the renders on help link, including:
- - Aria labels to update what action is being taken on click
- - Removed anchor tags on the help link, this should be a button as anchor tags need to link to a new page or ID on the page and effect the URL.
- - Created link button style to look similar to an anchor link

## QA

- Go to the add repos page, click the help link and try to click outside the add repos dialog. The popup should now stay active. Also clicking on the help link shouldn't jump a user to the top of the page now.

## Issue / Card

- Fixes #854 #853

### DIFF
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -184,7 +184,7 @@ export class SelectRepositoryListComponent extends Component {
               { selectedRepositories.length } selected
             </strong>
             {' '}
-            (<Button appearance={ 'link' } onClick={this.onHelpClick.bind(this)} aria-label={ this.state.showMissingReposInfo ? 'Click to close the missing / private repos dialog' : 'Click to open the missing / private repos dialog' }>
+            (<Button appearance={ 'link' } onClick={this.onHelpClick.bind(this)}>
               { this.state.showMissingReposInfo ? 'Return to repos list' : 'Anything missing?' }
             </Button>)
           </div>
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -198,7 +198,6 @@ export class SelectRepositoryListComponent extends Component {
               disabled={ !selectedRepositories.length || buttonSpinner }
               onClick={ this.handleAddRepositories.bind(this) }
               isSpinner={buttonSpinner}
-              id={ 'add-repo'}
             >
               Add
             </Button>
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -198,6 +198,7 @@ export class SelectRepositoryListComponent extends Component {
               disabled={ !selectedRepositories.length || buttonSpinner }
               onClick={ this.handleAddRepositories.bind(this) }
               isSpinner={buttonSpinner}
+              id={ 'add-repo'}
             >
               Add
             </Button>
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -42,12 +42,6 @@ export class SelectRepositoryListComponent extends Component {
     this.props.selectedRepositories && this.props.selectedRepositories.map(id => {
       this.props.dispatch(resetRepository(id));
     });
-
-    // bind document click event
-    if (typeof document !== 'undefined') {
-      this.onBoundDocumentClick = this.onDocumentClick.bind(this);
-      document.addEventListener('click', this.onBoundDocumentClick);
-    }
   }
 
   componentWillUnmount() {
@@ -55,12 +49,6 @@ export class SelectRepositoryListComponent extends Component {
     if (typeof document !== 'undefined') {
       document.removeEventListener('click', this.onBoundDocumentClick);
     }
-  }
-
-  onDocumentClick() {
-    this.setState({
-      showMissingReposInfo: false
-    });
   }
 
   onHelpClick(event) {
@@ -196,7 +184,7 @@ export class SelectRepositoryListComponent extends Component {
               { selectedRepositories.length } selected
             </strong>
             {' '}
-            (<a href="#" onClick={this.onHelpClick.bind(this)}>
+            (<a onClick={this.onHelpClick.bind(this)} aria-label={ this.state.showMissingReposInfo ? 'Click to close the missing / private repos dialog' : 'Click to open the missing / private repos dialog' }>
               { this.state.showMissingReposInfo ? 'Return to repos list' : 'Anything missing?' }
             </a>)
           </div>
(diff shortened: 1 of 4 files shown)


src/common/components/vanilla/button/button.css
```diff
@@ -99,6 +101,23 @@ $neutral-border-color: $dark-grey;
   background-color: $neutral-button-color-hover;
 }
 
+.link {
+  -webkit-appearance: none;
+  -moz-appearance:    none;
+  appearance:         none;
+  background-color: transparent;
+  border: 0;
+  color: $internal-button-text-color;
```
bartaz: We don't have variable called `$internal-button-text-color`, so this fails to compile.

Did you mean `$link-button-text-color` that you added in the top?

src/common/components/select-repository-list/index.js
```diff
@@ -196,9 +184,9 @@ export class SelectRepositoryListComponent extends Component {
               { selectedRepositories.length } selected
             </strong>
             {' '}
-            (<a href="#" onClick={this.onHelpClick.bind(this)}>
+            (<Button appearance={ 'link' } onClick={this.onHelpClick.bind(this)} aria-label={ this.state.showMissingReposInfo ? 'Click to close the missing / private repos dialog' : 'Click to open the missing / private repos dialog' }>
```
bartaz: If we want at add accessibility `aria-` labels different from button text we should confirm the label text with @matthewpaulthomas.

test/unit/src/common/components/select-repository-list/t_select-repository-list.js
```diff
@@ -92,7 +92,7 @@ describe('<SelectRepositoryListComponent /> instance', function() {
     });
 
     it('should render disabled Add button', function() {
-      expect(wrapper.find('Button').prop('disabled')).toBe(true);
+      expect(wrapper.find('#add-repo').prop('disabled')).toBe(true);
```
bartaz: I would prefer to avoid adding `id`s to elements just for test purposes.

We could use enzyme API to find a button we need, for example `wrapper.find('Button').last().prop('disabled')` should work in this case (as add button is last one).
Of course it's not perfect (breaks if buttons will be moved around in the component) but lets us avoid unnecessary element ids.

src/common/components/select-repository-list/index.js
richmccartney: @bartaz so with ```aria-labels``` you are to use them to explain the function in a more verbose way. This means it should describe the action it's trying to do or the page your trying to lead the user too. This is something we've done in MAAS and I feel this content would suffice.

bartaz: @richmccartney Sure, but I'd rather confirm if we want to call it "missing / private repos dialog", as such name with "/" may be quite confusing out of context.

bartaz: @richmccartney Is `aria-label` actually needed here at all?

From what I understand it's meant to be used when there is *no* visible element label (for example it has an icon only), but we have a clear text on this button (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute).

richmccartney: @bartaz as much as that is true to describe hidden content it should also be used to explain actions when they're not clear. For example a download button is great and could say "Download ubuntu". But for accessibility we would add an aria-label to explain it would be downloading ubuntu, the size of the document and the type of file. This is something we've done in MAAS and I will be trying to do this in BSI.

i'll definitely switch the "/" to or though :)

bartaz: @richmccartney OK. As 'private' repos are missing, I guess "Click to show info about missing repos" should be enough.

matthewpaulthomas: @richmccartney To expand on what I said yesterday, I think this really is a link, an in-page link that navigates to/from part of the page containing the explanation of why the list might be incomplete. It just so happens that in the current design, the explanation temporarily covers the list, instead of being (for example) at the bottom of the page.

But whether it is a link or a button, @bartaz appears to be correct: `aria-label` is not appropriate here. Besides [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute), [the W3C](https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html), [Microsoft](https://msdn.microsoft.com/en-us/library/hh968243(v=vs.85).aspx), and [Opera](https://dev.opera.com/articles/ux-accessibility-aria-label/) are all clear that `aria-label` is for use on elements that don’t already have understandable text labels (for example, a close button where the label is only “X”). Furthermore, [Google’s reference](https://developers.google.com/web/fundamentals/accessibility/semantics-aria/aria-labels-and-relationships) illustrates that using it to give a button supplementary information such as “the size of the document and the type of file” would be incorrect, because “if a button has both text content and an aria-label, only the aria-label value will be used”.

I think that in context, “Any missing?” and “Return to repos list” are understandable text labels. If they aren’t, let’s fix the actual labels, rather than adding substitute labels that take effect only if you’re using an assistive device.

richmccartney: I would say "Anything missing?" doesn't really gives decent context for the page or area of content it's in. What is missing exactly? If anything is missing what am I expecting to see when I click this? 

@matthewpaulthomas can you supply possibly an improvement to this, thus I would remove the ```aria-label``` because currently as it stands a user whom is vision impaired or requires any other form of extra context wouldn't know what is missing. 

Regarding to the link comments, unless the ```<a>``` element links to an ID or new page with in the site / app, a button element is far more applicable and semantic for the page.

matthewpaulthomas: Ah, now I understand the confusion. [The text I specced](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.1uxwfvh1rvjz) was not “Anything missing”, it was “{total} repos (**Any** missing?)”. In this context, “Any” obviously means “any repos”. But I guess it isn’t obvious in the code, where this part of the string is implemented in a separate file (select-repository-list/index.js) from the rest of the string (selectors/index.js). I’ve submitted #862 to fix the text.

(Ideally, yes, I do think the link would link to an ID, so you could bookmark the explanation, open it in a new tab, etc. But that probably shouldn’t be fixed until after the dialog as a whole becomes a fragment of the “My repos” URL, as specced and as possibly required to fix #400 and/or #829.)

PR #860, opened 2017-06-29
832 add repos amount number

### LINKED ISSUE #832: Show number of user repos in 'Add repos' footer
In 'Add repos' footer we currently only show the number of selected repos, as per design we should show `X selected out of Y repos`.

richmccartney:
## Done

- Matches design and placement of the selected and total repos message on the add repos listing page
- Creates new function to pull in the amount of repos and what is selected with unit tests

## QA

- Go to the add repos page once logged in. When loading a message should show calculating your repos. Once complete the selected and total message should be present. 

## Issue / Card

Fixes #832

### DIFF
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -129,9 +129,6 @@ export class SelectRepositoryListComponent extends Component {
     const { ids, error, isFetching, isDelayed } = this.props.repositories;
     const { selectedRepositories } = this.props;
 
-    let repoAmount = ids;
-    let selectedRepos = selectedRepositories;
-
     if (isFetching && ids.length === 0) {
       return (
         <span>Calculating your total repositories&hellip;</span>
@@ -141,13 +138,12 @@ export class SelectRepositoryListComponent extends Component {
     return (
       <div>
           <strong>
-            { selectedRepos.length } selected out of
+            { selectedRepositories.length } selected out of
           </strong>
           {' '}
           <span>
-            { repoAmount.length } Total repositories
+            { ids.length } Total repositories
           </span>
-          {' '}
       </div>
     );
   }
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -126,7 +126,7 @@ export class SelectRepositoryListComponent extends Component {
   }
 
   renderRepoAmount() {
-    const { ids, error, isFetching, isDelayed } = this.props.repositories;
+    const { ids, isFetching } = this.props.repositories;
     const { selectedRepositories } = this.props;
 
     if (isFetching && ids.length === 0) {
@@ -137,9 +137,9 @@ export class SelectRepositoryListComponent extends Component {
 
     return (
       <div>
-          <strong>
-            { selectedRepositories.length } selected out of { ids.length } Total repositories
-          </strong>
+        <strong>
+          { selectedRepositories.length } selected out of { ids.length } Total repositories
+        </strong>
       </div>
     );
   }
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -138,12 +138,8 @@ export class SelectRepositoryListComponent extends Component {
     return (
       <div>
           <strong>
-            { selectedRepositories.length } selected out of
+            { selectedRepositories.length } selected out of { ids.length } Total repositories
           </strong>
-          {' '}
-          <span>
-            { ids.length } Total repositories
-          </span>
       </div>
     );
   }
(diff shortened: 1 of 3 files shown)


src/common/components/select-repository-list/index.js
```diff
@@ -125,6 +125,25 @@ export class SelectRepositoryListComponent extends Component {
...
+    const { selectedRepositories } = this.props;
+
+    if (isFetching && ids.length === 0) {
+      return (
+        <span>Calculating your total repositories&hellip;</span>
+      );
+    }
+
+    return (
+      <div>
```
bartaz: Is that additional `div` needed here? I think rendering it as inline element should be fine.

src/common/components/select-repository-list/index.js
```diff
@@ -125,6 +125,25 @@ export class SelectRepositoryListComponent extends Component {
...
+        <strong>
+          { selectedRepositories.length } selected out of { ids.length } repositories
```
bartaz: [Spec says](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#) it should be "{number} selected of {total} repos"

src/common/components/select-repository-list/index.js
```diff
@@ -125,6 +125,25 @@ export class SelectRepositoryListComponent extends Component {
     return renderedRepos;
   }
 
+  renderRepoAmount() {
+    const { ids, isFetching } = this.props.repositories;
+    const { selectedRepositories } = this.props;
+
+    if (isFetching && ids.length === 0) {
+      return (
+        <span>Calculating your total repositories&hellip;</span>
```
bartaz: I don't think we should be showing it at all.

First of all, it's not in [spec](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#), secondly showing it is contrary to delaying showing the loading spinner. We delay showing the spinner for 2 seconds (not to show immediately that we are loading), but this text would basically be a "loading..." text that appears for a very short time.

The spec doesn't specify it in detail, but I suppose that before we load anything (so we don't know how many repos there are or how many pages there are) we shouldn't show any text at all.

src/common/components/select-repository-list/index.js
```diff
@@ -149,13 +168,13 @@ export class SelectRepositoryListComponent extends Component {
...
           <div className={ styles.summary }>
-            <strong>
-              { selectedRepositories.length } selected
-            </strong>
-            {' '}
-            (<Button appearance={ 'link' } onClick={this.onHelpClick.bind(this)}>
-              { this.state.showMissingReposInfo ? 'Return to repos list' : 'Anything missing?' }
-            </Button>)
+            { this.renderRepoAmount() }
+            <div>
```
bartaz: I'm not sure if this additional wrapper div is needed, I think this text and link can stay inline.

richmccartney: The addition wrapper is needed to keep it all inline.

bartaz: But now with `div` removed from `renderRepoAmount` they should be inline (as they are just inline elements), right?

src/common/components/select-repository-list/styles.css
```diff
@@ -7,9 +7,11 @@
 }
 
 .summary {
+  display: flex;
   flex: 1;
   margin-right: 10px;
   text-align: right;
+  justify-content: flex-end;
```
bartaz: I think this is not needed.

Text was right aligned just fine before. `justify-content: flex-end;` is needed because `display: flex;` was added here as well. If both `flex` and `justify-content` are removed `text-align` does the trick.

src/common/components/select-repository-list/index.js
```diff
@@ -125,6 +125,34 @@ export class SelectRepositoryListComponent extends Component {
...
+    if (selectedRepositories.length > 0) {
+      reposSelected = true;
+    } else {
+      reposSelected = false;
+    }
+
+    // Return selected repos amount and the total
+    return (
+      <strong>
+        { reposSelected &&
```
bartaz: I would say that that `reposSelected` is not necessary as variable here.

This should be sufficient:
```
{ (selectedRepositories.length > 0) &&
    <span>{ selectedRepositories.length } selected of </span>
 }
{ ids.length } repos
```

And we will avoid having additional variable and whole `if` block that sets its value on simple check.

bartaz: And please put `{ids.length} repos` in a separate line to keep the conditional block above it more separate.

PR #870, opened 2017-07-04
Updated breadcrumbs component to use vanilla generated styles

### LINKED ISSUE #869: Create new breadcrumbs React component that uses Vanilla CSS module

bartaz:
## Done

Created new component to replace 'vanilla-inspired' breadcrumbs with vanilla-generated breadcrumbs component.

## QA

- Sign in
- Go to builds and builds details pages

Breadcrumbs should look (almost) the same as before and use updated Vanilla styles (directly taken from Vanilla framework).

## Issue / Card

Fixes #869

## Screenshots

screen shot 2017-07-04 at 16 56 56


src/common/components/vanilla-modules/breadcrumbs/index.js
```diff
@@ -0,0 +1,41 @@
+import React, { PropTypes } from 'react';
+import { Link } from 'react-router';
+
+import style from '../../../style/vanilla/css/breadcrumbs.css';
+
+const breadcrumbsStyle = (element = '', modifier = '') => {
```
bartaz: A helper function to generate vanilla BEM class names and bind them to CSS modules `styles` object.

src/common/containers/build-details.js
```diff
@@ -72,9 +71,12 @@ class BuildDetails extends Component {
         <BetaNotification />
         <Breadcrumbs>
           { user &&
-            <Link to={`/user/${user.login}`}>My repos</Link>
+            <BreadcrumbsLink to={`/user/${user.login}`}>My repos</BreadcrumbsLink>
           }
-          <Link to={`/user/${repository.fullName}`}>{repository.fullName}</Link>
+          {' ' /* vanilla assumes space between breadcrumb elements */}
```
bartaz: Styles for Vanilla breadcrumbs pattern assume there will be a space between breadcrumb items, but JSX doesn't render space between elements, so we need to manually add that.

But was reported to Vanilla about it: https://github.com/vanilla-framework/vanilla-framework/issues/1146

src/common/containers/build-details.js
```diff
@@ -72,9 +71,12 @@ class BuildDetails extends Component {
...
+          <BreadcrumbsLink to={`/user/${repository.fullName}`}>{repository.fullName}</BreadcrumbsLink>
+          {' ' /* vanilla assumes space between breadcrumb elements */}
+          <BreadcrumbsLink>&nbsp;</BreadcrumbsLink> {/* hack to show arrow after last breadcrumb item */}
```
bartaz: Because we want to show an arrow after last item in the breadcrumb, we need to hardcode empty breadcrumb item.

richmccartney: I can see that the intent is so the breadcrumb links to the title of the page. I feel we should still include this in the breadcrumb but not as something that can be interacted with? 

This is a common pattern across our sites and I feel we should do the same in Snapcraft. @carlaberkers what are you thoughts? I've attached a screenshot of what it looks like with the new VF updates.

image

bartaz:
@richmccartney The point of this PR is just to replace the custom implementation with Vanilla implementation without doing any design changes.

If we feel like we should use breadcrumbs differently, let's raise separate design issue for that.

carlaberkers:
Hey @richmccartney and @bartaz - we should make the patterns in line with Vanilla as it is used in other places. so I would agree we should add the page title to the breadcrumb.

In this case the build number is in the breadcrumb and in the table, so I think the page header is redundant in this case.

Hope that helps!

bartaz:
@carlaberkers Yes, thanks for update. As this was only meant to be a implementation change (without raising to much design discussions), let's keep it as it is for now and have a separate card to update the design.

bartaz:
@carlaberkers @richmccartney I created an issue about breadcrumbs design: #872

PR #871, opened 2017-07-05
Added font smoothing to base styles same as Vanilla framework

bartaz:
## Done

Added font smoothing (same as Vanilla) to base CSS.

## QA

- All fonts should be smooooooth ;)

### DIFF
--- a/src/common/style/base.css
+++ b/src/common/style/base.css
@@ -2,6 +2,12 @@
   box-sizing: inherit;
 }
 
+* {
+  -moz-osx-font-smoothing: grayscale;
+  -webkit-font-smoothing: antialiased;
+  font-smoothing: subpixel-antialiased;
+}
+
 html, body {
   overflow-y: auto;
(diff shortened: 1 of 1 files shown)


matthewpaulthomas:
Browsers already smooth fonts by default. This change asks them to apply lower-resolution smoothing (grey-scale rather than subpixel). I’ve reported [vanilla-design#31](https://github.com/ubuntudesign/vanilla-design/issues/31) on removing this from base Vanilla.

richmccartney:
@matthewpaulthomas yeah thats an incorrect statement, by and large all browsers render fonts differently per browser. For Chrome setting the ```-webkit-font-smoothing: antialiased;``` will removed any sub pixel antialiasing. Likewise for grey scaling on Firefox. 

This is a design requirement,  and also only really effects OSX versions of the fonts. They will be staying in the site and framework to keep the fonts more inline with Windows, other OS's and the designs from the team.

PR #873, opened 2017-07-06
866 create search field component

richmccartney:
## Done

- Created the new search field component for the repositories list
- Included styles to align content in the add repos header

## QA

- Go to the add repos page and the new search component shall show in the header of the over all module.

## Issue / Card

#866


src/common/components/select-repository-search/index.js
```diff
@@ -0,0 +1,29 @@
+import React, { Component, PropTypes } from 'react';
+import styles from './search.css';
+
+export default class SelectRepositorySearch extends Component {
```
bartaz: Something to consider - as this component has only rather simple render function it can be just a function, not a class:

```js
export default function SearchInput(props) {
  let id // ... etc
  return (
     <label> ... </label>
  )
}

Comment (bartaz) [inline src/common/components/select-repository-search/index.js]: Also I would consider this as a more reusable 'search' component, as we could use it to search anything else, so I'd rename it to 'SearchInput' or something similar (not related to repos).

File: src/common/components/select-repository-search/index.js [hunk 2]
```diff
@@ -0,0 +1,29 @@
...
+
+  render() {
+    let id = this.props.id || 'search-repositories';
+    let label = this.props.label || 'Search';
```
bartaz: On visual designs (there are some in comments #715) repos search input doesn't have a label or placeholder, so I would make the defaults empty.

Also search input on design has an icon in it. Maybe we can find such in assets manager, or something similar in juju/maas?

src/common/components/select-repository-search/index.js
```diff
@@ -0,0 +1,29 @@
+import React, { Component, PropTypes } from 'react';
+import styles from './search.css';
+
+export default class SelectRepositorySearch extends Component {
+
+  render() {
+    let id = this.props.id || 'search-repositories';
```
bartaz: I wouldn't set id default value as it will be easily forgotten and we can end up with many inputs with same id.

I suppose if input is inside of the label id is really optional, so maybe we could have it without any id unless dev specifies it?

src/common/components/select-repository-search/index.js
```diff
@@ -0,0 +1,29 @@
...
+    let label = this.props.label || 'Search';
+    let placeholder = this.props.placeholder || 'Filter respositories...';
+
+    return (
+      <label htmlFor={ id }>
+        { label }
+        <input
```
bartaz: To make this input component functional later it would need a `value` and `onChange` handler bindings.

We could simply pass them from props:

```js
<input
  value={props.value}
  onChange={props.onChange}
/>

// ...

SearchInput.propTypes = {
  // ...
  value: PropTypes.string.isRequired,
  onChange: PropTypes.func.isRequired
}
```

src/common/components/select-repository-search/search.css
```diff
@@ -0,0 +1,6 @@
+.selectRepositorySearch {
+  border: 1px solid #cdcdcd;
```
bartaz: The colour should be a variable.

src/common/components/select-repositories-page/index.js
```diff
@@ -35,6 +36,7 @@ class SelectRepositoriesPage extends Component {
             <HeadingThree className={ styles.heading }>
               Add repos
             </HeadingThree>
+            <SelectRepositorySearch />
```
bartaz: I wouldn't add it here already in this PR, because if we merge just this we would end up with a component UI that doesn't have any functionality.

We can have all the bits in place, but add it to the page only if the functionality is finished.

richmccartney:
@bartaz if you can take a second look now 😄

src/common/components/search-input/index.js
```diff
@@ -0,0 +1,26 @@
...
+      />
+    </label>
+  );
+}
+
+SearchInput.propTypes = {
+  id: PropTypes.string.isRequired,
+  label: PropTypes.string,
+  onChange: PropTypes.func.isRequired,
+  placeholder: PropTypes.string.isRequired,
```
bartaz: Not sure if we need to make placeholder required. At least repos search input doesn't have any on designs.

src/common/components/search-input/index.js
```diff
@@ -0,0 +1,26 @@
...
+import styles from './search.css';
+
+export default function SearchInput(props) {
+  return (
+    <label htmlFor={ props.id }>
+      { props.label }
+      <input
+        id={ props.id }
+        className={ styles.selectRepositorySearch }
+        onChange={props.onChange}
```
bartaz: Just a nitpicking, but it would be nice to consistently use spaces inside of those values.

We don't really have a strict code style around it, so I'm fine with either `{props.onChange}` or `{ props.onChange }`, but it would be nice to have it consistently in whole component.

src/common/components/search-input/search.css
```diff
@@ -0,0 +1,10 @@
+$border-color: $mid-grey;
+
+.selectRepositorySearch {
+  border: 1px solid $border-color;
+  border-radius: 2px;
+  margin-left: 1rem;
+  outline: none;
+  padding: .275rem .75rem;
+  vertical-align: baseline;
+}
```
bartaz: There is still the 'search' icon missing from the input, like on designs from #715.

Maybe that one will do: https://assets.ubuntu.com/v1/4a72426e-search-icon-grey.svg?

richmccartney:
@bartaz should all be fixed now 😸

src/common/components/search-input/search.css
```diff
@@ -0,0 +1,14 @@
+$border-color: $mid-grey;
+
+.searchInput {
+  background-image: url('https://assets.ubuntu.com/v1/4a72426e-search-icon-grey.svg');
+  background-position: right 4px bottom 2px;
```
bartaz: Not that it looks bad, but positioning of search icon in bottom corner seems a little bit off:
screen shot 2017-07-06 at 16 24 44

And designs have it centered:

richmccartney: @bartaz I'm going to fix this in the next branch as I have all the functionality working now :)

PR #876, opened 2017-07-06
Add filtering repos to 'Add repos' page

richmccartney:
## Done

- Creates filter state and component in the repos list
- Applies filter state when rendering the repo list
- Resolves a few styling issues for the search component

## QA

- Going to the add repos listing, filter the list by entering a username, organisation or repo and the list should filter automatically.

## Issue / Card

#867


src/common/components/select-repository-list/index.js
```diff
@@ -69,6 +71,10 @@ export class SelectRepositoryListComponent extends Component {
     this.props.onRefresh();
   }
 
+  updateSearch(event) {
+    this.setState({ search: event.target.value.substr(0, 20) });
```
bartaz: Why substring of 20? Do we need to limit the search text?

src/common/components/select-repository-list/index.js
```diff
@@ -152,6 +166,14 @@ export class SelectRepositoryListComponent extends Component {
 
     return (
       <div>
+        <div className={ styles.repoSearch }>
```
bartaz: As search input is part of heading of this dialog it shouldn't be rendered here and positioned absolutely.

We need to put the `SearchInput` where it belongs (in `SearchRepositoriesPage` component) and only pass `searchText` or something like this as a property to `SearchRepositoriesList`.

src/common/components/select-repository-list/index.js
```diff
@@ -152,6 +166,14 @@ export class SelectRepositoryListComponent extends Component {
...
+          <SearchInput
+            id="search-repos"
+            placeholder="Filter all repositories..."
```
bartaz: Designs don't have placeholder, so I would suggest removing it.

src/common/components/select-repository-list/index.js
richmccartney: Ah good point I can't remember why I had added this there..

bartaz:
And, just BTW - it's not currently doing anything with 'redux state', so PR title is a little bit misleading.

The concept (to keep search term in component state) is ok as a first step and will help us make sure that everything works, but after that I'd propose working together to move filtering functionality out of component to redux.

@richmccartney Because of that I propose you don't spend too much time adding tests to this PR, just fix the conflicts and review comments and then we will merge it and work on another branch to move functionality to redux.

src/common/components/select-repositories-page/index.js
```diff
@@ -21,6 +30,10 @@ class SelectRepositoriesPage extends Component {
     }
   }
 
+  updateSearch(event) {
+    this.setState({ search: event.target.value.substr() });
```
bartaz: I guess `substr` doesn't need to be called at all here?

src/common/components/select-repositories-page/index.js
```diff
@@ -35,9 +48,16 @@ class SelectRepositoriesPage extends Component {
             <HeadingThree className={ styles.heading }>
               Add repos
             </HeadingThree>
+            <SearchInput
+              id="search-repos"
+              placeholder="Filter all repositories..."
```
bartaz: As mentioned before, designs don't have any placeholder in this field.

src/common/components/select-repositories-page/select-repositories-page.css
```diff
@@ -16,6 +16,7 @@
 .selectRepositoriesBox {
   border: 1px solid $mid-grey;
   border-radius: 2px;
+  position: relative;
```
bartaz: Is this needed for anything now? I guess it's just a leftover from previous absolute positioning.

bartaz:
@richmccartney The tests started failing on "TypeError: Cannot read property 'toLowerCase' of undefined" (because new searchTerm prop is not required and we don't pass it in tests).

We could either default it to empty string, or maybe even better - just skip whole filtering part when search term is empty:

```js
let filteredRepos = ids;
if (this.props.searchTerm) {
  filteredRepos = ids.filter(
     (id) => {
        return this.props.entities.repos[id].fullName.toLowerCase().indexOf(
            this.props.searchTerm.toLowerCase()) !== -1;
         }
     );
}
```

bartaz:
LGTM, thanks

PR #882, opened 2017-07-12
Adding conditions for search and selected responses when selected or …

### LINKED ISSUE #878: Empty state for repos search
When user is searching through their repos and they enter search text that doesn't match any repos we currently show empty list.

screen shot 2017-07-07 at 16 20 55

This is not very helpful.

We should show some empty state. Maybe just a helpful text explaining that given search didn't return any results.

richmccartney:
…filtered.

## Done

- Added new response for filtered content in the add repos list.

## QA

- Go to add repos and search, filter and select repos to see the new response.

## Issue / Card

Fixes #878

### DIFF
--- a/src/common/components/select-repository-list/index.js
+++ b/src/common/components/select-repository-list/index.js
@@ -128,33 +128,59 @@ export class SelectRepositoryListComponent extends Component {
   }
 
   renderRepoAmount() {
-    const { ids, isFetching } = this.props.repositories;
-    const { selectedRepositories } = this.props;
-    const { filteredRepos } = this.props;
-    let filteredAmount = null;
+    const { ids, isFetching, searchTerm } = this.props.repositories;
+    const { selectedRepositories, filteredRepos } = this.props;
 
     // Return nothing until isFetching completes
     if (isFetching || ids.length === 0) {
       return;
     }
 
-    if (filteredRepos.length < ids.length && filteredRepos.length > 1) {
-      filteredAmount = <span>{ filteredRepos.length } matches in </span>;
-    } else if (filteredRepos.length < ids.length && filteredRepos.length === 1) {
-      filteredAmount = <span>{ filteredRepos.length } match in </span>;
-    } else if (filteredRepos.length === 0) {
-      filteredAmount = <span>no matches </span>;
+    // Variables required to calucatate selected amount and filterd amount.
+    let reposSelected = '';
+    let reposFiltered = '';
+
+    // Set reposSelected to value amount of selectedRepositories
+    if (selectedRepositories.length > 0) {
+      reposSelected = `${ selectedRepositories.length } selected`;
     }
+
+    // Set reposFiltered message dependant on value of searchTerm
+    if (searchTerm) {
+
+      if (filteredRepos.length === 0) {
+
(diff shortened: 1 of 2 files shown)


src/common/components/select-repository-list/index.js
```diff
@@ -130,18 +130,29 @@ export class SelectRepositoryListComponent extends Component {
...
+    const { filteredRepos } = this.props;
+    let filteredAmount = null;
 
     // Return nothing until isFetching completes
     if (isFetching || ids.length === 0) {
       return;
     }
 
+    if (filteredRepos.length < ids.length && filteredRepos.length > 1) {
+      filteredAmount = <span>{ filteredRepos.length } matches in </span>;
```
bartaz: We don't need to do `<span>` here. There is no need to complicate it. `filteredAmount` can be simply a string that we will put into `<strong>` when returning from this function.

src/common/components/select-repository-list/index.js
```diff
@@ -130,18 +130,29 @@ export class SelectRepositoryListComponent extends Component {
...
+      filteredAmount = <span>no matches </span>;
+    }
+
     // Return selected repos amount and the total
     return (
       <strong>
-        { (selectedRepositories.length > 0) &&
-          <span>{ selectedRepositories.length } selected of </span>
+        { selectedRepositories.length > 0 &&
+          <span>{ selectedRepositories.length } selected, </span>
```
bartaz: We cannot assume there is always `,` there.

If there is no search we should keep showing what we used to: "1 selected of 42 repos" (with "of" instead of ",").

src/common/components/select-repository-list/index.js
```diff
@@ -130,18 +130,29 @@ export class SelectRepositoryListComponent extends Component {
...
     if (isFetching || ids.length === 0) {
       return;
     }
 
+    if (filteredRepos.length < ids.length && filteredRepos.length > 1) {
+      filteredAmount = <span>{ filteredRepos.length } matches in </span>;
+    } else if (filteredRepos.length < ids.length && filteredRepos.length === 1) {
+      filteredAmount = <span>{ filteredRepos.length } match in </span>;
+    } else if (filteredRepos.length === 0) {
+      filteredAmount = <span>no matches </span>;
```
bartaz: "no matches" needs to be in sentence case if there are no selected repos: "No matches in 42 repos".

test/unit/src/common/components/select-repository-list/t_select-repository-list.js
```diff
@@ -95,7 +96,7 @@ describe('<SelectRepositoryListComponent /> instance', function() {
     });
 
     it('should show not message about 0 selected repos', function() {
-      expect(wrapper.html()).toNotInclude('0 selected of');
+      expect(wrapper.html()).toNotInclude('0 selected,');
```
bartaz: I think we shouldn't just check if "0 selected" doesn't show up. The rest ("," or "of") depends on search which is separate feature and it would be best not depending this test on it.

So: `expect(wrapper.html()).toNotInclude('0 selected');`

richmccartney:
🎉

PR #912, opened 2017-08-02
Implementing a repository poller.

cprov:
## Done

Implementing a snap repository poller script.

It should poll registered repositories and their corresponding GH-hosted parts daily and trigger snap builds if any of them changes.

## QA

```npm run build:scripts 2>&1 > /dev/null && npm run poll-repositories -- --env=environments/dev.env```

### CODE REVIEW

File: src/server/handlers/github.js [hunk 1]
```diff
@@ -321,6 +321,19 @@ const getRequest = (owner, name, token, secret) => {
...
+
+export const hasRepoChanged = async (repositoryUrl, since, token) => {
+  if (since === undefined || !since) {
+    throw new Error('`since` must be an ISO8601 value.');
+  }
+  const { owner, name } = parseGitHubRepoUrl(repositoryUrl);
+  const uri = `/repos/${owner}/${name}/commits?since=${since}`;
+  const response = await requestGitHub.get(uri, { token, json: true });
+  await checkGitHubStatus(response);
+  return response.body.length > 0;
```
evandandrea: The GitHub `/commits` API will return an empty list when there are no commits since `${since}`.

src/server/handlers/github.js
```diff
@@ -321,6 +321,19 @@ const getRequest = (owner, name, token, secret) => {
...
 
+
+export const hasRepoChanged = async (repositoryUrl, since, token) => {
+  if (since === undefined || !since) {
+    throw new Error('`since` must be an ISO8601 value.');
+  }
+  const { owner, name } = parseGitHubRepoUrl(repositoryUrl);
+  const uri = `/repos/${owner}/${name}/commits?since=${since}`;
+  const response = await requestGitHub.get(uri, { token, json: true });
+  await checkGitHubStatus(response);
```
evandandrea: This doesn't handle exceeding the rate limit (403). You can check distance from the limit at https://api.github.com/rate_limit.

src/server/handlers/github.js
```diff
@@ -321,6 +321,19 @@ const getRequest = (owner, name, token, secret) => {
...
 };
 
+
+export const hasRepoChanged = async (repositoryUrl, since, token) => {
+  if (since === undefined || !since) {
+    throw new Error('`since` must be an ISO8601 value.');
+  }
+  const { owner, name } = parseGitHubRepoUrl(repositoryUrl);
+  const uri = `/repos/${owner}/${name}/commits?since=${since}`;
+  const response = await requestGitHub.get(uri, { token, json: true });
```
evandandrea: Shouldn't we cache using ETags? GitHub does not count cache hits (304) against the rate limit.

src/common/helpers/poller.js
```diff
@@ -0,0 +1,11 @@
+
+// Extracts unique repository URLs from a given (parsed) snapcraft.yaml
+export function extractPartsToPoll(snapcraft_yaml) {
+  const parts = Object.values(snapcraft_yaml.parts || {});
+  const gh_parts = parts.filter(function (e) {
+    return (e['source-type'] == 'git' &&
```
evandandrea: `source-type` isn't always set. Snapcraft does its best to [guess in that situation](https://github.com/snapcore/snapcraft/blob/1fdc9b39d96ffd043911a4586bd90c7cfb5e07b0/snapcraft/internal/sources/__init__.py#L176), doing so correctly if the ends in `.git`, for example.

src/server/handlers/github.js
cprov: Yup, seems sufficiently cheap and effective. The individual commit representation is a bit fat, but even if when there are a lot of bulk commits the result is paginated.

evandandrea: Right, but `response.body.length` will be > 0 when there are no commits since `${since}` because the body will be `[ ]`.

src/common/helpers/poller.js
```diff
@@ -0,0 +1,45 @@
...
+
+
+const logger = logging.getLogger('poller');
+
+
+// Extracts unique repository URLs from a given (parsed) snapcraft.yaml
+export function extractPartsToPoll(snapcraft_yaml) {
+  const parts = Object.values(snapcraft_yaml.parts || {});
+  const gh_parts = parts.filter(function (p) {
+    return ((p.source || '').startsWith('https://github.com/') &&
```
evandandrea: `||`

Not all GitHub `source:` entries will end in `.git`.

src/server/handlers/github.js
```diff
@@ -321,6 +322,20 @@ const getRequest = (owner, name, token, secret) => {
...
+
+export const hasRepoChanged = async (repositoryUrl, last_updated_at, token) => {
+  if (last_updated_at === undefined || !last_updated_at) {
+    throw new Error('`last_updated_at` must be given.');
+  }
+  const since = moment(last_updated_at).toISOString();
+  const { owner, name } = parseGitHubRepoUrl(repositoryUrl);
+  const uri = `/repos/${owner}/${name}/commits?since=${since}`;
+  const response = await requestGitHub.get(uri, { token, json: true });
+  await checkGitHubStatus(response);
```
evandandrea: This doesn't handle exceeding the rate limit (403). You can check distance from the limit at https://api.github.com/rate_limit.

src/server/handlers/github.js
```diff
@@ -321,6 +322,20 @@ const getRequest = (owner, name, token, secret) => {
...
 
+
+export const hasRepoChanged = async (repositoryUrl, last_updated_at, token) => {
+  if (last_updated_at === undefined || !last_updated_at) {
+    throw new Error('`last_updated_at` must be given.');
+  }
+  const since = moment(last_updated_at).toISOString();
+  const { owner, name } = parseGitHubRepoUrl(repositoryUrl);
+  const uri = `/repos/${owner}/${name}/commits?since=${since}`;
+  const response = await requestGitHub.get(uri, { token, json: true });
```
evandandrea: Shouldn't we cache using ETags? GitHub does not count cache hits (304) against the rate limit.

src/common/helpers/poller.js
cprov: Correct. Let's drop the '.git' clause for now and go only with `startsWith('https://github.com/')`. We can have a better idea of the currently registered repos when it reaches QA.

src/server/handlers/github.js
cprov: Good idea, added basic support for `If-Modified-Since` based on https://developer.github.com/v3/#conditional-requests.
We can continue to use a single global `last_updated_at` timestamp for snap repositories.

src/server/handlers/github.js
cprov: Yup, will work on something clever for handling 403 & X-RateLimit-Remaining / -Reset  response headers.
This associated with GH authentication to escape 60 req/h limit

src/server/handlers/github.js
cprov: response.body is JSON (see request options `{json: true}` above.)

evandandrea: Oops, completely missed that.

src/common/helpers/poller.js
evandandrea: 👍

cprov:
@bartaz could you please take a look on why Travis seems to be using an old raven dependency ?

While here, It'd be great to have a review as well :-)

README.md
```diff
@@ -20,8 +20,29 @@ Go and [register a new OAuth application](https://github.com/settings/applicatio
 
 Once you've created the application, you should be given the client id and the client secret. **Make a note of these**.
 
+Additionally, if the intention is to run the GitHub repository poller (`npm run poll-repositories`), a API token should be acquired so the API calls are submitted to an acceptable rate-limit (5k req/s instead of 60 req/s for anonymous access)::
```
bartaz: It's 5k or 60 req per **hour** not per second.

README.md
```diff
@@ -20,8 +20,29 @@ Go and [register a new OAuth application](https://github.com/settings/applicatio
...
+
+  curl -s -u '<GH_USER>' https://api.github.com/authorizations \
+    -H 'Content-Type: application/json' \
+    -H 'X-GitHub-OTP: <GH 2FA CODE IF ENABLED>' \
+    -d '{"client_id": "c54ee622a3fa9622701f", \
+         "client_secret": "2a6010d06c7c90b1fcf6518d72eac0bab64cf35b", \
+	 "note": "Build.s.io poller token"}' | jq '.token' -r
+
+`curl` will ask for you GitHub password (basic-auth) and acquire a 40-char authorization token. Make note of it, it should be set in the running environment as `GITHUB_AUTH_CLIENT_TOKEN`.
```
bartaz: In theory you don't need to generate the token. BSI is a registered app, so it can use its own rate limit just by sending client_id and client_secret with in query of API request: https://developer.github.com/v3/#increasing-the-unauthenticated-rate-limit-for-oauth-applications

We use that in BSI to update snapcraft.yaml data when we recieve push from webhook.

So, talking about rate limits.
It seems that the only place where BSI calls GH on its own (via client_id + client_secret) is when we get webhook push and we update the status of snapcraft.yaml data (to make sure we have a fresh snap name in case it was updated in a commit). So thats the only action that affects BSI own rate limit (5k/h).

If poller script shares the credentials it would take from the same rate limit. That would mean that if pooler sends more than 5k req to GH in an hour BSI will not respond correctly to webhooks.

But, from my understanding, if poller uses separate token (not client_id + client_secret) pair it is likely to have a separate rate limit for that token.

So, on one hand poller script auth may be simplified to not require separate token and just use client_id and client_secret, but that would mean shared rate limit between BSI webapp and poller script.
Having separate token for poller script means more configuration (and need to request the token, authenticate to GH, etc), but it should make sure that webapp and poller have separate rate limits.

README.md
```diff
@@ -20,8 +20,29 @@ Go and [register a new OAuth application](https://github.com/settings/applicatio
 
 Once you've created the application, you should be given the client id and the client secret. **Make a note of these**.
 
+Additionally, if the intention is to run the GitHub repository poller (`npm run poll-repositories`), a API token should be acquired so the API calls are submitted to an acceptable rate-limit (5k req/s instead of 60 req/s for anonymous access)::
+
+  curl -s -u '<GH_USER>' https://api.github.com/authorizations \
+    -H 'Content-Type: application/json' \
+    -H 'X-GitHub-OTP: <GH 2FA CODE IF ENABLED>' \
+    -d '{"client_id": "c54ee622a3fa9622701f", \
+         "client_secret": "2a6010d06c7c90b1fcf6518d72eac0bab64cf35b", \
```
bartaz: Hopefully this is not a real client secret (or at least not one we don't want leaked).
It would be good not to have specific ids in README and replace them with some <PLACEHOLDERS>.

README.md
```diff
@@ -72,3 +102,21 @@ ubuntu@build-snapcraft-io:~/work$ npm test
 
 Emacs is confused if 'build.snapcraft.io' is used in the project directory
 name, using 'build-snapcraft-io' restores sanity.
+
+## Updating dependencies
+
+Managing dependencies with npm can be a little .... interesting, especially if
```
bartaz: As much as npm can be a pain sometimes we probably don't need opinions in README.

README.md
```diff
@@ -72,3 +102,21 @@ ubuntu@build-snapcraft-io:~/work$ npm test
...
+
+First, you must add or update the dependency in package.json. This file
+contains both a 'dependencies' section (for run-time dependencies) and a
+'devDependencies' section (for build-time dependencies). Edit the appropriate
+section to add/remove/edit the dependencies that you care about. Version
+numbers are specified using a [special syntax](https://docs.npmjs.com/misc/semver#prerelease-identifiers).
+
+Next, run `npm update` to get the new dependencies locally.
+
+Finally, run `./node_modules/shonkwrap/shonkwrap` (I wish I was joking) to
```
bartaz: Not sure if updating dependencies and docs about shronkwrap should be part of this PR, but as they are already here...

Yes, the need for 'shonkwrap' is quite peculiar in npm (it's said to be much better in newer versions of npm, but we are not there yet) and usage of shonkwrap was not really documented anywhere (just silently assumed by devs working on this project). So adding it to dev dependencies is a good point, so it is not necessary as external tool.

Taking it a step further (to avoid calling it directly out from `./node_modules/...`) we could add a npm script to package.json, like:

```json
{
  scripts: {
      "shonkwrap": "shonkwrap"
  }
}
```

npm scripts can read project dependencies, so it could be run as `npm run shonkwrap` instead of accessing from the path.

npm-shrinkwrap.json
```diff
@@ -47,26 +50,43 @@
...
     "aws4": {
-      "version": "1.5.0"
-    },
-    "babel-code-frame": {
-      "version": "6.16.0"
+      "version": "1.6.0"
     },
     "babel-messages": {
-      "version": "6.8.0"
+      "version": "6.23.0",
```
bartaz: I don't thing that updating **all** dependencies here is a good thing to do.

If some new packages are needed by poller script they should be added, if some package needs updating for this feature (raven?) only it should be updated.
Updating all dependencies will affect other parts of BSI and may cause some unexpected changes and bugs (if they are introduced by these dependencies).

Especially if we want to first dry run poller we want to have this merge to have as little impact on BSI webapp as possible. Updating all dependencies is a big impact that we don't really control.

README.md
cprov: Right, fixed.

README.md
cprov: As real as my local creds.  Redacted.

README.md
cprov: Current `GITHUB_AUTH_CLIENT_TOKEN` is optional, it will give us enough flexibility to move forward. I presume we do want the safety of isolated rate-limits for the webhook processor and the poller.

README.md
cprov: paragraph dropped.

README.md
cprov: right, npm shonkwrap entrypoint added. Thanks for the advice.

npm-shrinkwrap.json
cprov: Thanks for the instructions, Bartek. I think I've cleaned this up by reverting the dep changes, installed new deps specifically `npm install --save async-lock && npm install --save raven@2.1.1` and finally `npm run shonkwrap`.

src/common/helpers/poller.js
```diff
@@ -0,0 +1,272 @@
+import moment from 'moment';
+
```
bartaz: `src/common` is meant for code shared between server and client app, which `poller.js` isn't.

I think it would be better moved to `src/server/` or somewhere around there - as it seems it imports quite a bunch of modules from server anyway.

src/common/helpers/poller.js
```diff
@@ -0,0 +1,272 @@
...
+   */
+  static fromSnapcraftPart(part) {
+    // TODO: Warn if source-commit or source-subdir are set, since we don't
+    //       support these.
+    // TODO: Not sure if we can support setting tags _and_ branch in the same
+    //       part.
+    const gh_repo_prefix = conf.get('GITHUB_REPOSITORY_PREFIX');
+    if (part.source == undefined) {
+      logger.info('Skipping part with no source set.');
+    } else if (part.source.startsWith(gh_repo_prefix)) {
```
bartaz: GH repo prefix is configured to be "https://github.com" so it's only https repo links that we support, we don't allow to link via "git@" URLs?

src/common/helpers/poller.js
```diff
@@ -0,0 +1,272 @@
...
+  const sourceParts = parts.map(
+    GitSourcePart.fromSnapcraftPart).filter(part => part != undefined);
+  return Array.from(new Set(sourceParts));
+}
+
+
+/// GitSourcePart encapsulates the relevant information from a snapcraft
+/// source part, and contains the methods to determine whether it's up to
+/// date.
+export class GitSourcePart {
```
bartaz: Quite a lot of code in this file, maybe later it can be split into several files with at least this class being a separate module.

README.md
```diff
@@ -20,8 +20,29 @@ Go and [register a new OAuth application](https://github.com/settings/applicatio
 
 Once you've created the application, you should be given the client id and the client secret. **Make a note of these**.
 
+Additionally, if the intention is to run the GitHub repository poller (`npm run poll-repositories`), a API token should be acquired so the API calls are submitted to an acceptable rate-limit (5k req/h instead of 60 req/h for anonymous access)::
+
+  curl -s -u '<GH_USER>' https://api.github.com/authorizations \
+    -H 'Content-Type: application/json' \
+    -H 'X-GitHub-OTP: <GH 2FA CODE IF ENABLED>' \
+    -d '{"client_id": "<CLIENT-ID", \
```
bartaz: Just a missing `>` here.

package.json
```diff
@@ -17,7 +17,9 @@
     "test:routes": "BABEL_DISABLE_CACHE=1 HTTP_PROXY= HTTPS_PROXY= TZ=US/Pacific mocha --env=environments/test.env --recursive --require babel-core/register --require babel-polyfill --compilers js:babel-register ./test/setup.js ./test/routes/**/*",
     "migrate:latest": "knex migrate:latest",
     "migrate:rollback": "knex migrate:rollback",
-    "generate-badges": "node scripts/badges.js"
+    "shonkwrap": "shonkwrap",
+    "generate-badges": "node scripts/badges.js",
+    "poll-repositories": "babel --out-dir=dist src scripts/poller.js --copy-files 2>&1 > /dev/null && node dist/scripts/poller.js"
```
bartaz: This builds and runs the poll script. Also it seems that it's build in the build.sh script.

So in theory on production (after whole app is build) calling ./dist/scripts/poller.js is enough, to skip unnecessary build step on prod? Especially that I'm not sure if dev dependencies are available on prod environment.

src/common/helpers/poller.js
cprov: `src/server/scripts/poller.js`

src/common/helpers/poller.js
cprov: Agreed, we have to reorganize / clean it at the earliest opportunity.

src/common/helpers/poller.js
cprov: I don't think SSH access is supported in snapcraft.yaml, but let me check.

README.md
cprov: fixed.

package.json
cprov: Good point, I've create a convenient (quicker) `npm run build:scripts` with that so one can run `npm run build:scripts 2>&1 > /dev/null && npm run poll-repositories -- --env=environments/dev.env` and skip the "webpack" part of the default build process.

README.md
```diff
@@ -53,6 +76,13 @@ npm start -- --env=environments/dev.env
 
 After about 20 seconds, the web application should be available at <http://127.0.0.1:8000>.
 
+
+### Run the repository poller script
+
+```
+npm run poll-repositories -- --env=environments/dev.env
```
cprov: Righto, fixed.

src/server/scripts/poller.js
```diff
@@ -0,0 +1,274 @@
...
+      pollRepoLock.acquire('PROCESS-REPO-SYNC', async () => {
+        const owner = repo.get('owner');
+        const name = repo.get('name');
+
+        // XXX skip repos already updated recently, create a new datetime
+        // column exclusively for the poller.
+        const last_polled_at = repo.get('polled_at') || repo.get('updated_at');
+
+        if (!repo.get('snapcraft_name')) {
+          logger.info(`${owner}/${name}: NO SNAPCRAFT.YAML`);
```
evandandrea: Minor nit, but this is "NO NAME IN SNAPCRAFT.YAML"

src/server/scripts/poller.js
```diff
@@ -0,0 +1,274 @@
...
+          return;
+        }
+
+        if (!repo.get('store_name')) {
+          logger.info(`${owner}/${name}: NOT REGISTERED`);
+          return;
+        }
+
```
evandandrea: What if `repo.get('store_name') != repo.get('snapcraft_name')`?

src/server/scripts/poller.js
```diff
@@ -0,0 +1,274 @@
...
+    token,
+    headers: {
+      'If-Modified-Since': last_polled.format('ddd, MM MMM YYYY HH:mm:ss [GMT]')
+    },
+    json: true
+  };
+
+  const response = await requestGitHub.get(uri, options);
+
+  switch (response.statusCode) {
```
evandandrea: Worth handling 404? I've seen more than one deleted repo in https://launchpad.net/~build.snapcraft.io/+snaps.

src/server/scripts/poller.js
```diff
@@ -0,0 +1,274 @@
...
+      },
+      json: true
+    };
+
+    if (this.branch === 'master' && this.tag == undefined) {
+      // check master branch, no tag.
+      const uri = `/repos/${owner}/${name}/commits?since=${since}`;
+      const response = await requestGitHub.get(uri, options);
+
+      switch (response.statusCode) {
```
evandandrea: 409: repo is empty.
404: repo or user does not exist.

src/server/scripts/poller.js
```diff
@@ -0,0 +1,274 @@
...
+
+    const options = {
+      token,
+      headers: {
+        'If-Modified-Since': last_polled.format('ddd, MM MMM YYYY HH:mm:ss [GMT]')
+      },
+      json: true
+    };
+
+    if (this.branch === 'master' && this.tag == undefined) {
```
evandandrea: This should check for the default branch, not hardcode 'master'.

e.g. https://api.github.com/repos/ethereum/cpp-ethereum/commits shows the commits for the 'develop' branch in https://github.com/ethereum/cpp-ethereum

src/server/scripts/poller.js
```diff
@@ -0,0 +1,274 @@
...
+      // check master branch, no tag.
```
evandandrea: s,master,default,

src/server/scripts/poller.js
cprov: fixed.

src/server/scripts/poller.js
cprov: LP would fail to upload the snap, but I am not even sure the UI would let you do that. Better to be on the safe side and skip repos in that state.

evandandrea: The UI [changes to say "Name mismatch"](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.wyo3cgnn0ubx), but presumably `repo.get('store_name') != repo.get('snapcraft_name')` would appear in state.

src/server/scripts/poller.js
cprov: Right, I've dropped the commits API usage and it's now routing every change check via the branches API right after figuring out what is the 'default' branch.

src/server/scripts/poller.js
cprov: yup, fixed.

src/server/scripts/poller.js
cprov: We are not using the commits API anymore, via branches API anything else than 200 and 304 trigger a generic error.

src/server/scripts/poller.js
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        if (!repo.get('store_name')) {
+          logger.info(`${owner}/${name}: NO NAME REGISTERED IN THE STORE`);
+          return;
+        }
+
+        const store_name = repo.get('store_name');
+        const snapcraft_name = repo.get('snapcraft_name');
+        if (store_name != snapcraft_name) {
```
bartaz: These is an edge case related to that where store_name and snapcraft_name (from db) may mismatch, but a recent commit to repo updated the name in snapcraft.yaml to a correct one.
This script currently relies on the state of BSI database that is updated (for the sake of metrics) when user interacts with the UI.

So in such case this database may contain a different state from what really is in GH and LP/Store.
In theory it would be safer to request the build even if names in db don't match, because they may actually match in the code (updated snapcraft.yaml and registered name).
The issue with that would be that for repos where names really don't match (both in db AND in the code) it would trigger unnecessary builds that will show up in UI as build but not uploaded.

The other side of the coin is that there may be a case where the names do match in the db, but snapcraft.yaml was changed to a different name, so it doesn't match anymore. In such case we would request a build (because we think names match) and it will build and fail to upload too.

bartaz: On the other hand `checkSnapRepository` fetches snapcraft.yaml anyway (to look into parts), so maybe we should not rely on `snapcraft_name` from database, but the current one from freshly fetched snapcraft.yaml.

evandandrea:
Thanks for persisting. I think this is ready to land 🛩

PR #924, opened 2017-08-16
913 icon components

### LINKED ISSUE #913: Create vanilla icon components
https://github.com/vanilla-framework/vanilla-dashboard-theme/issues/35

richmccartney:
## Done

- Created the new Icons react component, this has pre-build icons and the ability to refine any new icons using `<Icon apperance='error' />` or e.g `<IconError />`.
- Replaced all icons on the repository pages to use the new spinner, success, error and copy icons.
- Removed Icon styles from Repository row
- Updated the copy share function to use the new `<IconCopy />`
- Updated the button component to use the new spinner

## QA

- View any new icons at the repository listing, Error / Success. 
- Add a new repo to see the spinner icon working in a button
- Build copy icons have also been updated, check a failed build page to see the new icon.

## Issue / Card

Fixes #913


src/common/components/vanilla-modules/icons/index.js
```diff
@@ -0,0 +1,202 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+import styles from '../../../style/vanilla/css/icons.css';
+
+// Creates icon style class name
+const iconStyle = (icon = '') => {
+  icon = icon ? '--' + icon : '';
```
bartaz: `icon` name here is a bit confusing when it's used for other things (size, color), so maybe a `modifier` would work better as this variable name?

src/common/components/vanilla-modules/button/index.js
```diff
@@ -22,7 +22,7 @@ function createButtonComponent(Component) {
     return (
       <Component {...rest} className={ className }>
         { isSpinner &&
-          <span className={ customStyle.spinner }><Spinner light/></span>
+          <IconSpinner color='inherit-color'/>
```
bartaz: This causes button to change the size when spinner appears.

Spinner in button needs to be placed in a wrapper that is positioned in the center of the button, and text needs to be transparent (not moved out with text indent).

```js
<span className={ customStyle.spinner }><IconSpinner color='inherit-color' /></span>
```

src/common/components/vanilla-modules/button/customButton.css
```diff
@@ -5,14 +5,8 @@
 .hasSpinner .text {
   /* hide text without chaning button width */
   opacity: 0;
-}
-
-.spinner {
```
bartaz: All these styles (and additional 'spinner' wrapper) are needed so that button doesn't change the width when spinner appears.

bartaz:
@richmccartney @kwm14 @mpt This needs design review and possibly some visual design updates.

### 1. New icons

First of all with this PR we start to use new icons (from dashboard theme) in place of current BSI icons.
This is likely fine and expected, but it would be good to have a +1 from design on that.

Before:

screen shot 2017-08-16 at 13 36 16

After:

screen shot 2017-08-16 at 13 35 46

### 2. New copy icon (and conflict with share icons)

Bigger issue is with changed style of copy icon. We are currently presenting copy icon in a dark grey circle (following the style of share icons). But new icon comes in grey and with a removed background looks out of place next to other share icons:

Before:

screen shot 2017-08-16 at 13 32 35

After:

screen shot 2017-08-16 at 13 31 47

We need some visual design input on that.
We either need to change other share icons (twitter) to match copy icon style, or change copy icon to be white on grey background.

src/common/style/vanilla/icons.scss
```diff
@@ -0,0 +1,189 @@
+@import 'vanilla-framework/scss/settings';
+
+@mixin icon-size($size) {
```
bartaz: Do these styles need to be copied here?

Is there a plan to move them to vanilla framework, so we can use icons from there?
If no, maybe we should include dashboard theme and use it from there?

src/common/components/share/styles.css
```diff
@@ -18,8 +18,12 @@
 }
 
 .clipboard {
-  background-image: url(https://assets.ubuntu.com/v1/9cbcbbc7-copy-clipboard.svg);
-  background-size: 12px;
+  background: transparent;
```
bartaz: As commented separately - we need design review for how clipboard should look in share icons.

src/common/components/vanilla-modules/icons/index.js
richmccartney: This is a modifier.. classes in the framework all start with `p-icon--error` for the error icon and sizes or colour are e.g `p-icon--large` so for the classes to work I have applied this. Also the class `p-icon` can be used on a image say an SVG so it'll work the same way if required.

src/common/style/vanilla/icons.scss
richmccartney: For now we do, as these styles are no apart of any NPM package. They will be changed once the conversion of the dashboard theme to vanilla begins

src/common/components/share/styles.css
richmccartney: Yeah I have just bought this to @kwm14 attention to see what we should do with teh share icons.

kwm14: @richmccartney @bartaz share icons were a topic of conversation in our Blog theme inventory workshop last week. The plan is to generate a set of social share icons which is in our backlog of work to be done.

kwm14:
@bartaz we've just had a quick visual/ux review of the copy to clipboard/tweet section. Overall consensus for now is to move the copy to clipboard icon into the command dialog and Twitter remains in it's current position.

Going forward I will think about these issues when tackling social/share icon set for Vanilla.

@richmccartney will implement the changes discussed. Thanks

richmccartney:
@kwm14 thanks, all those changes have been done now. @bartaz if you can review please.

src/common/components/vanilla-modules/button/index.js
```diff
@@ -22,7 +22,7 @@ function createButtonComponent(Component) {
     return (
       <Component {...rest} className={ className }>
         { isSpinner &&
-          <span className={ customStyle.spinner }><Spinner light/></span>
+          <IconSpinner inheritColor className={ customStyle.icon } />
```
bartaz: The spinner still changes the width of the button:

add

To test that just go to Add repos, choose a repo and click 'Add'.

To avoid that spinner (not text) needs to be positioned absolutely over a transparent text.
That's what `<span className={ customStyle.spinner }>` wrapper over the spinner was doing before.

src/common/components/help/help-install-snap.js
```diff
@@ -21,11 +21,17 @@ export default class HelpInstallSnap extends Component {
       <div className={styles.helpFlexWrapper}>
         <HeadingThree className={styles.helpFlexHeading}>{ headline }</HeadingThree>
         <div className={styles.helpText}>
-          <pre>
-            <code className={ styles.cli }>
+          <pre className={styles.cli}>
```
bartaz: These styles need adding everywhere where we show some instructions:

For example when you go to build details page release instructions in the bottom are broken:
screen shot 2017-08-17 at 08 54 48

src/common/components/help/help.css
```diff
@@ -25,7 +26,17 @@
   flex: 1 0 auto;
 }
 
+.helpText {
+  margin-bottom: 1rem;
```
bartaz: Is this margin needed? Why adding this style?

src/common/components/vanilla-modules/icons/index.js
bartaz: @richmccartney I understand that, my point here is just that `icon` name is not very clear here, because it doesn't mean 'icon' but modifier to icon class.

So my only suggestion is to change `icon` to `modifier`:

```js
const iconStyle = (modifier = '') => {
  modifier = modifier ? `--${modifier}` : '';
  // ...
}
```

src/common/components/help/help-install-snap.js
bartaz: Possibly related, going to such page throws a bunch of React warnings:
screen shot 2017-08-17 at 09 04 00

bartaz: Related issue. Copy button is showing up possibly when it shouldn't.

When build fails we show more detailed instructions on how to debug the build:

screen shot 2017-08-17 at 09 11 05

It's a multi line help, where it probably doesn't make much sense to copy it, especially that is contains some optional lines and comments.

src/common/components/vanilla-modules/icons/index.js
richmccartney: Ah I see what you are getting at now. I'll make that change.

src/common/style/vanilla/icons.scss
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    &--xx-large {
+      @include icon-size($sp-xxx-large);
+    }
+  }
+
+  [class*="p-button-"] {
+
+    [class*="p-icon-"] {
```
bartaz: That is some custom styling that I don't think we should put in here.
This file is meant to be taken directly from vanilla (or a theme), adding this here makes it very likely that someone updating this file (or replacing it with one from vanilla) would skip those changes and override them.

Also, it's not a good pattern to have icon pattern style depending on the class from parent pattern.
If it is really needed (and it doesn't seem so for the case of spinner) it should be a modifier on icon pattern, not a dependency on button pattern.

src/common/style/vanilla/icons.scss
```diff
@@ -0,0 +1,198 @@
...
+      height: 13px;
```
bartaz: There is also an issue of spinner in button not really spinning in the center but 'wobbling' around.
I made it bigger to make it more visible:

spin

It seems to be caused by the fact that width and height of the icon are different (only height is set to 13px).
I'm also not sure why it's set to 13px. It seems that spinner looks good as 16px default in the button (at least it's bigger and more visible then).

src/common/style/vanilla/icons.scss
richmccartney: I completely disagree, these are styles that only effect the icon element if a project was to be used where icons were not loaded / used we now have redundant styles living in the `button.scss`.  

I have already ported this over to the dashboard project and there is a PR waiting for review.

bartaz: OK, if that is going into vanilla or dashboard theme I'm fine with that, just for now it seemed like we have some custom styles here that are not in vanilla.

Can you link to a PR with that to keep track?

bartaz:
There is still an issue with showing copy icon next to multiline instructions:

screen shot 2017-08-17 at 16 08 33

Copying those actually ends up as:
```
sudo snap install lxd # if you don’t have LXD already,[object Object],sudo snap install --classic --edge snapcraft # if you don’t have snapcraft already,[object Object],[object Object],git clone ,https://github.com/bartaz/speed-test-snap,[object Object],cd ,speed-test-snap,[object Object],snapcraft cleanbuild --debug
```

There is no much sense in copying this, so I would suggest that copy functionality should be optional (param in the component), so it would be only enabled when we what to show it.

matthewpaulthomas:
The design spec [does not include a Copy button for the debug instructions](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.9zzs16bsn7o8).

richmccartney:
@bartaz can you take a look now please 😄

PR #930, opened 2017-08-21
Showing build upload error message on build details

### LINKED ISSUE #714: "Built, failed to publish" doesn't explain why
screen shot 2017-05-04 at 08 28 23

When a build fails to publish there is no further explanation, and nothing suggests how you can resolve this (wait?).

bartaz:
## Done

Simple showing of build upload error message (as returned by API).

## QA

- Add a snap that builds but fails to publish
- Go the the build page

## Issue / Card

Fixes #714 

## Screenshots

screen shot 2017-08-21 at 11 29 14

### DIFF
--- a/src/common/components/build-history/buildHistory.css
+++ b/src/common/components/build-history/buildHistory.css
@@ -1,0 +1,3 @@
+.notificationWrapper {
+  margin-bottom: 1em;
+}
--- a/src/common/components/build-history/index.js
+++ b/src/common/components/build-history/index.js
@@ -4,7 +4,9 @@ import { connect } from 'react-redux';
 import { snapBuildsInitialStatus } from '../../reducers/snap-builds';
 import BuildRow from '../build-row';
 import { Table, Head, Body, Row, Header } from '../vanilla/table-interactive';
-import { Message } from '../forms';
+import Notification from '../vanilla-modules/notification';
+
+import styles from './buildHistory.css';
 
 export const BuildHistory = (props) => {
   const { repository, success, builds } = props;
@@ -16,7 +18,11 @@ export const BuildHistory = (props) => {
   }
 
   if (!hasBuilds) {
-    return <Message status='info'>This snap has not been built yet.</Message>;
+    return (
+      <div className={styles.notificationWrapper}>
+        <Notification status='information' appearance="information">This snap has not been built yet.</Notification>
+      </div>
+    );
   }
 
   const buildRows = builds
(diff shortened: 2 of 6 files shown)


richmccartney:
@bartaz I've pulled the branch downa nd I have a failing build, but I am not getting the error 🤔 Anything I am doing wrong to show the error?

https://cl.ly/0I1U072y210H

bartaz:
@richmccartney Hmm.. Weird, I forked your repo but for me it built and released just fine:
screen shot 2017-08-22 at 15 40 33

Would need to dig into redux data (and/or network API response) to see what build status your build has.

richmccartney:
@bartaz Following state doesn't show an error message - storeUploadStatus(pin): "Failed to release to channels"

kwm14:
@bartaz @richmccartney 

Regarding your design update to 'Build failed release' section I'm not sure on the user interaction here but;

- if it's a prompt to the user we should be using our [Notification](https://docs.vanillaframework.io/en/patterns/notification) pattern and in this instance it should use `.p-notification--negative`
- but if it's an input field a user types invalid info we should be using our [Forms](https://docs.vanillaframework.io/en/base/forms) pattern applying `.is-error` validation class

matthewpaulthomas:
The release error is not a prompt to the user and it is not an input field. If it looks like either of those things, we should fix the visual design so that it doesn’t. But I don’t think that should block merging this.

The release error is not a notification for the same reasons that the build log is not a notification: it is information about the results of the build, it is not interactive at all, and it does not disappear (either manually or automatically) like a notification does. It should be shown whenever you look at that build page, even weeks or months later.

bartaz:
@kwm14 @matthewpaulthomas Thanks for comments.

Yes, this is not a prompt or notification. So in this sense notification pattern may not be the best choice.

I can add an error icon to this, to make it more clear that it's an error 'message' and make it look less like an input field:

screen shot 2017-08-23 at 09 53 33

kwm14:
@bartaz @matthewpaulthomas 

If your suggesting that it's not a notification or input field then we can't apply a mix of styles to solve the solution here. We can't just create a new pattern for this use case.

As you mention it's a error 'message' so in this case it should use our notification pattern whether it's an error, warning or informative message. This negative variant should be used to convey information that is critical (error) and the user should take action (in this case you can remove the CTA from the notification.

bartaz:
@kwm @matthewpaulthomas @richmccartney 

Using notification pattern it would be like this:

screen shot 2017-08-23 at 11 09 02

kwm14:
@bartaz looking good! Thanks

richmccartney:
Looks great @bartaz +1

evandandrea:
Star; thank you for this!

PR #931, opened 2017-08-22
Stateless poller

cprov:
## Done

Step forward making the poller stateless, it doesn't write on DB anymore, but still reads from it. 
The poller now relies on LP information (datebuild/date_created) for checking `last_built` time. 

## QA

```npm run migrate:latest && npm run build:scripts 2>&1 > /dev/null && npm run poll-repositories -- --env=environments/dev.env```

### CODE REVIEW

File: src/config/defaults.js [hunk 1]
```diff
@@ -13,6 +13,7 @@ export default {
   STORE_DEVPORTAL_URL: 'https://dashboard.snapcraft.io/dev',
   STORE_ALLOWED_CHANNELS: ['edge'],
   STORE_PACKAGE_UPLOAD_REQUEST_LIFETIME: '7200',
+  BUILD_TRIGGER_THRESHOLD: 24,
```
bartaz: It would be worth adding it to the docs: https://github.com/canonical-websites/build.snapcraft.io/blob/master/docs/environment-variables.md

src/server/scripts/poller.js
```diff
@@ -19,54 +20,81 @@ import requestGitHub from '../helpers/github';
...
+  let locked_promises = [];
 
   return db.model('Repository').fetchAll().then(function (results) {
     logger.info(`Iterating over ${results.length} repositories.`);
     results.models.forEach((repo) => {
-      pollRepoLock.acquire('PROCESS-REPO-SYNC', async () => {
+      const p = pollRepoLock.acquire('PROCESS-REPO-SYNC', async () => {
         const owner = repo.get('owner');
         const name = repo.get('name');
+        const store_name = repo.get('store_name');
```
bartaz: I know that not reading from db is not in scope of this PR, but just a reminder that couple lines later we do `const snap = await internalFindSnap(repositoryUrl);` that would give us a more up to date value of `store_name`.

src/config/defaults.js
cprov: Good point, updated the docs and prefixed variable names for POLLER_*

src/server/scripts/poller.js
cprov: Yes, the early returns on local mismatches are going to bite badly. I will work on them at the light of the daily-build requirements too.

src/server/helpers/config.js
```diff
@@ -11,6 +11,8 @@ import dotenv from 'dotenv';
 const CLIENT_SIDE_WHITELIST = [
   'NODE_ENV',
   'BASE_URL',
+  'POLLER_REQUEST_BUILDS',
```
bartaz: Any particular reason for exposing it to client UI?
To make sure we could display it in the UI?

test/unit/src/server/scripts/t_poller.js
```diff
@@ -139,7 +141,18 @@ describe('Poller script helpers', function() {
...
-        const threshold = conf.get('BUILD_TRIGGER_THRESHOLD');
+        // TODO: Exotic dance to re-import `pollRepositories` with the
+        // `POLLER_REQUEST_BUILD` configuration option enabled.
+        const requireWithMockConfig = requireWithMockConfigHelper.bind(
+          null,
+          path.resolve(__dirname, '../../../../../src/server/scripts/poller'),
+          '../helpers/config'
+        );
+        const pollRepositoriesMock = requireWithMockConfig({
+          POLLER_REQUEST_BUILD: true
```
bartaz: There is a typo here (missing S in `POLLER_REQUEST_BUILDS`) which causes this test to fail.

cprov: Bah! Sorry must have typo-ed it before committing. Fixed.

src/server/helpers/config.js
cprov: We need that to override it on the script commandline e.g. `npm run poll-repositories -- --env=environments/dev.env --POLLER_REQUEST_BUILDS`

bartaz: I'm not sure if I understand.

`CLIENT_SIDE_WHITELIST` is a list of config variables that are whitelisted to be sent to client app in the browser. It's meant to make part of the config shared between browser and server.

This means that all the variables in `CLIENT_SIDE_WHITELIST` are sent in HTML to the browser and are available by JavaScript there:
screen shot 2017-08-28 at 10 56 48

At this stage it doesn't seem that we need to expose `POLLER_REQUEST_BUILDS` or `POLLER_BUILD_THRESHOLD` to client side app.

I also don't think that adding them to this list is needed to make it possible to override them in command line.
I tried running `npm run poll-repositories -- --env=environments/dev.env --POLLER_BUILD_THRESHOLD=1` and it correctly overridden value of this var even if it was not in `CLIENT_SIDE_WHITELIST`.

If by any chance `CLIENT_SIDE_WHITELIST` is needed for anything else then exposing config items to client side app it's a bug somewhere in our config code.

cprov: @bartaz You are right, I misunderstood the discrete list of arguments mechanics. `yargs` does the heavy-lifting to process.env on its own. Thanks for pointing this out.

PR #932, opened 2017-08-22
911 forms component

### LINKED ISSUE #911: Create new vanilla forms component
## Summary

- Include the new dashboard theme form styles into BSI.
- SCSS to be found here https://github.com/vanilla-framework/vanilla-dashboard-theme/blob/develop/scss/_patterns_forms.scss

richmccartney:
## Done

- Included the new styles from the dashboard theme to BSI
- Created new form components based off the original form components
- Select, input, checkbox, radio button and textareas now available for us

## QA

- QA code for each component

## Issue / Card

Fixes #911

### DIFF
--- a/src/common/components/vanilla-modules/forms/form-group-input.js
+++ b/src/common/components/vanilla-modules/forms/form-group-input.js
@@ -1,16 +1,14 @@
-import React, { Component, PropTypes } from 'react';
-import classNames from 'classnames';
+import React, { PropTypes } from 'react';
 
 import styles from '../../../style/vanilla/css/forms.css';
 
 export default function FormGroupInput(props) {
   const { disabled, name, label, type, placeholder } = props;
   const id = `ID_${name}`;
-  const status = props.touched ? ( props.valid ? 'success' : 'error' ) : null;
-  const statusStyle = `is-${status}`;
+  const status = props.touched ? ( props.valid ? 'p-form-validation is-success' : 'p-form-validation is-error') : null;
 
   return (
-    <div className={ styles['p-form__group'] }>
+    <div className={ `${styles['p-form__group']} ${styles[status]}` }>
       <label htmlFor={ id } className={ styles['p-form__label'] }>
         { label }
       </label>
@@ -27,7 +25,11 @@ export default function FormGroupInput(props) {
           onBlur={ props.onBlur }
           value={ props.value || '' }
         />
-        { props.errorMsg }
+        { props.errorMsg &&
+          <p className={ styles['p-form-validation__message'] } role="alert">
+            <strong>Error:</strong> { props.errorMsg }
+          </p>
+        }
       </div>
     </div>
   )
@@ -43,7 +45,10 @@ FormGroupInput.propTypes = {
   sensitive: PropTypes.bool,
   valid: PropTypes.bool,
   touched: PropTypes.bool,
-  value: PropTypes.string,
+  value: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+  ]).isRequired,
   errorMsg: PropTypes.string,
   onChange: PropTypes.func,
   onBlur: PropTypes.func
--- a/src/common/components/vanilla-modules/forms/form-group-input.js
+++ b/src/common/components/vanilla-modules/forms/form-group-input.js
@@ -32,7 +32,7 @@ export default function FormGroupInput(props) {
         }
       </div>
     </div>
-  )
+  );
 }
 
 FormGroupInput.propTypes = {
--- a/src/common/components/vanilla-modules/forms/form-group-input.js
+++ b/src/common/components/vanilla-modules/forms/form-group-input.js
@@ -1,0 +1,50 @@
+import React, { Component, PropTypes } from 'react';
+import classNames from 'classnames';
+
+import styles from '../../../style/vanilla/css/forms.css';
+
+export default function FormGroupInput(props) {
+  const { disabled, name, label, type, placeholder } = props;
+  const id = `ID_${name}`;
+  const status = props.touched ? ( props.valid ? 'success' : 'error' ) : null;
+  const statusStyle = `is-${status}`;
+
+  return (
+    <div className={ styles['p-form__group'] }>
+      <label htmlFor={ id } className={ styles['p-form__label'] }>
+        { label }
+      </label>
+      <div className={ styles['p-form__control'] }>
+        <input
+          id={ id }
+          name={ props.sensitive ? null : name }
+          data-name={ name }
+          type={ type }
+          required={ props.required }
+          disabled={ disabled }
+          placeholder={ placeholder }
+          onChange={ props.onChange }
+          onBlur={ props.onBlur }
+          value={ props.value || '' }
+        />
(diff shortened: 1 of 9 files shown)


src/common/style/vanilla/forms.scss
```diff
@@ -0,0 +1,107 @@
+@import 'vanilla-framework/scss/settings';
+@import 'vanilla-framework/scss/base_forms.scss';
+
+@mixin theme-p-forms {
```
bartaz: Where do these theme styles come? Is it dashboard theme?

It would be good to have a comment about it here to know where to look for it in case of updates.

src/common/components/vanilla-modules/forms/form-group-input.js
```diff
@@ -0,0 +1,55 @@
...
+          id={ id }
+          name={ props.sensitive ? null : name }
+          data-name={ name }
+          type={ type }
+          required={ props.required }
+          disabled={ disabled }
+          placeholder={ placeholder }
+          onChange={ props.onChange }
+          onBlur={ props.onBlur }
+          value={ props.value || '' }
```
bartaz: You marked `value` as required property, but you check if it has any value. There are couple of issues with that.

One - you cannot have `value=0`, because 0 will be false and it will end up as empty string.

But more importantly if you require value property and bind input value this property it cannot be changed by user unless there an onChange handler that updates the value.
As a result you can't type anything to such input (it's practically read only).

Binding input values is a good pattern in general, but may make it difficult to make simple forms that don't need to be bound to application state.

More on that if you're interested: https://facebook.github.io/react/docs/forms.html#controlled-components

But shortly speaking, ideally value should not be required to make it possible to create so called 'uncontrolled' input that user can type in without any onChange handler making it happen.

src/common/components/vanilla-modules/forms/form-group-select.js
```diff
@@ -0,0 +1,61 @@
+import React, { PropTypes } from 'react';
+
+import styles from '../../../style/vanilla/css/forms.css';
+
+export default function FormGroupSelect(props) {
```
bartaz: There are small visual bugs with Select, reported to vanilla: https://github.com/vanilla-framework/vanilla-framework/issues/1289

src/common/components/vanilla-modules/forms/form-group-select.js
```diff
@@ -0,0 +1,61 @@
...
+        <select
+          id={ id }
+          name={ props.sensitive ? null : name }
+          data-name={ name }
+          required={ props.required }
+          disabled={ disabled }
+          onChange={ props.onChange }
+          value={ props.selectedOption }
+        >
+          <option value="">{ placeholder }</option>
```
bartaz: Is placeholder required? Could there be a select without placeholder item?

src/common/components/vanilla-modules/forms/form-group-select.js
```diff
@@ -0,0 +1,61 @@
...
+          { props.options.map(opt => {
+            return (
+              <option
+                key={ opt }
+                value={ opt }
```
bartaz: Making `value` and option text being the same may not be good in some cases.

For example it's quite easy to imagine a select for a list of countries where we'd like to have options values be iso codes (PL, US, UK) while text being a name of the country.

Putting any arbitrary structure to this data also doesn't seem very flexible.

Maybe the easiest way would be to allow devs to use `<option>` as a child of `<FormGroupSelect>`.

So it would end up as:

```jsx
<FormGroupSelect name='country'>
   <option>-- Optional placeholder --</option>
   <option value="PL">Poland</option>
   <option value="UK">United Kingdom</option>
</FormGroupSelect>
```

src/common/components/vanilla-modules/forms/form-group-input.js
```diff
@@ -0,0 +1,55 @@
...
+
+import styles from '../../../style/vanilla/css/forms.css';
+
+export default function FormGroupInput(props) {
+  const { disabled, name, label, type, placeholder } = props;
+  const id = `ID_${name}`;
+  const status = props.touched ? ( props.valid ? 'p-form-validation is-success' : 'p-form-validation is-error') : null;
+
+  return (
+    <div className={ `${styles['p-form__group']} ${styles[status]}` }>
```
bartaz: It's the same for all other form components, so I'm not repeating it there.

src/common/components/vanilla-modules/forms/form-group-input.js
bartaz: The same issue is also repeated in other component, so not commenting there on that.

src/common/components/vanilla-modules/forms/form-group-input.js
```diff
@@ -0,0 +1,55 @@
...
+    </div>
+  );
+}
+
+FormGroupInput.propTypes = {
+  name: PropTypes.string,
+  label: PropTypes.string,
+  placeholder: PropTypes.string,
+  required: PropTypes.bool,
+  disabled: PropTypes.bool,
```
bartaz: I guess we did it like that before as well, but right now as a convention I'd rather name bool props as `isDisabled`.

On the other hand with props like this (that link directly to a DOM property) it's quite clear to have it as `disabled`, as it speaks for itself.
So I'm not exactly sure what to do with that, but kinda opening a discussion ;)

src/common/components/vanilla-modules/forms/form.js
```diff
@@ -0,0 +1,29 @@
...
+      { props.children }
+    </form>
+  );
+}
+
+Form.propTypes = {
+  children: PropTypes.node,
+  onSubmit: PropTypes.func,
+  stacked: PropTypes.bool,
+  inline: PropTypes.bool
```
bartaz: Can a form be both `stacked` and `inline` at the same time?
Having 2 separate props for that kinda suggest that.

If it can't (at least thats how I understand it), maybe it should be a single `appearance` prop with 2 possible values?

src/common/style/vanilla/forms.scss
richmccartney: Hey @bartaz so these will be pulled straight from VF soon. https://github.com/vanilla-framework/vanilla-framework/pull/1290 This PR will include the new styles hoping this week 🤞

src/common/components/vanilla-modules/forms/form-group-select.js
richmccartney: For a select dropdown I would prefer there was an option like this. As the placeholder attribute doesn't work with `<select>` this is the better option.

src/common/components/vanilla-modules/forms/form-group-select.js
richmccartney: Ah okay I agree and I will make the change.

src/common/components/vanilla-modules/forms/form.js
richmccartney: Ah yes you are correct, I'll make this change too.

src/common/components/vanilla-modules/forms/form-group-input.js
richmccartney: I've now changed this to be one class now.

Issue #936, opened 2017-08-29
“No repos” shown: “Cannot read property 'text' of undefined”


matthewpaulthomas:
I have about four repos set up in BSI, but the “My repos” table just says “No repos”.

If I choose “Add repos…”, then re-select one of my already-chosen repos, and choose “Add”, the button spins indefinitely.

Every 20 seconds this appears in Firefox’s Developer Console:
`GET XHR https://build.snapcraft.io/api/launchpad/snaps/list [HTTP/1.1 500 Internal Server Error ___ms]`

The full text of the response is:
`{"status":"error","payload":{"code":"internal-error","message":"Cannot read property 'text' of undefined"}}`

The problem occurs only on build.snapcraft.io, not build.staging.snapcraft.io.

cjwatson:
It looks like the memcached units were rebooted and BSI failed to reconnect properly.  I restarted the BSI app servers and they sorted themselves out from there (although you may have to reload the page since your CSRF token may be invalid).

PR #937, opened 2017-08-29
Making sure all errors are properly sent in API responses

bartaz:
## Done

Making sure any error is properly propagated to response, so it's easier to debug issues.

## Issue / Card

Related to #936

### DIFF
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -480,16 +480,9 @@ const internalFindSnaps = async (owner, token) => {
     }
     return snaps;
   } catch (error) {
-    // At least for the moment, we just wrap the error we get from
-    // Launchpad.
-    const text = await error.response.text();
-    throw new PreparedError(error.response.status, {
-      status: 'error',
-      payload: {
-        code: 'lp-error',
-        message: text
-      }
-    });
+    // At least for the moment, we just throw the error and let it be caught
+    // later and sent in the response
+    throw error;
   }
 };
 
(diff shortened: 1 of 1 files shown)


src/server/handlers/launchpad.js
```diff
@@ -480,16 +480,9 @@ const internalFindSnaps = async (owner, token) => {
     }
     return snaps;
   } catch (error) {
-    // At least for the moment, we just wrap the error we get from
-    // Launchpad.
-    const text = await error.response.text();
```
bartaz: We were assuming error will have a `response` object (as any error from LP client would), but if error was not coming from LP client and it didn't have `response` property that would throw here again with `Cannot read property 'text' of undefined` that would be then sent in API response hiding the real error behind it.

src/server/handlers/launchpad.js
```diff
@@ -480,16 +480,9 @@ const internalFindSnaps = async (owner, token) => {
...
-    throw new PreparedError(error.response.status, {
-      status: 'error',
-      payload: {
-        code: 'lp-error',
-        message: text
-      }
-    });
+    // At least for the moment, we just throw the error and let it be caught
+    // later and sent in the response
+    throw error;
```
bartaz: Now we just throw any error and catch it later to send in response.

cjwatson: `try { foo; } catch (error) { throw error; }` is equivalent to just `foo;`, so you might as well remove the `try` block.

Otherwise this is fine.

bartaz: Ah right, I even wanted to do that and then forgot.
Thanks for catching it.

PR #954, opened 2017-09-14
Fixing poller setup in charm.

cprov:
## Done

Fixing poller cronscript (using bash for loading secrets sanely), several fixes on the charm leadership handling and also allowing the relevant poller options to be set via juju config.


src/server/scripts/poller.js
```diff
@@ -34,6 +34,11 @@ export const pollRepositories = (checker) => {
   let pollRepoLock = new AsyncLock();
   let locked_promises = [];
 
+  let poller_request_builds = false;
+  try {
+    poller_request_builds = JSON.parse(conf.get('POLLER_REQUEST_BUILDS'));
+  } catch (e) {}
```
bartaz: Linter is complaining about this empty catch block.

It would be good to at least log something here or add a comment why it's empty (comment should be enough for linter to pass).

cprov: Sorry for letting it pass. It is fixed now, by logging an error message.

src/server/scripts/poller.js
```diff
@@ -82,7 +89,7 @@ export const pollRepositories = (checker) => {
 
           // Do not even check for changes if the snap was already built in the
           // last `POLLER_BUILD_THRESHOLD` interval (typically 24h).
-          const threshold = conf.get('POLLER_BUILD_THRESHOLD');
+          const threshold = parseInt(conf.get('POLLER_BUILD_THRESHOLD'));
```
bartaz: It would be safer to use `parseInt(conf.get('POLLER_BUILD_THRESHOLD'), 10)` to make sure JS is not trying to guess the base of the number based on contents.

src/server/scripts/poller.js
```diff
@@ -103,13 +110,14 @@ export const pollRepositories = (checker) => {
...
           raven_client.captureException(e);
-          logger.error(`${owner}/${name}: FAILED (${e.message})`);
+          logger.error(`${owner}/${name}: FAILED (${e.message || e})`);
         }
         logger.info('==========');
       });
 
       locked_promises.push(p);
     });
+    if (locked_promises.length == 0) { return []; }
```
bartaz: Is this needed?

If `locked_promises` is empty `Promise.all(locked_promises)` would return automatically resolved promise, which is expected I think? Right now it seems that we return promise or empty array.

src/server/scripts/poller.js
cprov: sure, fix on its way.

src/server/scripts/poller.js
cprov: IF block dropped, it doesn't help with the problem we are still seeing with knex + postgres when there are no repositories registered.

bartaz:
@nottrobin @cjwatson Could any of you give charm updates a review? Thanks.

charm/config.yaml
```diff
@@ -32,6 +32,17 @@ options:
     github_webhook_secret:
         type: string
         description: GitHub webhook secret
+    poller_github_auth_token:
+        type: string
+        description: GitHub auth token to be used exclusively by the poller
+    poller_request_builds:
+        type: string
+        description: set to 'true' in other to allow poller to request builds
```
cjwatson: I think that should be "in order", not "in other".

charm/templates/snap-build-poller_cron.j2
```diff
@@ -1,3 +1,4 @@
 # DO NOT EDIT: Managed by juju charm
-42 * * * * {{user}} cd {{code_dir}} && KNEX_CONFIG_PATH={{knex_config_path}} npm run poll-repositories >> {{logs_dir}}/poller.log 2>&1
+SHELL=/bin/bash
```
cjwatson: This wouldn't be necessary if you just changed `source secrets.env` to `. secrets.env` below.  As far as I can see there are no other bashisms.

charm/reactive/snap-build.py
```diff
@@ -116,6 +122,26 @@ def configure(pgsql, cache):
                 'node_env': get_node_env(environment),
                 'db_conn': pgsql.master.uri,
             })
+
+        # Render shell secrets to keep crontab cleaner.
+        render('snap-build-poller_secrets.j2', SECRETS_PATH,
```
cjwatson: Should we perhaps make sure this is owned by a suitable user/group and mode 0600?

charm/config.yaml
cprov: typo, fixed. Thanks

charm/templates/snap-build-poller_cron.j2
cprov: oh, thanks. Fixed.

charm/reactive/snap-build.py
cprov: user/group is already 'root' by default, but I am now forcing both along with perms=0600.

cprov:
@bartaz can you please help me with the test error ? Tests pass locally, but somehow the empty (`Promise.all([])`) returned by `pollRepositories` is not immediately resolving in Travis. Is there a possible nodejs env mismatch ?

bartaz:
@cprov Hmm... I don't have an idea what could be causing differences in promises between local and travis. I will try to investigate.

bartaz:
@cprov I don't think it was any issue with your Promise.

Tests on travis timeout sometimes (he haven't found why yet), and re-running them usually helps - like in this case.

cprov:
@bartaz ah nice! Thanks for sorting it out. Can you go ahead and merge it, please ?

PR #961, opened 2017-09-26
Fix nock request body matching

cjwatson:
nock 9.0.19 started enforcing the specified request body
(https://github.com/node-nock/nock/pull/921), so tighten up our tests to
cope with that.


sparkiegeek:
:+1:

PR #962, opened 2017-09-27
Add missing statements to set up the container

elopio:


src/common/containers/build-details.js
```diff
@@ -37,12 +37,12 @@ class BuildDetails extends Component {
       helpBox = (
         <HelpBox>
           <HelpInstallSnap headline='To debug this build:'>
-            sudo snap install lxd # if you don’t have LXD already<br/>
+            sudo snap install lxd && sudo snap lxd init # if you don’t have LXD already<br/>
```
cjwatson: Shouldn't this be `sudo lxd init`?

elopio: yes, sorry.

src/common/containers/build-details.js
```diff
@@ -37,12 +37,12 @@ class BuildDetails extends Component {
...
         <HelpBox>
           <HelpInstallSnap headline='To debug this build:'>
-            sudo snap install lxd # if you don’t have LXD already<br/>
+            sudo snap install lxd && sudo lxd init # if you don’t have LXD already<br/>
             sudo snap install --classic --edge snapcraft # if you don’t have snapcraft already<br/>
             <br/>
             git clone {repository.url}<br/>
             cd {repository.name}<br/>
-            snapcraft cleanbuild --debug
+            sudo snapcraft cleanbuild --debug
```
cjwatson: I generally feel uncomfortable with advising that people run build tools (as opposed to setup) as root.  Can we instead give people instructions that will get `snapcraft` working for them as an ordinary user?

elopio: According to Stephan, it is now possible to run the lxd snap for the lxd group members. I'm getting a clean machine to try this, because in my current dev station it doesn't work.

elopio:
@cjwatson I've added a step to add the user to the lxd group.

Issue #972, opened 2017-10-06
empty repo gives wrong error message


spruceriver:
I tried using an empty repo and it said my snapcraft.yaml was invalid.

<img width="1063" alt="screen shot 2017-10-06 at 12 26 27 pm" src="https://user-images.githubusercontent.com/31933866/31288178-d55e0c5e-aa91-11e7-9988-e11659f6f259.png">

bartaz:
Thanks for the report. We will look into this.

matthewpaulthomas:
Design spec updated:
> If the repo does not contain a snapcraft.yaml file (including if the repo is empty, #972), the cell should read “Not configured”, and clicking it should expand the row to reveal a panel explaining snapcraft.yaml.

bartaz:
Fixed by #973

PR #973, opened 2017-10-10
Fix showing repositories not available on GH

### LINKED ISSUE #968: Can't see any repos in BSI if repo was removed from GitHub
## Summary

Found then debugging #925 and quite likely being the main cause of the issues of @kz6fittycent
If user removes from GitHub any repos previously added to BSI, BSI will fail to show ANY repos to this user simply showing empty ("No repos") My repos page.
Any attempt to add repos again will fail, because there repos exist on Launchpad but are not displayed to the user due to this bug.

## Process

1. Have some dummy repository on GH
2. Add this dummy repo to BSI (also have other repos added to BSI)
3. Remove dummy repo from GH
4. Refresh My repos page
5. No repos will be shown

In browser console API call to list snaps returns with 404:
http://localhost:8000/api/launchpad/snaps/list?owner=bartaz
`{"status":"error","payload":{"code":"github-snapcraft-yaml-not-found","message":"Cannot find snapcraft.yaml in this GitHub repository"}}`

Currently the only way to workaround this issue is manually removing repository from Launchpad.

## Expected

If repos were removed from GH they should simply be ignored (not throw an error).
They should appear on the repos list in BSI (because they still exist in Launchpad) and user should be able to remove (from BSI) on their will.

## Screenshot

screen shot 2017-10-05 at 11 05 16

bartaz:
## Done

Updated getting repo data (branch and snapcraft.yaml) not to fail when repo doesn't exist.

Marking repo with "Error" instead of "Not configured" when it doesn't exist (or is not accessible) on GitHub.

## QA

- Sign in
- Add repository to BSI
- Remove this repo from GH (or make it private)
- Go back to "My repos" page

## Issue / Card

Fixes #968 

## Screenshots

screen shot 2017-10-11 at 13 52 24

### DIFF
--- a/src/server/handlers/github.js
+++ b/src/server/handlers/github.js
@@ -6,14 +6,14 @@ import logging from '../logging';
 import requestGitHub from '../helpers/github';
 import { conf } from '../helpers/config';
 import { getMemcached } from '../helpers/memcached';
-import { checkGitHubStatus, internalGetSnapcraftYaml } from './launchpad';
+import { internalGetSnapcraftYaml } from './launchpad';
 import { parseGitHubRepoUrl } from '../../common/helpers/github-url';
+import { PreparedError } from '../helpers/prepared-error';
 import { getGitHubRootSecret, makeWebhookSecret } from './webhook';
 
 const logger = logging.getLogger('express');
 const SNAPCRAFT_INFO_WHITELIST = ['name', 'confinement'];
 
-// TODO: same in launchpad.js
 const RESPONSE_NOT_FOUND = {
   status: 'error',
   payload: {
@@ -27,14 +27,6 @@ const RESPONSE_AUTHENTICATION_FAILED = {
   payload: {
     code: 'github-authentication-failed',
     message: 'Authentication with GitHub failed'
-  }
-};
-
-const RESPONSE_OTHER = {
-  status: 'error',
-  payload: {
-    code: 'github-error-other',
-    message: 'Something went wrong when creating a webhook'
   }
 };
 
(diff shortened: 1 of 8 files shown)


bartaz:
@matthewpaulthomas I need some UX input on error message to show here.

Use case is like this:

User has a repo and adds it to BSI. Later this repo is removed from GH (or made unaccessible in some other way, for example made private).

The bug on the backend makes whole list of repos disappear in such case. Which is fixed here.
But if we don't change the UI such repo (not accessible on GH) will show as 'Not configured' (because we can't find snapcraft.yaml):

screen shot 2017-10-10 at 11 50 50

The problem with this is that we also give suggestions to fix the issue by adding snapcraft.yaml (link to GH) which would not work, because repo doesn't exist on GH.

It will probably be more helpful to mark such repository with an error, which I done as below:

screen shot 2017-10-10 at 11 51 32

Is this error message fine, or do we want do say it differently?

I believe there are different ways in which repo may not be available to us:
- repo was removed
- repo was made private
- repo exists but is empty (doesn't even have any branch to create snapcraft.yaml in) - possibly as in #972
- maybe some others I'm not aware of

src/server/handlers/launchpad.js
```diff
@@ -562,15 +571,46 @@ export const getGitBranch = async (snap, token) => {
   }
 };
 
+const getSnapRepoData = async(snap, token) => {
+  let gitBranch, snapcraftData;
+  try {
+    // first try to check defaut branch
+    gitBranch = await getGitBranch(snap, token);
+    // then get snapcraft.yaml data - there is no point trying to fetch this if fetching branch fails
+    snapcraftData = await getSnapcraftData(snap.git_repository_url, token);
```
cjwatson: This means that we're doing the round-trips to GitHub consecutively rather than concurrently, which is likely to make it slower.  As far as I can see the point of this is mainly to ensure that errors from `getSnapcraftData` are ignored if `getGitBranch` has already failed, not so much the edge-case optimisation of performing fewer requests.  So how about something like this:

    let gitBranchError = null, snapcraftDataError = null;
    const [gitBranch, snapcraftData] = await Promise.all([
      getGitBranch(snap, req.session.token)
        .catch((error) => { gitBranchError = error; }),
      getSnapcraftData(snap.git_repository_url, req.session.token)
        .catch((error) => { snapcraftDataError = error; })
    ]);
    if (gitBranchError !== null) {
      ...
    } else if (snapcraftDataError !== null) {
      ...
    }

That's untested and maybe you can come up with a better formulation, but hopefully you get the idea?

src/server/handlers/github.js
```diff
@@ -13,6 +13,7 @@ import { getGitHubRootSecret, makeWebhookSecret } from './webhook';
 const logger = logging.getLogger('express');
 const SNAPCRAFT_INFO_WHITELIST = ['name', 'confinement'];
 
+// TODO: same in launchpad.js
```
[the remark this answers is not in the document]
bartaz: Yes, I'm thinking about moving all GH related functions (like `checkGitHubStatus`) to `github.js` from `launchpad.js` as this is getting confusing.

That would probably avoid need for exporting GH related response objects.

src/server/handlers/launchpad.js
bartaz: Yes, thanks for suggestion. I haven't thought about catching errors inside of `Promise.all`.

matthewpaulthomas:
> @matthewpaulthomas I need some UX input on error message to show here.

[Spec updated](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.ugspgc2n005y):
> **“Repo missing”**
>
> (#968, #972)
>
> If the repo is not found, the cell should read “Repo missing”, and clicking it should expand the row to reveal a panel with the text:
>> Check whether the repo has been made private or deleted from GitHub.

PR #979, opened 2017-10-23
Adjust for current "missing namespace" response

### LINKED ISSUE #792: "Forbidden" error message without any explanation while registering a snap name
Hello,

While trying to QA a few things, I tried to setup a build for a repo I cloned, and after I logged in to Ubuntu One, and after choosing a snap name and clicking on "Register name" all I see is a "Forbidden" error that is very confusing, and there does not seem to be a way out.

See image:
bsi-error-2

(also note how my username, which is 39 chars long, clobbers the UI and make the name unreadable. This is now reported as #793.)

### LINKED ISSUE #945: "Forbidden" when trying to register a name
## Summary
When trying to register a name for repo I get 'Forbidden' (with no explanation)

## Process
Enter any name in the box and press 'Register name'

The name should be registered, instead I get 'Forbidden'.

cjwatson:
In software-center-agent r4326, the store started responding with
"Developer profile is missing namespace." rather than "Developer profile
is missing short namespace."  It remains unfortunate that we don't have
a proper error code for this, but for the meantime let's at least adjust
to the current response.

Fixes #792; fixes #945.

QAing this will require creating a new account, so it's probably best to do this with a local build configured to point to staging.

### DIFF
--- a/src/common/actions/auth-store.js
+++ b/src/common/actions/auth-store.js
@@ -248,7 +248,7 @@ async function fetchAccountInfo(root, discharge) {
       if (payload.message.indexOf('has not signed agreement') !== -1) {
         data.signedAgreement = false;
       }
-      if (payload.message.indexOf('missing short namespace') !== -1) {
+      if (payload.message.indexOf('missing namespace') !== -1) {
         data.hasShortNamespace = false;
       }
       return data;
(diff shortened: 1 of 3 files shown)


cjwatson:
I also filed https://bugs.launchpad.net/software-center-agent/+bug/1726492 for the API problem here.

PR #980, opened 2017-10-24
Implementing build annotations.

cprov:
Recording `reason` (text) why builds are triggered (`/launchpad/snaps/request-builds`) and returning corresponding 'build_annotations' map when builds are retrieved (`/launchpad/builds`)

### DIFF
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -672,11 +672,24 @@ export const getSnapBuilds = async (req, res) => {
     const snap = await getLaunchpad().get(snapUrl);
     const builds = await internalGetSnapBuilds(snap, req.query.start, req.query.size);
 
+    // Let's collect corresponding build annotations.
+    let build_annotations = {};
+    const build_ids = builds.entries.map((b) => { return b.self_link.split('/').pop(); });
+    const results = await db.model('BuildAnnotation')
+      .where('build_id', 'IN', build_ids)
+      .fetchAll();
+    for (const m of results.models) {
+      build_annotations[m.get('build_id')] = {
+        reason: m.get('reason')
+      };
+    }
+
     return res.status(200).send({
       status: 'success',
       payload: {
         code: 'snap-builds-found',
-        builds: builds.entries
+        builds: builds.entries,
+        build_annotations: build_annotations
       }
     });
   } catch (error) {
@@ -684,7 +697,7 @@ export const getSnapBuilds = async (req, res) => {
   }
 };
 
-export const internalRequestSnapBuilds = async (snap, owner, name) => {
+export const internalRequestSnapBuilds = async (snap, owner, name, reason) => {
   // Request builds, then make sure that auto_build is enabled so that
   // future push events will cause the webhook to dispatch builds.  Doing
   // things in this order ensures that Launchpad's internal
@@ -697,6 +710,7 @@ export const internalRequestSnapBuilds = async (snap, owner, name) => {
   if (!snap.auto_build) {
     await lpClient.patch(snap.self_link, { auto_build: true });
   }
+
   // We can't do this properly transactionally (i.e. don't request builds if
   // the DB connection fails), because we don't know how many builds we're
   // going to request up-front.  If we find a way to fix that then we should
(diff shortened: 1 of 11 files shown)


bartaz:
Quite recently we had issues with relying on human readable messages (#979), so I would feel safer if we didn't repeat the same with build annotations.

It's quite likely that UI will display these annotations differently from the messages stored in the database, so they will need to be mapped into other text, icon or whatever UI chooses to use.
It would be better if database would keep identifiers less prone to change (codes like "BUILD_REASON_MANUAL" for example).

Or at least it would be good to define all possible reasons (used currently in the app) in one place and only define these human readable text representation there. In all other places we should reference these messages by variable name to avoid changing them in many places.

cprov:
@bartaz thanks for the direction, I've updated it to store identifiers.

This implementation is not quite what I envisioned for "Buildsets" (group of snap revisions with a common trigger and possibly a changelog message), but it would drastically affect the UX. Plain annotations is less smart but suits the existing UX better.

Let me know if we need more fixes before landing it.

src/server/handlers/launchpad.js
```diff
@@ -672,19 +672,32 @@ export const getSnapBuilds = async (req, res) => {
     const snap = await getLaunchpad().get(snapUrl);
     const builds = await internalGetSnapBuilds(snap, req.query.start, req.query.size);
 
+    // Let's collect corresponding build annotations.
+    let build_annotations = {};
```
bartaz: Ideally it would be great for collecting annotations to be part of `internalGetSnapBuilds`, so that `internalGetSnapBuilds` already returns annotated builds. In current implementation annotations are returned separately to the UI, which means that it's the front-end that will need to join annotations with build object.
Preferably this join would be done in `internapGetSnapBuilds`, so that any part of the app that uses `internalSnapBuilds` would get annotated builds.

If that seems unnecessary, I'd suggest at least extracting getting annotations to some `internalGetBuildAnnotations` method, so the logic is not hidden only in `getSnapBuilds` handler.

cprov: I've isolated `internalGetBuildAnnotations` and `getBuildId(<build>)` common helper so the client work is clearer. I thought it's better to keep build records pristine, as they come from LP, for the time being.

src/server/handlers/launchpad.js
```diff
@@ -726,8 +760,9 @@ export const requestSnapBuilds = async (req, res) => {
     const { owner, name } = await checkAdminPermissions(
       req.session, req.body.repository_url
     );
+    const reason = req.body.reason || 'triggered-manually';
```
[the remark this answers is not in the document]
cprov: Right, it's definitely cleaner now. Thanks for the suggestion.

src/server/handlers/launchpad.js
```diff
@@ -726,8 +769,9 @@ export const requestSnapBuilds = async (req, res) => {
     const { owner, name } = await checkAdminPermissions(
       req.session, req.body.repository_url
     );
+    const reason = req.body.reason || BUILD_TRIGGERED_MANUALLY;
     const snap = await internalFindSnap(req.body.repository_url);
-    const builds = await internalRequestSnapBuilds(snap, owner, name);
+    const builds = await internalRequestSnapBuilds(snap, owner, name, reason);
     return res.status(201).send({
       status: 'success',
       payload: {
```
bartaz: Currently `requestSnapBuilds` payload only contains builds.
It would be good if it would return `build_annotations` as well, similar to `getSnapBuilds`, so once new builds are requested we can immediately show them in UI with annotations (without refetching them from server).

test/routes/src/server/routes/t_webhook.js
```diff
@@ -307,20 +307,16 @@ describe('The WebHook API endpoint', () => {
...
-                    self_link: `${lp_api_base}${lp_snap_path}/+build/1`
-                  },
-                  {
-                    resource_type_link: `${lp_api_base}/#snap_build`,
-                    self_link: `${lp_api_base}${lp_snap_path}/+build/2`
-                  }
-                ]
-              });
+              .reply(200, [
+                {
```
bartaz: Why is this change added?
Doesn't LP collection response contain `total_size` etc properties?

cprov: Not the response of `requestAutoBuilds`, it simply returns the created builds in the body. It must have been confused with the response of snap builds which is indeed a LP collection with total_size and start.

src/server/handlers/launchpad.js
cprov: Good point, I will include the just created build_annotations.

PR #1007, opened 2017-11-09
Show 'Commit' as build trigger even if commit id is unknown.

### LINKED ISSUE #999: Build triggered by commit show 'Unknown' trigger at first
## Summary

When build is triggered by commit it shows up with build trigger 'Unknown' at first.
Only after some time (when LP returns `revision_id` from API, so we know the commit id).

## Process

Currently only on staging (not released to prod yet):
Go to build.staging.snapcraft.io

1. Have a repo set up
2. Commit to it
3. Go to repo page and look at latest builds

Currently new builds in progress will show 'Unknown' as build trigger.

Ideally expected result would be to show that build was triggered by commit and a link to a commit.
But LP API doesn't return commit ID for some time. Until that happens we don't know the commit ID to show. But at least we could show that build was triggered by commit (without showing its ID).

bartaz:
## Done

Show build triggered by webhook as 'Commit' even if commit id is not known yet (from LP API).

## QA

Is hard to QA locally as it requires webhook to work.

- Commit to a repo added to BSI
- Open repo page, new builds show have 'Commit' as build trigger
- Once the build proceeds commit id should be shown as well

## Issue / Card

Fixes #999 

## Screenshots

screen shot 2017-11-09 at 10 12 56

### DIFF
--- a/src/common/components/build-row/index.js
+++ b/src/common/components/build-row/index.js
@@ -33,7 +33,7 @@ const getBuildTriggerMessage = (repository, reason, commitId) => {
           </span>
         );
       }
-      return 'Unknown';
+      return 'Commit';
     default:
       return 'Unknown';
   }
--- a/src/common/components/build-row/index.js
+++ b/src/common/components/build-row/index.js
@@ -25,7 +25,7 @@ const getBuildTriggerMessage = (repository, reason, commitId) => {
             <a
               target="_blank"
               rel="noopener noreferrer"
-              href={`https://github.com/bartaz/snap-build-test/commit/${commitId}`}
+              href={`${repository.url}/commit/${commitId}`}
             >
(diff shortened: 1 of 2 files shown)


nottrobin:
Code looks good, but I couldn't quite work out how to QA it - I couldn't work out how to trigger a build via commit for my local repos.

I'm off this afternoon and tomorrow. Maybe ping @tbille and walk him through QAing it? It would be good for his education as well.

tbille:
Not from your PR: Update commit link `src/common/components/build-row/index.js` line 28 :smile: 

Otherwise everything is fine QA +1

bartaz:
@tbille Good catch, thanks.

If you are done with review please add 'Review: Code +1' and 'Review: QA +1' and approve it through GH UI, so that GitHub recognises it was approved.

Thanks!

PR #1012, opened 2017-11-14
Add npm version requirement to README

bartaz:
As npm@5 introduces changes that don't seem to work with our outdated dependencies we need to make sure we run older version of npm (like npm@4).

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Snapcraft Build site.
 
 ## Local development
 
-First, make sure all dependencies are installed, you will need node@6 (for example version 6.11.3) and npm@4 (for example version 4.6.1):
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,7 +4,7 @@ Snapcraft Build site.
 
 ## Local development
 
-First, make sure all dependencies are installed, you will need the version 6.11.3 of node:
+First, make sure all dependencies are installed, you will need node@6 (for example version 6.11.3) and npm@4 (for example version 4.6.1):
```
tbille: For me it's ok, but we could but the version that is used by nvm (npm@3) since we use this one in our CI?

PR #1019, opened 2017-12-01
Add a trivial /_status/check endpoint

cjwatson:
This is just enough to keep the ols-http charm layer happy.  I'm not
quite sure why this suddenly started to be a problem, but it's currently
causing deployments to fail to complete.

We don't have a good way to unit-test the full server app assembly, but
I've tested this manually by pointing curl at `/_status/check`, either
from localhost or from an IP address in `TRUSTED_NETWORKS`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "dotenv": "^2.0.0",
     "es6-promise": "^4.0.5",
     "express": "^4.14.0",
+    "express-healthcheck": "^0.1.0",
     "express-prom-bundle": "^2.1.0",
     "express-session": "^1.14.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -50,6 +50,7 @@
     "dotenv": "^2.0.0",
     "es6-promise": "^4.0.5",
     "express": "^4.14.0",
+    "express-healthcheck": "^0.1.0",
```
bartaz: Shouldn't it show up in npm-shrinkwrap.json as well?

cjwatson: Oops, yes.  Done.

PR #1021, opened 2017-12-06
Adds location to GA pageview event.

### LINKED ISSUE #997: Add GA events to build

bartaz:
## Done

Adds location to pageview GA event to possibly fix /select-repository not appearing in GA in history change.

## QA

To properly QA it we need to get it to production and run agains our GA suite.

For testing locally you would need to set up some testing GA account ids for GOOGLE_ANALYTICS_ID or  GOOGLE_OPTIMIZE_ID env variables so that GA scripts are properly loaded. (possibly using values from production.env is fine, they will not be tracked in GA from localhost anyway).
- Open network panel in browser debug tools
- Go to "Add repos"
- Look into network panel, find request to GA, it should contain `dl` parameter set correctly to full /select-repository location

screen shot 2017-12-06 at 14 31 08

## Issue / Card

Fixes #997

### DIFF
--- a/src/common/root.js
+++ b/src/common/root.js
@@ -18,6 +18,7 @@ export default class Root extends Component {
     */
     history.listen((location) => {
       if (window.ga) {
+        window.ga('set', 'location', window.location.href);
         window.ga('set', 'page', location.pathname);
         window.ga('send', 'pageview');
       }


Lukewh:
I'd just get this merged @bartaz... Not sure it's worth the time to fiddle to QA it. There's no reason it wouldn't work.

PR #1027, opened 2017-12-08
Adds cookie notification based on Canonical cookie-policy module

### LINKED ISSUE #786: No cookie warning
I noticed when following footer link from build.snapcraft.io to www.ubuntu.com/legal that BSI does not have an EU cookie notification.

bartaz:
## Done

Adds a cookie policy notification to the page.
Based on `cookie-policy` (https://github.com/canonical-webteam/cookie-policy) module, but reimplemented to fit React and CSS Modules ecosystem.

## QA

- Cookie policy notification should show up on the bottom of the screen.
- Close it by clicking on X
- Check dev tools to see if cookie `_cookies_accepted` was added.
- Reload the page (notification should not appear if you still have the cookie)
- Delete `_cookies_accepted` cookie
- Reload the page (notification should appear again)

## Issue / Card

Fixes #786

## Screenshots

screen shot 2017-12-08 at 13 37 44

### DIFF
--- a/src/common/components/cookie-notification/index.js
+++ b/src/common/components/cookie-notification/index.js
@@ -1,0 +1,99 @@
+import React, { Component } from 'react';
+
+// importing styles directly from cookie-policy module
+import styles from '../../../../node_modules/cookie-policy/build/css/cookie-policy.css';
+
+// Quick React-friendly reimplementation of cookie-policy script:
+// https://github.com/canonical-webteam/cookie-policy/blob/master/src/js/cookie-policy.js
+// Note: it doesn't support any settings (like custom content or delay)
+export default class CookieNotification extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isOpen: false
+    };
+  }
+
+  componentDidMount() {
+    this.setState({
+      isOpen: (this.getCookie('_cookies_accepted') !== 'true')
+    });
+  }
+
+  closeCookie() {
+    if (this.state.isOpen) {
+      this.setState({ isOpen: false });
+      this.setCookie('_cookies_accepted', 'true', 3000);
+    }
+  }
+
+  setCookie(name, value, exdays) {
+    var d = new Date();
+    d.setTime(d.getTime() + exdays * 24 * 60 * 60 * 1000);
+    var expires = 'expires=' + d.toUTCString();
+    document.cookie = name + '=' + value + '; ' + expires;
+  }
+
+  getCookie(name) {
+    name = name + '=';
+    var ca = document.cookie.split(';');
+    for (var i = 0; i < ca.length; i++) {
+      var c = ca[i];
+      while (c.charAt(0) == ' ') {
+        c = c.substring(1);
+      }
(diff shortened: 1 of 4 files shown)


src/common/components/cookie-notification/index.js
```diff
@@ -0,0 +1,99 @@
+import React, { Component } from 'react';
+
+// importing styles directly from cookie-policy module
+import styles from '../../../../node_modules/cookie-policy/build/css/cookie-policy.css';
+
+// Quick React-friendly reimplementation of cookie-policy script:
+// https://github.com/canonical-webteam/cookie-policy/blob/master/src/js/cookie-policy.js
+// Note: it doesn't support any settings (like custom content or delay)
+export default class CookieNotification extends Component {
```
anthonydillon: Interesting, is there a benefit of rewriting this into a React component over using the module JS?

src/common/components/cookie-notification/index.js
```diff
@@ -0,0 +1,99 @@
+import React, { Component } from 'react';
+
+// importing styles directly from cookie-policy module
+import styles from '../../../../node_modules/cookie-policy/build/css/cookie-policy.css';
```
anthonydillon: Doesn't the sass command have an `includePath` to node_modules?

bartaz: We use CSS Modules in BSI, not sass directly. So at this point we can't really use cookie-policy sass here.
Fortunately it's convenient enough to import built CSS as CSS Module which works just fine.

src/common/components/cookie-notification/index.js
bartaz: cookie-policy.js doesn't work well as JS module. It's implemented in a way that only supports using it as a global script that adds itself to global `ubuntu` namespace.

I tried to import it as a module in hope that I could reuse the implementation, but it gets encapsulated in a module, so it doesn't add any global variables. And because it doesn't export anything, it's not visible outside of local module scope.
Instead of working around that and trying to make cookie-policy.js fit into quite uncommon (at least for current webteam projects) node build system it was easier and quicker to duplicate the implementation in React component.

I guess it would be possible to update cookie-policy script to allow importing it as JS module in node environment, but I didn't feel it would fit into scope of this fix.

bartaz: @anthonydillon I had a second look and in theory it wouldn't be too hard to make cookie-policy work as a module. Just a matter of adding:

```js
if (typeof module === "object" && module.exports) {
  module.exports = ubuntu.cookiePolicy;
}
```

But, it still doesn't work out of the box in BSI because of how CSS Modules obfuscate class names. cookie-policy script creates HTML using Vanilla class names, but if we include cookie-policy.css as a CSS Modules these class names are not available anymore.
So even while we could quite easily make script itself work, it would still require loading CSS not as a CSS Module which would be something we didn't need to do in BSI before.

Probably in the future once build functionality is moved more and more into unified snapcraft.io maybe we will need to get rid of CSS module loading to have better and easier reusability of Vanilla CSS in BSI.

PR #1037, opened 2017-12-20
Adds a column to db to store public user email.

### LINKED ISSUE #731: Add email addresses to requested GitHub permissions
We'd like to send out an email announcing organisations support (once landed) to all GitHub users who have ever logged in. Can you please generate a list of these addresses in a secure internal location?

If we haven't requested that permission from GitHub, can we please start?

bartaz:
## Done

Added email column to github_user database table.
Saving public user email from GitHub when user signs in.

## QA

- Make sure you have public email in your GH profile
- Sign in to BSI
- Your public email should be stored in db

## Issue / Card

Fixes #731

### DIFF
--- a/migrations/20171219153812_user-email.js
+++ b/migrations/20171219153812_user-email.js
@@ -1,0 +1,15 @@
+exports.up = function(knex, Promise) {
+  return Promise.all([
+    knex.schema.table('github_user', function(table) {
+      table.text('email');
+    })
+  ]);
+};
+
+exports.down = function(knex, Promise) {
+  return Promise.all([
+    knex.schema.table('github_user', function(table) {
+      table.dropColumn('email');
+    })
+  ]);
+};
--- a/src/server/db/models/github-user.js
+++ b/src/server/db/models/github-user.js
@@ -4,6 +4,7 @@ export default function register(db) {
    *   github_id: ID from GitHub
    *   name: display name from GitHub
    *   login: login name from GitHub
+   *   email: public email from GitHub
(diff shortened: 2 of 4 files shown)


nottrobin:
What's the purpose of this? Is it to store a list of email addresses to send out an email to them?

Bear in mind that we're [looking to make build.snapcraft.io stateless](https://bugs.launchpad.net/snapstore/+bug/1738241), and at that point we won't be able to store persistent information locally to the build.snapcraft.io app.

Is it possible to send this email somewhere else to be stored? Marketo or something?

tbille:
email well registered in the DB

bartaz:
@nottrobin Yes, #731 has the whole discussion about getting access to email (for marketing) and allowing to unsubscribe.

The point is to collect emails and import them to Marketo (as far as I understand).

bartaz:
@nottrobin While adding email is an additional data we store, I'm not sure if it strongly breaks  BSI being stateless. My understanding is that we need a 'log' of emails of users who logged into BSI. We store logged in users for metrics in db, email is just another field that is stored there.
If we loose this db we only loose the log of who logged in and what actions they did, but we will not loose any significant data.

If Marketo allows that we could send these emails directly there. But this is a completely different task that would probably allow set up on Marketo site and work in BSI.

nottrobin:
@bartaz OK cool. I guess it's okay to merge this for now then.

Lukewh:
Merging so it doesn't go stale over the holiday break....

PR #1050, opened 2018-01-19
Updates header to with new snapcraft navigation

### LINKED ISSUE #1047: Update header to have same navigation and style as snapcraft.io
## Summary

Now that snapcraft.io header navigation was updated we need to update BSI navigation as well.

screen shot 2018-01-18 at 16 50 08

bartaz:
## Done

Updates header and navigation to be consistent with snapcraft.io.

Fixes #1047 

## QA

- Navigation should have Vanilla styles and items same as on snapcraft.io
- Sign in
- Navigation should show that you are signed in

## Screenshots

Not signed in

screen shot 2018-01-19 at 16 36 43

Signed in

screen shot 2018-01-19 at 16 36 55


src/common/style/vanilla/navigation.scss
```diff
@@ -0,0 +1,60 @@
+@import 'vanilla-framework/scss/_utilities_clearfix.scss';
+@import 'vanilla-framework/scss/_base_typography.scss';
+
+// Vanilla 1.5.0 default navigation is dark, so we are using --light navigation,
```
bartaz: BSI is on Vanilla 1.5.0.

I tried to update to latest 1.6.4 in this PR, but it introduced some unrelated issues on home page (bugs reported to vanilla), so I decided to stay on 1.5.0 to reduce the scope of this changes.

bartaz:
@Lukewh @matthewpaulthomas @carlaberkers 

Now that we want to have same navigation on BSI and snapcraft.io there are couple of questions/issues:

- where should logo link to? - to https://snapcraft.io in both cases?
- should Store be external link on BSI? If so it looks strange to have external link first in the navigation. Also, if both sites have same navigation maybe we should treat them as same 'site', so both Store and Build should not be external links?
- sign in is currently only available on Build and we can't easily share it with snapcraft.io, because it's GitHub auth specifically for Build, so this part of navigation is going to be inconsistent between the two sites

src/common/components/header/index.js
```diff
@@ -11,34 +12,55 @@ export default class Header extends Component {
...
+                <img className={ style['p-navigation__image'] } src={ wordmark } alt="Snapcraft.io" width={ 148 } />
+              </a>
+            </div>
+            <a href="#navigation" className={ style['p-navigation__toggle--open'] } title="menu">Menu</a>
+            <a href="#navigation-closed" className={ style['p-navigation__toggle--close'] } title="close menu">Close menu</a>
+          </div>
+          <nav className={ style['p-navigation__nav'] } role="menubar">
+            <ul className={ style['p-navigation__links']} role="menu">
+              <li className={ style['p-navigation__link'] } role="menuitem">
+                <a href="https://snapcraft.io/store/">Store</a>
```
tbille: You need to add `className={ style['p-link--external'] }`

tbille: oh I just read your comment
> should Store be external link on BSI? If so it looks strange to have external link first in the navigation. Also, if both sites have same navigation maybe we should treat them as same 'site', so both Store and Build should not be external links?

bartaz: Yes, I know... I'm asking about it in comment above, because it looks quite weird to have Store as an external link.

screen shot 2018-01-19 at 17 02 26

So maybe this would need some other changes, like changing the order of links or treating Store and Build and not external to each other... UX input is needed here.

Lukewh:
I agree with @matthewpaulthomas

src/common/components/header/header.css
```diff
@@ -1,71 +0,0 @@
...
-}
-
-.link:hover {
-  background-color: $border-color;
-  text-decoration: none;
-}
-
-.link:last-child {
-  border-right: 1px solid $border-color;
-}
```
Lukewh: Beautiful

PR #1055, opened 2018-02-01
Removes non-functional “Sign out” item

### LINKED ISSUE #305: [header] “Sign out” doesn’t do anything
0\. Sign in to build.snapcraft.io if you aren’t already.
1\. Choose “Sign out”.
2\. Choose “Sign in”.

What happens: You see your list of repos, and can add more, remove them (once that’s implemented), register snap names, etc.

What should happen: You’re required to authenticate to GitHub again before you can see your repos, and to Ubuntu One before you can register a snap name.

Apparently Travis does something similar. I don’t think that makes it okay — especially when we’re demoing it on a public-facing machine.

---

Copied from Trello card: https://trello.com/c/pmqxkHhH/254-header-sign-out-doesn-t-do-anything
Labels: Bug

matthewpaulthomas:
## Done

Removes the “Sign out” item from the header. A temporary solution, until proper sign-in and sign-out is implemented when build.snapcraft.io is integrated with the rest of snapcraft.io.

## QA

- sign in
- verify that in the header, “Hi, {your name}” is no longer followed by “Sign out”

## Issue / Card

Fixes #305.

### DIFF
--- a/src/common/components/header/index.js
+++ b/src/common/components/header/index.js
@@ -47,9 +47,6 @@ export default class Header extends Component {
                   <li className={ style['p-navigation__item'] } role="menuitem">
                     <a>Hi, {user.name || user.login}</a>
                   </li>
-                  <li className={ style['p-navigation__link'] } role="menuitem">
-                    <a href="/auth/logout">Sign out</a>
-                  </li>
                 </ul>
               :
                 <ul className={ style['p-navigation__links--right']} role="menu">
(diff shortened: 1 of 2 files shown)


bartaz:
It's not true that 'Sign out' currently doesn't to anything. It clears current user session in build.snapcraft.io which means for example that they can sign in again using different GH or different Store accounts (assuming of course that they logged out from them separately).

Removing this link will lead to confusing UI where we will have 'Sign in' but will not allow users to 'Sign out'. Also, the only way for user to close their session with BSI would be to wait until their session times out.

I think we should keep current not-perfect solution until proper integration of build with snapcraft.io and proper sign in is implemented.

PR #1092, opened 2018-05-11
Reduce polling frequency for repos

Lukewh:
## Done

Polling repo's happens every 15 seconds, this reduces that frequency in an attempt to alleviate some issues caused by long repo lists.

## QA

- Make sure the app doesn't blow up after an hour or so.


bartaz:
We should later follow up with better handling of unnecessary polling, but this should hopefully reduce number of requests.

PR #1097, opened 2018-05-17
Sentry preparations

bartaz:
## Done

- updated raven module, so server side sentry API can use the same public sentry_dsn
- added raven to HTML, so we can capture client side errors


bartaz:
Not to be merged before we clarify where to send the errors.

Issue #1090, opened 2018-05-10
Authentication failed. API rate limit exceeded for user ID 1841272.


popey:
## Summary

Trying to use build.snapcraft.io and now I can't even login

## Process

visit build.snapcraft.io
login

## Current and expected result

Expected to get logged in and see my repos
Got an error "Authentication failed. API rate limit exceeded for user ID 1841272."

## Screenshot

![screenshot from 2018-05-10 14-59-34](https://user-images.githubusercontent.com/1841272/39873497-03c1ca56-5463-11e8-9075-879b711b5f14.png)

bartaz:
GH API rate limit is 5000 requests per hour (per authenticated user) and we never had any issue with it being not enough.

But we may need to verify if we don't poll API too often (for example if user has many repos). Possibly some extensive reloading could have reached this limit.

As far as I remember GH clears rate limits every hour, so it should be back operational when it's cleared on GH side.

bartaz:
I just had couple BSI tabs open with snapcrafters repos for less then an hour and it killed my GH rate as well.

I don't think we are calling GH API when polling for build statuses (which is the most frequent call), but it seems we do call it a lot when there are many repos.

Lukewh:
Immediate fix:

- When a tab is not focused, stop polling.
- Show an overlay explaining the tab is paused, including a 'continue' CTA (just in case there's a delay when...)
- Resume polling on tab focus (and remove the overlay)

Longer term investigations and improvements:

- Audit calls to Github and optimize where possible
- Poll less frequently until a build is detected, increase polling then back off again when the build is complete.

bartaz:
### Quick GH calls audit

On 'My repos' page we poll (currently every 30 seconds) for list of user snaps/repos and build status of these repos. Build status calls (even that there are lots of them) only call LP, so they don't add up to GH API rate limit.

Request for list of user repos consists of couple of parts:
- call for list of user GH orgs
- call for lists of user snaps (for GH user and all the orgs)
- call for branch and snapcraft.yaml data (for every repo on the list)

So in theory every time we request list of repos (including their config status) we would call GH multiple times (for each repo).
Fortunately these calls are memcached with timeout set for an hour (3600 sec). This means that for **most** of repos we only call GH for their snapcraft.yaml data once an hour.

[...]

PR #1099, opened 2018-05-22
Don't poll when page is inactive

bartaz:
## Done

Disables polling for snaps and builds when page is inactive (tab is invisible).

## QA

- Go to My repos page
- Open "Network" panel of developer tools
- when page is visible every 30 seconds there should be requests for snaps and builds
- Switch to another tab, wait for more then 30 seconds (more then minute ideally)
- Switch back to 'My repos'
- Immediately when page is made visible snaps and builds should be fetched
- If you wait with page visible every 30 seconds there should be requests
- Switch to another tab and go back to 'My repos' quickly (in less then 30 seconds)
- There should be no requests until full 30 seconds pass

Similar on repo builds page:
- go to builds page for single repo
- when page is visible there should be a request for build statuses every 15 seconds
- if page is invisible there should be no requests in the background
- when page is visible again there should immediately be request to update builds (but only if more then 15 seconds has passed).

## Issue / Card

Part of fixes around #1090

### DIFF
--- a/src/common/components/repositories-home/index.js
+++ b/src/common/components/repositories-home/index.js
@@ -24,17 +24,42 @@ class RepositoriesHome extends Component {
     });
   }
 
-  componentWillMount() {
+  onVisibilityChange() {
+    if (document.visibilityState === 'visible') {
+      const { authenticated } = this.props.auth;
+      const { updateSnaps } = this.props;
+      const owner = this.props.user.login;
+      const now = Date.now();
+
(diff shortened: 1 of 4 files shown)


bartaz:
@Lukewh With `visibilitychange` event being triggered immediately and using it to update the snaps/builds I don't think we need any overlay explaining that we don't fetch in the background. This seems like unnecessary information to the user (that would be barely noticable).

PR #1102, opened 2018-05-23
Detect expired session

### LINKED ISSUE #1003: Can't leave build.snapcraft.io page up
If you leave leave a repo page up long enough, eventually you'll see an unauthorised error:
screen shot 2017-11-08 at 15 29 13

It gets worse if you navigate back to your dashboard, where it will claim you have no repos:
screen shot 2017-11-08 at 15 29 51

Because:
https://build.snapcraft.io/api/launchpad/snaps/list?owner=evandandrea
```{"status":"error","body":{"code":"token-not-valid","message":"CSRF token not valid"}}```

### PR DESCRIPTION
**Pull request by bartaz**
## Done

- added detecting `401 Unauthorized` to API request handlers
- when expired session is detected (API returns 401) overlay is shown asking to reload and sign in again
- fixed unhandled sign out link in header (to properly clear session on sign out)

## QA

Not to wait 2h for session to expire `src/server/helpers/session.js:11` can be edited to change session cookie maxAge from 2h to something shorter like a minute or 10 seconds.

- Sign in with Github
- Possibly sign in into the Store on one of repos
- As long as session is not expired everything should work as expected, no overlay should be shown unnecessarily
- After the session has expired try to make an action that requires user being sign in (like registering the name or adding new repo)
- Overlay should be shown with message about expired session
- Clicking 'Reload the page' in the notification should reload the page (possibly back to home page if user was on one of pages requiring auth)

## Issue / Card

Fixes #1003 

## Screenshots

screen shot 2018-05-23 at 12 53 29

### DIFF
--- a/src/common/components/session-overlay/index.js
+++ b/src/common/components/session-overlay/index.js
@@ -1,0 +1,59 @@
+import React, { Component, PropTypes } from 'react';
+import { connect } from 'react-redux';
+
+import { signOut } from '../../actions/auth-store';
+
+import Notification from '../vanilla-modules/notification';
+
+import style from './sessionOverlay.css';
+import containerStyle from '../../containers/container.css';
+
+export class SessionOverlay extends Component {
+
+  onReloadClick(event) {
+    event.preventDefault();
+    this.props.clearSession();
+    window.location.reload();
+  }
+
+  render() {
+    return !this.props.isExpired ? null : (
+      <div className={style.sessionOverlay}>
+        <div className={containerStyle.wrapper}>
+          <Notification appearance="caution">
+            Your session has expired.
+            {' '}
+            <a href="/" onClick={this.onReloadClick.bind(this)}>
+              Reload the page
+            </a>
+            {' '}
+            and sign in again.
+          </Notification>
+        </div>
+      </div>
+    );
+  }
+}
+
+SessionOverlay.propTypes = {
+  isExpired: PropTypes.bool,
+  clearSession: PropTypes.func
+};
+
+function mapStateToProps(state) {
+  const {
+    authError
+  } = state;
+
+  return {
+    isExpired: !!authError.expired
+  };
+}
+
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/common/components/session-overlay/index.js [hunk 1]
```diff
@@ -0,0 +1,59 @@
...
+
+import style from './sessionOverlay.css';
+import containerStyle from '../../containers/container.css';
+
+export class SessionOverlay extends Component {
+
+  onReloadClick(event) {
+    event.preventDefault();
+    this.props.clearSession();
+    window.location.reload();
```
Lukewh: Don't we want to redirect to `/auth/authenticate`? It's a bit annoying that the session expires then you have to click through from the homepage again.

bartaz: @Lukewh We could, but (because of current sign in implementation) this would mean that users would always end up on 'My repos' page. Currently going through GH sign in always redirects to 'My repos' regardless where the sign in originated.

If we just reload (not force user to sign in automatically) we (in theory) stay on the same page. What really happens depends on the page.
If user is on any page that is available without sign in (like repo page or build page) they will stay on the same page (just not being signed in).
If user is on page only available to authenticated users (like 'My repos' or 'Add repos') they will get redirected to home page (because they are not signed in).

So the question here is more from UX perspective - once the session expires do we want to just refresh the view to properly sign out the user, or do we want to automatically sign them back in (once they click on 'Refresh') - which will be almost invisible process assuming they are still signed in to their GH account.

Lukewh: I think we should aim to have something smarter happen (like use the current path as the redirect after sign in?) but for now, let's get this merged this is already way better then what we have!

bartaz: Sure. Ideally sign in should lead back where it was originated (unless it's a home page that should lead to 'My snaps' when signed in).

Issue #1006, opened 2017-11-08
"No repos" frequently shown - page slow to load


popey:
Visiting https://build.snapcraft.io/user/popey frequently gives me the text "No repos" in the middle. Which is a bit alarming. Then some time passes and it loads. I realise I probably have more snaps (via snapcrafters) than most, but it's a bit disconcerting to keep being told there's "no repos" when I know there are. If I wait a while it eventually loads.

evandandrea:
Yes, I’ve seen this as well. We should have a “checking, just a minute” type message.

matthewpaulthomas:
[From the spec](https://docs.google.com/document/d/1koTwI3cRr2S2vj5zHKKbR7aWxLRfShDVq2gHQ29cDyU/edit#heading=h.5c3e9xiupuq5):
> The table headers should appear immediately. If the list of repos has taken more than two seconds so far to load, a spinner should appear centered in the table.

However, I followed this with text that was imprecise: “Whenever there are no repos in the table …  the table should contain only the centered text ‘No repos’.” I’ve just changed this to “When you have no repos configured for building…”

Lukewh:
As a first step we will:

Improve messaging - indicate when something is happening/ has happened, be it loading still, timing out or any other state.

A longer term solution:

- Optimise the requests made to Github (currently we wait for all the requests to return before showing anything in the UI)
- Improve UX of loading a lot of repos (progressive loading or something else)

bartaz:
Now that I was added to `snapcrafters` and have ~70 repos on my list I almost always have this problem when first logging into BSI after some time.

My process is:
- Sign in
- 'My repos' shows spinner for quite long time
- Ends up showing No repos
- In console I can see that there was a timeout on loading list of repositories `Failed to load resource: the server responded with a status of 504 (Gateway Time-out)`

When I refresh the page it usually takes some time, but eventually loads properly.

The request that times out (listing of repos) consists of:
- calling LP for list of repos of given user (and their orgs)
- calling GH for each repo to fetch snapcraft.yaml data

When first logging in snapcraft.yaml data for repos in likely not in memcached anymore therefore BSI needs to call GH for all (70) repos. My suspision is that it takes too much time and times out. Next time (after refresh) at least some of repos are already memcached, so response is quicker and doesn't time out.

First of all - we need to properly show errors when loading repos (not to end up with 'No repos' and no explanation).
Then we need to figure out the way how to improve performance of listing of the repos.
[...]

PR #1108, opened 2018-05-25
Batch "ensure webhook" calls to LP (for better performance)

bartaz:
## Done

Part of #1006 

When debugging the timing of `findSnaps` response it turned out that (when nothing is memcached) one of the biggest bottlenecks is checking with LP that every repo/snap has a webhook.

When there is over 70 repos the whole response takes ~30 seconds, ~25 of which is spent asking LP to for webhooks for every snap one after another.

This PR changes for loop into `Promise.all` to check webhooks for all snaps concurrently.

## QA

- Sign in, wait for repos to load
- Everything should work as expected, if you have many repos they should probably load quicker then before

## Issue / Card

Part of fixes around #1006

### DIFF
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -401,7 +401,12 @@ const internalFindSnaps = async (owner, token) => {
   ));
   snaps = snaps.concat(result.entries);
   // ensure LP webhook is set up for all the snaps
-  await Promise.all(snaps.map(snap => ensureWebhook(snap)));
+  // but batch it by 10 not to overload LP servers
+  const BATCH_SIZE = 10;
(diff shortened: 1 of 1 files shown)


bartaz:
@cjwatson Do you think there can be any downside of calling LP API concurrently for dozens of snaps at once to ensure webhook?

cjwatson:
@bartaz `snaps` can be arbitrarily large here, so this could end up hammering Launchpad by tying up lots of its appservers in parallel, which isn't great.  Could you at least batch it into chunks of say 10 or so and check just each chunk in parallel?

We could also consider a bulk API endpoint in Launchpad that takes a collection of snaps and returns those that don't have a given webhook URL, although that feels very ad-hoc.

PR #1109, opened 2018-05-25
Add error notification to My repos page

bartaz:
## Done

Adds error notification to 'My repos' page when there is any error fetching list of snaps.

Currently if snaps listing request ends up with an error we show 'No repos' without any error message.
This at least gives user an idea that something is wrong.

## QA

To mock some server side error you can throw error inside `findSnaps` in `handlers/launchpad.js:556`. 

- Sign in
- Wait for snaps listing to return an error
- Error should show up above snaps list
- Error message will depend on whatever error was returned (and if API failing gave any error message)

## Issue / Card

Helps with #1006

## Screenshots

screen shot 2018-05-25 at 16 07 31

### DIFF
--- a/src/common/components/first-time-heading/index.js
+++ b/src/common/components/first-time-heading/index.js
@@ -73,7 +73,7 @@ class FirstTimeHeading extends Component {
   render() {
     const { message, progress } = this.getCurrentState();
 
-    return (
+    return ( (message || progress) &&
       <div className={styles.firstTimeHeading}>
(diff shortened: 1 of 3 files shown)


Lukewh:
The error shows fine, however there's a really large space because of the padding on the empty `FirstTimeHeading`.

screenshot-2018-5-25 build snapcraft io - home

Lukewh:
Thanks @bartaz looks good thanks

PR #1111, opened 2018-05-30
Only store one token per session

### LINKED ISSUE #1084: Error:Unauthorized
## Summary

Trying to trigger a build manually of a snap and I keep getting "Error: Unauthorized". I have tried signing out and back in, but it keeps happening. I need to trigger these builds because there are security incidents in some parts.

## Process

visit build, login, click on one of my snaps.

image

bartaz:
## Done

Attempting generate and store CSRF tokens was causing some race conditions when multiple tabs were opened at once and multiple tokens were stored into the session overriding each other.

It seems that generating token per request is not necessary. Single token per user session is secure enough (as per [synchronizer tokens](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_%28CSRF%29_Prevention_Cheat_Sheet#Synchronizer_.28CSRF.29_Tokens)). So this PR simplifies token generation to just use one token per authenticated user and fixes the issue with concurrent requests overriding each other tokens.

## QA

- While you are not signed in there should be no token generated
- Sign in with GH - new token should be generated (it is visible as part of state passed to the app in HTML source)
- Open multiple tabs for you repositories
- Each tab should use the same token (see in HTML source)
- You should be able to trigger manual builds without getting `401 Unauthorized` error on console

## Issue / Card

Fixes #1084

### DIFF
--- a/src/server/handlers/universal.js
+++ b/src/server/handlers/universal.js
@@ -42,9 +42,7 @@ export const handleMatch = (req, res, error, redirectLocation, renderProps) => {
       initialState.user = req.session.user;
     }
 
-    const csrfToken = req.session.csrfTokens
-      ? req.session.csrfTokens[req.session.csrfTokens.length - 1]
-      : null;
+    const csrfToken = req.session.csrfToken || null;
 
     if (req.session.error) {
       initialState['authError'] = { message: req.session.error };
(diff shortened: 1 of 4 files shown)


tbille:
Do we still want to keep this when we are not logged in (on the frontend)

```html
<script data-reactid="12">window.__CSRF_TOKEN__ = null</script>
```

PR #1112, opened 2018-05-30
Add configuration for public sentry dsn var

tbille:
Add configurations for public sentry dsn var in the charm


cprov:
@tbille the charm charms are correct and build locally.

PR #1127, opened 2018-07-20
Check the pushed ref in the GitHub webhook handler

cjwatson:
We noticed that the Launchpad build farm was being spammed by identical
builds of certain snaps.  On investigation, this turned out to be
because BSI's GitHub webhook handler doesn't check that the pushed
branch is the one that Launchpad is configured to build for this
repository, so dispatched builds even in the rather common case that the
push was to some other branch.  Checking the branch should mean that we
have more resources for builds that aren't redundant.

## QA

It's probably simplest to QA this once it's on staging, but once it's there that can be done by pushing to a test repository's default branch vs. some other random branch and making sure that only the former case causes builds to be dispatched.

### DIFF
--- a/src/server/handlers/webhook.js
+++ b/src/server/handlers/webhook.js
@@ -8,6 +8,7 @@ import { getMemcached } from '../helpers/memcached';
 import logging from '../logging';
 import { getSnapcraftYamlCacheId } from './github';
 import {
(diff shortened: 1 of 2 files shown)


bartaz:
Tested on staging, works as expected.

PR #1129, opened 2018-08-02
Update navigation and redirect to snapcraft.io on login and logout

Lukewh:
This PR is one of the pieces required to make build.snapcraft.io and snapcraft.io seamless.

The idea is that `https://build.snapcraft.io` will redirect to `https://snapcraft.io/build`, however, after logging in to `https://snapcraft.io` you'll be able to click the 'Build with GitHub' link in the user menu which will redirect to `https://build.snapcraft.io/auth/authenticate`. Either starting the authentication dance, or redirecting to `https://build.snapcraft.io/user/{github_user}` where the navigation will be the same as on `https://snapcraft.io` (see the note below).

When a user logs out of either site they will be redirected to the log out of both, ending up back on the `https://snapcraft.io` homepage.

## Done

Fixes https://github.com/canonical-websites/snapcraft.io/issues/921

- Made the header visually the same as snapcraft.io
- Added drop down when logged in as on snapcraft.io
- Redirected `/auth/logout` to `https://snapcraft.io/logout`

### Note

The user's name in the header will only match if they've set it the same on launchpad and github. There's nothing we can do about that, but I'm sure they know that they are the same person.

## QA

- Checkout the branch
- `npm start -- --env=environment/dev.env`
- Compare the header to https://snapcraft.io
- Login
- Compare the header to https://snapcraft.io
- Logout
- Be on https://snapcraft.io

## TODO before merging

- Redirect https://snapcraft.io/logout to https://build.snapcraft.io/auth/logout and catch the second redirect implemented in this PR so as not to loop https://github.com/canonical-websites/snapcraft.io/issues/970

### DIFF
--- a/.env
+++ b/.env
@@ -3,3 +3,4 @@ EXTRA_PORTS='8082'
 WEBPACK_DEV_URL=http://0.0.0.0:8082
 BASE_URL=http://0.0.0.0:8083
 OPENID_VERIFY_URL=http://0.0.0.0:8083/login/verify
+SNAPCRAFT_URL=https://snapcraft.io
--- a/.gitignore
+++ b/.gitignore
@@ -108,6 +108,7 @@ version-info.txt
 
 # Environment variables
 environments/development.env
+environments/dev.env
 
 # Development and test SQLite databases
 /dev.sqlite3
--- a/.gitignore
+++ b/.gitignore
@@ -108,7 +108,6 @@ version-info.txt
 
 # Environment variables
 environments/development.env
-environments/dev.env
 
 # Development and test SQLite databases
 /dev.sqlite3
--- a/environments/test.env
+++ b/environments/test.env
@@ -3,6 +3,8 @@ ENVIRONMENT=test
 
 BASE_URL=http://localhost:8000
 
+SNAPCRAFT_URL=https://snapcraft.io
+
 UBUNTU_SSO_URL=https://login.ubuntu.com
 OPENID_VERIFY_URL=http://localhost:8000/login/verify
 
--- a/src/common/components/header/index.js
+++ b/src/common/components/header/index.js
@@ -2,6 +2,8 @@ import React, { Component, PropTypes } from 'react';
 
 import { signOut } from '../../actions/auth-store';
 
+import { conf } from '../../helpers/config';
+
 import style from '../../style/vanilla/css/navigation.css';
 import { IconChevron, IconUser } from '../vanilla-modules/icons';
 
@@ -19,6 +21,8 @@ export default class Header extends Component {
   render() {
     const { authenticated, user } = this.props;
 
+    const SNAPCRAFT_URL = conf.get('SNAPCRAFT_URL');
+
     return (
       <header id="navigation" className={ style['p-navigation'] }>
         <div className={ style['p-navigation__banner'] }>
@@ -34,13 +38,13 @@ export default class Header extends Component {
         <nav className={ style['p-navigation__nav'] } role="menubar">
           <ul className={ style['p-navigation__links']} role="menu">
             <li className={ style['p-navigation__link'] } role="menuitem">
-              <a href="https://snapcraft.io/store/">Store</a>
+              <a href={ `${SNAPCRAFT_URL}/store` }>Store</a>
             </li>
             <li className={ style['p-navigation__link'] } role="menuitem">
-              <a href="https://snapcraft.io/blog/">Blog</a>
+              <a href={ `${SNAPCRAFT_URL}/blog` }>Blog</a>
             </li>
             <li className={ style['p-navigation__link'] } role="menuitem">
-              <a className={ authenticated ? '' : style['is-selected'] } href="https://snapcraft.io/build">Build</a>
+              <a className={ authenticated ? '' : style['is-selected'] } href={ `${SNAPCRAFT_URL}/build}` }>Build</a>
             </li>
             <li className={ style['p-navigation__link'] } role="menuitem">
               <a href="https://docs.snapcraft.io">Docs</a>
@@ -60,13 +64,13 @@ export default class Header extends Component {
                   </a>
                   <ul className={ style['p-dropdown__menu']} id="account-menu" aria-hidden={ !this.state.showUserDropdown }>
                     <li className={ style['p-navigation__link']} role="menuitem">
-                      <a href="https://snapcraft.io/account/snaps">My published snaps</a>
+                      <a href={ `${SNAPCRAFT_URL}/account/snaps` } >My published snaps</a>
                     </li>
                     <li className={ style['p-navigation__link']} role="menuitem">
                       <a href={ `/user/${user.login}`} className={ style['is-selected']}>Build with GitHub</a>
                     </li>
                     <li className={ style['p-navigation__link']} role="menuitem">
-                      <a href="https://snapcraft.io/account/details">Account details</a>
+                      <a href={ `${SNAPCRAFT_URL}/account/details` }>Account details</a>
                     </li>
                     <li className={ style['p-navigation__link']} role="menuitem">
                       <a href="/auth/logout" onClick={ this.onLogoutClick.bind(this)}>Sign out</a>
@@ -77,7 +81,7 @@ export default class Header extends Component {
             ) : (
               <ul className={ style['p-navigation__links--right']} role="menu">
                 <li className={ style['p-navigation__link'] } role="menuitem">
-                  <a href="/auth/authenticate">
+                  <a href="http://0.0.0.0:8004/login?next=/login/bsi">
                     <IconUser/>Developer account
                   </a>
                 </li>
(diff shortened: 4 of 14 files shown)


bartaz:
I find this new navigation a little bit confusing. It's trying to pretend that BSI and snapcraft.io use the same account, when they don't.

When I'm signed in to snapcraft.io I use my store account (and it says "Bartosz Szopka"):
screen shot 2018-08-03 at 09 28 41

BSI uses GitHub account, but displays it in exactly the same place with my GH name "Bartek Szopka":
screen shot 2018-08-03 at 09 28 12

It pretends that I'm signed in to both, while I may not.

Even if I am signed in snapcraft.io with my store account, go to BSI with my github account, I still need to sign in to store on BSI if I want to register the name.

src/common/components/header/index.js
```diff
@@ -2,73 +2,99 @@ import React, { Component, PropTypes } from 'react';
 
 import { signOut } from '../../actions/auth-store';
 
-import containerStyles from '../../containers/container.css';
 import style from '../../style/vanilla/css/navigation.css';
+import {IconChevron, IconUser} from "../vanilla-modules/icons";
```
bartaz: There are couple of linting issues in this file (see Travis build log).

src/common/components/header/index.js
```diff
@@ -2,73 +2,99 @@ import React, { Component, PropTypes } from 'react';
...
+                  <a className={ style['p-dropdown__toggle']} aria-controls="account-menu" aria-expanded={ this.state.showUserDropdown }
+                    onClick={ this.onDropdownClick.bind(this)}>
+                    {user.name || user.login}<IconChevron/>
+                  </a>
+                  <ul className={ style['p-dropdown__menu']} id="account-menu" aria-hidden={ !this.state.showUserDropdown }>
+                    <li className={ style['p-navigation__link']} role="menuitem">
+                      <a href="https://snapcraft.io/account/snaps">My published snaps</a>
+                    </li>
+                    <li className={ style['p-navigation__link']} role="menuitem">
+                      <a href="/auth/authenticate">Build with GitHub</a>
```
bartaz: This link is quite confusing in BSI context.

I already am in BSI (possibly via this link), so maybe at least it could be marked as active?

src/server/handlers/github-auth.js
```diff
@@ -146,6 +146,6 @@ export const logout = (req, res, next) => {
       return next(new Error('Failed to log out.'));
     }
     // FIXME redirect to page that initiated the sign in request
-    res.redirect('/');
+    res.redirect('https://snapcraft.io/logout');
```
bartaz: While I kind of understand why we want do to it, I'm afraid it may lead to some unexpected redirect issues or raise conditions.

Lukewh: Yeah, I want to catch this as part of snapcraft.io. I think it could be an issue, but the other worry is the inconsistency of 'logout' not actually logging out of everything and confusing the user.

src/common/components/header/index.js
Lukewh: I'm wondering if there's a better title then "Build with Github", but I've updated the link as suggested.

src/common/components/header/index.js
Lukewh: Should be fixed now :)!

src/common/components/header/index.js
```diff
@@ -2,73 +2,104 @@ import React, { Component, PropTypes } from 'react';
...
+                    <li className={ style['p-navigation__link']} role="menuitem">
+                      <a href="/auth/logout" onClick={ this.onLogoutClick.bind(this)}>Sign out</a>
+                    </li>
+                  </ul>
+                </li>
+              </ul>
+            :
+              <ul className={ style['p-navigation__links--right']} role="menu">
+                <li className={ style['p-navigation__link'] } role="menuitem">
+                  <a href={ `${SNAPCRAFT_URL}/login?next=/login/bsi` }>
```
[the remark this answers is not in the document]
Lukewh: Correct. Part of the reason it has 'Don't merge' is so that we can run it by people and gauge thoughts. In the future when we eventually move BSI into SI this will also be the case!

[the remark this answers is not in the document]
bartaz: Also, now that BSI login goes through snapcraft.io login we can't deploy BSI before snapcraft.io, because it will be impossible to log in to BSI before snapcraft.io goes live. Especially that it seems that `/login/bsi` part on snapcraft.io side was made in the same PR as rest of logout redirect code that was meant to be safely deployed after.

I know that if we coordinate it well we could deploy snapcraft.io quickly enough to minimise possibility of this issue affecting anyone, but in case of any deployment issues we will end up with BSI deployment broken until snapcraft.io is deployed, which we should avoid if possible.

If we want to keep BSI login going through snapcraft.io/login - snapcraft.io login code should be deployed first as a small update, followed by BSI, followed by snapcraft.io logout deployment. So that at any point both BSI and snapcraft.io would be fully functional.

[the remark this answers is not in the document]
Lukewh: This has now been removed :)

environments/test.env
```diff
@@ -3,6 +3,8 @@ ENVIRONMENT=test
 
 BASE_URL=http://localhost:8000
 
+SNAPCRAFT_URL=https://snapcraft.io
+
```
bartaz: `+` from diff got copied in here.

.gitignore
```diff
@@ -108,6 +108,7 @@ version-info.txt
 
 # Environment variables
 environments/development.env
+environments/dev.env
```
bartaz: Nit picking a bit, but do we need another one here? there is already development.dev

src/common/components/header/index.js
```diff
@@ -2,74 +2,106 @@ import React, { Component, PropTypes } from 'react';
...
+
+    this.state = {
+      showUserDropdown: false
+    };
+  }
+
   render() {
     const { authenticated, user } = this.props;
 
+    const SNAPCRAFT_URL = conf.get('SNAPCRAFT_URL');
```
bartaz: This could be moved to top of the module, no need to read it from conf on every render.

src/common/style/vanilla/navigation.scss
```diff
@@ -60,15 +73,55 @@ $color-navigation-background: #252525;
   // right aligned sign-in navigation
   .p-navigation__links--right {
     @extend .p-navigation__links;
+    margin-right: 20px;
+
+    i {
```
bartaz: If possible this could be more specific - maybe a full icon class name?

src/common/style/vanilla/navigation.scss
```diff
@@ -60,15 +73,55 @@ $color-navigation-background: #252525;
...
+    }
 
     &:last-of-type {
       border-right: none;
     }
+
+    .p-dropdown__toggle {
+      cursor: pointer;
+
+      i {
```
bartaz: This as well (similar to `i` above) - not sure if that's the same icon, or different one. Ideally - class name selector to make it clear.

.gitignore
Lukewh: The `README` says to create `dev.env` so we should either change the README to be `development.env` or leave this ignore in... Or remove the `development.env` line above? Wdyt?

cjwatson: In any case this sort of unrelated change should be in a separate PR, as a courtesy to other developers trying to work out why things changed later on.

Lukewh: Good point @cjwatson removed

src/common/components/header/index.js
Lukewh: done

src/common/style/vanilla/navigation.scss
Lukewh: done, thanks for the help on IRC

src/common/style/vanilla/navigation.scss
Lukewh: done as above

src/common/components/header/index.js
```diff
@@ -2,74 +2,105 @@ import React, { Component, PropTypes } from 'react';
...
-                </ul>
-              )
-            }
-          </nav>
+
+          <a href="#navigation" className={ style['p-navigation__toggle--open'] } title="menu">Menu</a>
+          <a href="#navigation-closed" className={ style['p-navigation__toggle--close'] } title="close menu">Close menu</a>
         </div>
+        <nav className={ style['p-navigation__nav'] } role="menubar">
+          <ul className={ style['p-navigation__links']} role="menu">
```
bartaz: I see that some old inconsistency got copied around and spreads too much. Could we fix the missing space between closing brackets `]}`? It's around in at least several classNames now...

.env
```diff
@@ -2,4 +2,5 @@ PORT=8083
 EXTRA_PORTS='8082'
 WEBPACK_DEV_URL=http://0.0.0.0:8082
 BASE_URL=http://0.0.0.0:8083
-OPENID_VERIFY_URL=http://0.0.0.0:8083/login/verify
\ No newline at end of file
+OPENID_VERIFY_URL=http://0.0.0.0:8083/login/verify
+SNAPCRAFT_URL=https://snapcraft.io
```
bartaz: I think this `.env` is only used when running with run script. Would be good to add `SNAPCRAFT_URL` to `src/config/defaults.js` as well, to make sure some default value is set.

Lukewh: Done, thanks. Didn't know that was the case :+1:

PR #1130, opened 2018-08-03
Upgrade to sinon 6.1.4

cjwatson:
sinon added deep comparison of errors way back in 2.0.0 or so, which
needed a small test adjustment.


cjwatson:
I did this as part of another branch that turned out not to quite work in that form, but regardless of that we might as well bring this dependency up to date.  This only affects tests and doesn't require QA.

Issue #556, opened 2017-03-28
build.snapcraft.io should honour target-architectures when that lands in snapcraft


popey:
My yaml has:-
architectures: [amd64]
Because I know it fails to build on armhf, and I'm only testing on amd64, so pointless wasting cpu cycles on 40 min armhf builds. 
I realise build is explicitly limited to amd64 and armhf (and currently not i386), but it would be nice if I could force _not_ to build on armhf for example.

flexiondotorg:
I am running into the same issue for a different use case. A snapcraft.yaml that ingests a tarball from CI and said tarball is amd64 only, but there are now builds in the store for armhf and amd64.

evandandrea:
Also encountered [here](https://forum.snapcraft.io/t/how-to-trigger-builds-on-launchpad-upon-upstream-repository-update/643)

cjwatson:
Sadly this isn't what the `architectures` field means in `snapcraft.yaml`.  Setting that field declares that you're building a single multi-architecture snap that works on that set of architectures: you get a `*_multi.snap` with that set of architectures in its `meta/snap.yaml`.  That's not at all what's wanted here.

I agree that we should have the feature of being able to declare what architectures you want the snap to be built for, but it can't be done with the existing `architectures` field.

evandandrea:
Sorry for the confusion. I just spoke with Sergio and Snapcraft will gain an option for specifying what architectures a snap should be built on:
https://lists.ubuntu.com/archives/snappy-app-devel/2016-April/000654.html

evandandrea:
Another reason for letting the developer limit the target architectures in the snapcraft.yaml, [anbox's build badge shows as "failing"](https://build.snapcraft.io/user/anbox/anbox) because it's (seemingly) not intended to work on ARM.

evandandrea:
@sergiusens, can you estimate when this might land in snapcraft?

sergiusens:
Let me bring the mailing list post into the forums.

elopio:
For reference, this is the discussion in the forum:
https://forum.snapcraft.io/t/snapcraft-build-on-hint-for-builders/939

matthewpaulthomas:
Looks like the current dependency is [snapcraft#1681](https://github.com/snapcore/snapcraft/issues/1681).

jdxcode:
because https://github.com/snapcore/snapcraft/issues/1681 is merged does that mean this is available now?

[...]

PR #1150, opened 2018-09-27
Move "Build trigger" column before "Architecture"

cjwatson:
The "Build trigger" column is currently missing from the design
specification.  Design work for #556 calls for failed build requests to
have their request ID and error message rendered spanning the
"Architecture", "Duration", and "Result" columns; since failed build
requests still have a meaningful build trigger, having "Build trigger"
in the middle of this group of columns makes it hard to comply with that
otherwise-reasonable suggestion.

Moving "Build trigger" two columns to the left resolves this difficulty,
and seems to make just as much sense.  In fact, once we start using
build requests, it arguably makes more sense (because the build trigger
is known before the set of built-on architectures is known).

### DIFF
--- a/src/common/components/build-row/index.js
+++ b/src/common/components/build-row/index.js
@@ -75,14 +75,14 @@ const BuildRow = (props) => {
           : <span>{`#${buildId}`}</span>
         }
       </Data>
+      <Data col="20">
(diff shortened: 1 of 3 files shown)


Lukewh:
Looks good thanks @cjwatson :+1:

PR #1151, opened 2018-10-01
Simplify internalGetSnapBuilds return value

cjwatson:
`internalGetSnapBuilds` previously returned a somewhat mangled
`Collection` (a type used by the Launchpad web service client that
automatically handles batching and iteration of large collections).
However, this didn't make much sense here because the mangling meant
that the automatic batching wouldn't work quite right, and the callers
all just picked out the `entries` array anyway; it will make even less
sense as a `Collection` after upcoming work for #556.

We now just return an array instead.

### DIFF
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -685,24 +685,32 @@ export const authorizeSnap = async (req, res) => {
 };
 
 export async function internalGetSnapBuilds(snap, start = 0, size = 10) {
-  const builds = await getLaunchpad().get(snap.pending_builds_collection_link, {
-    start: start, size: size
-  });
+  const builds = [];
+  let gotItems = 0;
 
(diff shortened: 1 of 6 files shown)


src/server/handlers/launchpad.js
```diff
@@ -817,7 +828,7 @@ export const requestSnapBuilds = async (req, res) => {
     const reason = req.body.reason || BUILD_TRIGGERED_MANUALLY;
     const snap = await internalFindSnap(req.body.repository_url);
     const builds = await internalRequestSnapBuilds(snap, owner, name, reason);
-    const build_annotations = await internaGetBuildAnnotations(builds);
+    const build_annotations = await internalGetBuildAnnotations(builds);
```
bartaz: Thanks for picking up on the typos.

PR #1152, opened 2018-10-01
Prepare UI for upcoming build request changes

cjwatson:
We will shortly need to change new snap build requests to be dispatched
more asynchronously, in order that Launchpad can fetch `snapcraft.yaml`
from the repository and decide which architectures the snap should be
built for.  This means that a snap can be in a state where it has
pending build requests (different from pending builds: for instance,
they don't have an architecture), and also where it has failed build
requests in its history (for example, because `snapcraft.yaml` is
malformed in such a way that Launchpad can't determine which
architectures it should be built for).

This prepares the UI layer for that work by ensuring that it can
understand the modified server responses and render build requests in
the builds list.  The server side hasn't been changed yet (that will be
a follow-up PR), so this should result in no functional change for the
moment.

Build requests will be interleaved with builds in server responses, and
the UI tells them apart by looking at the `resource_type_link`.  This
wasn't my first choice for designing the internal API, but it turns out
to be the simplest option: in particular, if we want to be able to
support sensible pagination of the combined build and build request
history in future then we need to be iterating over a single collection.

Part of #556.

### DIFF
--- a/src/common/components/build-row/index.js
+++ b/src/common/components/build-row/index.js
@@ -42,6 +42,7 @@ const getBuildTriggerMessage = (repository, reason, commitId) => {
 const BuildRow = (props) => {
 
   const {
+    isRequest,
     repository,
     architecture,
     buildId,
@@ -49,6 +50,7 @@ const BuildRow = (props) => {
     duration,
     colour,
     statusMessage,
+    dateCreated,
     dateStarted,
     reason,
     commitId,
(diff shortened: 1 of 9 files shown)


src/common/components/build-row/index.js
```diff
@@ -67,33 +69,57 @@ const BuildRow = (props) => {
...
-        />
-      </Data>
-    </Row>
+  const buildTriggerCell = (
+    <Data col="20">
+      { getBuildTriggerMessage(repository, reason, commitId) }
+    </Data>
   );
+
+  if (isRequest) {
```
bartaz: Seems like row for build request row is quite different from build row, so maybe it would be worth creating separate `RequestRow` component for that?
This should allow to avoid having `BuildRow` returning 2 different things depending on the build/request type. Also maybe we can make `RequestRow` get proper params to avoid detecting error based on status message comparison.

cjwatson: That was what I did in my first pass.  When I worked out that the internal API needs to merge builds and requests I merged them back into a single component, but maybe you're right that they'd be better separated.  I'll look into that.

The `statusMessage` comparison is IMO not as bad as it looks, because that's effectively an enumerated list controlled by BSI (see `UserFacingState`), and there's some precedent such as in `src/common/containers/build-details.js`.  I agree that it would be nicer to have this be something that's more obviously an enum rather than display text, but since it's following an existing pattern I think we could reasonably do that in a separate PR?

bartaz: Yes, I just looked into the build object and it seems we don't have anything better then `statusMessage` in there to do the comparison. So we can leave it as it is or address it later.

This also makes separating `BuildRow` and `RequestRow` less important, so I'll let you decide if that still makes sense. Thanks!

cjwatson: I've separated them out now.  Thanks for the suggestion.

PR #1158, opened 2018-10-04
Request new builds using build requests

### LINKED ISSUE #556: build.snapcraft.io should honour target-architectures when that lands in snapcraft

cjwatson:
This makes Launchpad responsible for asynchronously figuring out which
architectures the snap should be built for, thereby allowing us to
support Snapcraft's `architectures` keyword in `snapcraft.yaml`.

Fixes #556.

## QA

We'll need to walk through the whole workflow of requesting new builds to make sure that everything still works.

### DIFF
--- a/docs/routes.md
+++ b/docs/routes.md
@@ -106,8 +106,10 @@ To request builds of an existing snap:
       "repository_url": "https://github.com/:owner/:name"
     }
 
-On success, returns the following, where the items in `builds` are
+On success, returns the following, where the items in `builds` are either
 [snap\_build entries](https://launchpad.net/+apidoc/devel.html#snap_build)
+or
+[snap\_build\_request entries](https://launchpad.net/+apidoc/devel.html#snap_build_request)
 as returned by the Launchpad API:
 
     HTTP/1.1 201 Created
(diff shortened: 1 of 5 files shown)


bartaz:
While testing locally: for a build requested manually I got a build request (which was 'pending' for w while) with proper 'Manual' as a trigger, and then builds started, but the trigger for them is listed as 'Unknown'. Should builds take the trigger from their request?

screen shot 2018-10-04 at 15 03 42

bartaz:
Looking at the contents of local db I can see the build request stored in the database, but not the builds triggered by it. Will investigate further.

cjwatson:
The reason this doesn't work locally is that the link between the build and the build request is populated by the webhook handler, and GitHub can't make webhook deliveries to your local deployment.  If you manually insert `BuildAnnotation(build_id, request_id)` rows (see the second part of the 'created' case in `handleLaunchpadSnapBuild`) then you'll see the correct build trigger.  This won't be a problem on staging or production because the webhook endpoint is reachable by GitHub in those cases.

bartaz:
@cjwatson Ah, right. You mean LP can't make webhook deliveries in this case right? As it's LP webhook that is triggered by the build?

PR #1159, opened 2018-10-04
Sync up table headings on build details page

### LINKED ISSUE #1066: Build details page has misaligned table header
## Summary

Build details page has misaligned table header, because 'Build trigger' column header is missing.

## Process

- Go to any build details page: https://build.snapcraft.io/user/bartaz/speed-test-snap/59536
- Build details table heading is misaligned

## Screenshot

screen shot 2018-02-28 at 09 44 46

cjwatson:
This now uses the BuildsList component, so it shouldn't get out of sync
again.

Fixes #1066.

## QA

Look at a build details page; the table headings should match, and the cell under "Number" should remain as plain text rather than being a self-link.

### DIFF
--- a/src/common/containers/build-details.js
+++ b/src/common/containers/build-details.js
@@ -2,8 +2,7 @@ import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
 import Helmet from 'react-helmet';
 
-import BuildRow from '../components/build-row';
-import { Table, Head, Body, Row, Header } from '../components/vanilla/table-interactive';
+import BuildsList from '../components/builds-list';
 import BuildLog from '../components/build-log';
 import {
   HelpBox,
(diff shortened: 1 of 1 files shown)


src/common/containers/build-details.js
```diff
@@ -92,23 +91,14 @@ class BuildDetails extends Component {
...
-              </Head>
-              <Body>
-                <BuildRow
-                  repository={repository}
-                  {...build}
-                  isLinked={false} // disable link to the same build we are already looking at
-                />
-              </Body>
-            </Table>
+            <BuildsList
```
bartaz: Wow, why haven't we done this like that before? :) Good change, thanks!

src/common/containers/build-details.js
```diff
@@ -92,23 +91,14 @@ class BuildDetails extends Component {
...
+              repository={repository}
+              builds={[{
+                ...build,
+                // Disable link to the same build we're already looking at.
+                isLinked: false
```
bartaz: This part is a little bit hacky. It depends on internal implementation of `BuildsList` and the fact that it passes whole build object to `BuildRow` as props.

But on the other hand it's the simplest way to avoid duplication between builds list and build details pages, so let's keep it.

cjwatson: Yeah, it's not entirely perfect, but as you say.  (For the future, it might be better to explicitly pass down `isLinked` as an option to `BuildsList`.)

PR #1162, opened 2018-10-07
Fetch snapcraft.yaml before requesting builds

cjwatson:
It's possible for a repository to have once been configured on
build.snapcraft.io, but for the contributors to that repository to later
remove snapcraft.yaml for whatever reason.  We have a few cases of this
on production today, and they generate a large volume of email
notifications to the operators of the ~build.snapcraft.io Launchpad
account because there ends up being a failed build request for every
commit to one of the repositories in question.

The simplest way to avoid this is to always check for a snapcraft.yaml
before dispatching automatic builds of a snap (webhook or poller).  This
does have the downside that the webhook won't automatically cause failed
build requests if snapcraft.yaml is present but syntactically invalid;
but if somebody is trying to debug that then they'll probably try to
dispatch a manual build, and the problem will become clear at that
point, at least once we start displaying recent failed build requests.

While I was here, I took the opportunity to make some of the poller
tests a bit more precise about expected URLs.


cjwatson:
> But just to be clear - so, the downside of this change is that if snapcraft.yaml is present, but invalid (loading YAML would fail) then the whole webhook call will fail and no builds will be requested (in a same way as if snapcraft.yaml would be unavailable)?

That's correct.

bartaz:
Ok. Thanks. If this becomes a problem we can find a way to better differentiate between unavailable snapcraft.yaml and invalid one. Which I guess should be possible now if we check at the details of the error being thrown.

PR #1167, opened 2018-10-17
Upgrade to react-hot-loader 4

cjwatson:
Among other things, this is needed in order to support a later upgrade
to babel 7; and the new API is rather more compact.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -59,12 +59,6 @@
     "aws4": {
       "version": "1.5.0"
     },
-    "babel-code-frame": {
-      "version": "6.16.0"
-    },
-    "babel-messages": {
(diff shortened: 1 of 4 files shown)


bartaz:
@cjwatson What do you mean by using old API for root reducer? I don't think react hot loader is used anywhere else apart from Root component.

cjwatson:
> @cjwatson What do you mean by using old API for root reducer? I don't think react hot loader is used anywhere else apart from Root component.

I was thinking of https://github.com/canonical-websites/build.snapcraft.io/blob/master/src/common/store/configureStore.js, but now I look again I see that's doing its own hot module handling rather than using react-hot-loader.  I'll amend the commit message to be less confusing.

bartaz:
Thanks!

PR #1168, opened 2018-10-17
Upgrade to enzyme 3

cjwatson:
This will make upgrading React somewhat easier in future.

See: https://airbnb.io/enzyme/docs/guides/migration-from-2-to-3.html


test/unit/src/common/components/repositories-home/t_repositories-home.js
```diff
@@ -39,7 +39,9 @@ describe('The RepositoriesHome component', () => {
         router: {}
       };
 
-      wrapper = shallow(<RepositoriesHome { ...props } />);
+      wrapper = shallow(
+        <RepositoriesHome { ...props } />, { disableLifecycleMethods: true }
```
bartaz: You are only disabling it for tests that were failing with this enabled?

cjwatson: Correct.  (Would you prefer I disabled it globally?  I don't have a strong opinion.)

bartaz: On one hand turning it off globally would keep it consistent with version we've been using, but if the change affected only couple of tests (as it seems), I guess it's fine to stick with new default and only update where needed. So new tests can be written with default enzyme behaviour from the start.

So I'm fine with keeping it as it is. Thanks.

cjwatson: Thanks for articulating something equivalent to my rather vague intuition :-)

PR #1172, opened 2018-11-15
.gitignore environments/dev.env

Lukewh:
## Done

The README says to use `environments/dev.env` but the `.gitignore` only ignores `environments/development.env`. This add's `dev.env` to the `.gitignore` for new users and leaves `development.env` ignored for anyone that still uses that file.

As seen here: https://github.com/canonical-websites/build.snapcraft.io/pull/1129#pullrequestreview-143883946

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,6 +108,7 @@ version-info.txt
 
 # Environment variables
 environments/development.env
+environments/dev.env
 
 # Development and test SQLite databases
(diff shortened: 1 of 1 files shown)


Lukewh:
@cjwatson, I hear you and I'd agree but I'm just trying to mitigate frustration for anyone that has committed to `development` or `dev`

PR #1173, opened 2018-11-16
If in development, don't redirect to snapcraft.io

Lukewh:
## Done

When `NODE_ENV=devleopment`:
- Don't redirect `/` to (staging.)snapcraft.io/build
- On log out don't redirect to (staging.)snapcraft.io/logout

## QA

- Ensure you have `NODE_ENV=development` set in `environments/dev.env`
- You should see the original build homepage (no redirect)
- Authenticate with Github using the green Github button on the homepage
- Once logged in, log out and you should be returned to the original build homepage (no redirect)

## Issue / Card
Fixes https://github.com/canonical-websites/build.snapcraft.io/issues/1153

### DIFF
--- a/src/server/handlers/login.js
+++ b/src/server/handlers/login.js
@@ -118,8 +118,12 @@ export const logout = (req, res, next) => {
       // TODO log errors to sentry
       return next(new Error(constants.E_LOGOUT_FAIL));
     }
-    // FIXME redirect to page that initiated the sign in request
-    res.redirect('/');
+    if (process.env.NODE_ENV !== 'development') {
+      res.redirect(`${SNAPCRAFT_URL}/logout?no_redirect=true`);
+    } else {
+      // FIXME redirect to page that initiated the sign in request
+      res.redirect('/');
+    }
   });
 };
 
--- a/src/server/routes/universal.js
+++ b/src/server/routes/universal.js
@@ -4,7 +4,9 @@ import { universal, homepage } from '../handlers/universal';
 
 const router = Router();
 
-router.get('/', homepage);
+if (process.env.NODE_ENV !== 'development') {
+  router.get('/', homepage);
+}
 router.get('*', universal);
 
 export default router;
--- a/src/server/handlers/github-auth.js
+++ b/src/server/handlers/github-auth.js
@@ -146,7 +146,11 @@ export const logout = (req, res, next) => {
     if (err) {
       return next(new Error('Failed to log out.'));
     }
-    // FIXME redirect to page that initiated the sign in request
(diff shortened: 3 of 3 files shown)


bartaz:
Tests fail:

```
/home/travis/build/canonical-websites/build.snapcraft.io/src/server/handlers/login.js
  122:23  error  'SNAPCRAFT_URL' is not defined  no-undef
✖ 1 problem (1 error, 0 warnings)
```

src/server/handlers/login.js
```diff
@@ -118,8 +118,12 @@ export const logout = (req, res, next) => {
       // TODO log errors to sentry
       return next(new Error(constants.E_LOGOUT_FAIL));
     }
-    // FIXME redirect to page that initiated the sign in request
-    res.redirect('/');
+    if (process.env.NODE_ENV !== 'development') {
```
bartaz: Is this needed here? If home page is not redirecting further, redirecting to `/` (as it currently is) should be still fine(?)

src/server/routes/universal.js
```diff
@@ -4,7 +4,9 @@ import { universal, homepage } from '../handlers/universal';
 
 const router = Router();
 
-router.get('/', homepage);
+if (process.env.NODE_ENV !== 'development') {
```
bartaz: It feels a bit weird to be adding this conditional on this level.
Wouldn't it make more sense to have devel/prod logic inside the `homepage` handler (https://github.com/canonical-websites/build.snapcraft.io/blob/master/src/server/handlers/universal.js#L12-L14)?

Lukewh: I originally had in the homepage handler but figured short-circuiting earlier would be better. Happy to move it if you think it would be better in the handler.

bartaz: Yes please, I think this file is good just to list the routes without any logic. We already have some ENV checks inside the handlers, so for consistency lets move it there as well.

src/server/handlers/login.js
Lukewh: Thanks, I wasn't sure about this handler, as it seems only the github one is used

src/server/routes/universal.js
```diff
@@ -4,9 +4,7 @@ import { universal, homepage } from '../handlers/universal';
 
 const router = Router();
 
-if (process.env.NODE_ENV !== 'development') {
```
bartaz: Sorry, I agree that this solution looks cleaner than doing it in inside of the route and calling `universal` from there.
Just change the name from `homepage` to `homepageRedirect` to make it more clear that "if not in dev env, redirect home page".

Lukewh: Updated :+1:

PR #1202, opened 2019-03-08
Add global nav

bartaz:
## Done

Adds Canonical global nav (and fixes related styling conflicts).

## QA

- Check out this feature branch
- Run the site using the command ```npm start -- --env=environments/dev.env```
- View the site locally in your web browser at: [http://0.0.0.0:8000/](http://0.0.0.0:8000/)
- Global nav should appear on top
- Open/close global nav 'Products' dropdown
- Navigate to different BSI pages, see if global nav works as expected
- Check if there are no styling issues, z-index problems, etc

## Issue / Card

Fixes #1183 

## Screenshots

screenshot 2019-03-08 at 12 14 47

PR #1204, opened 2019-03-11
Propely namespace all global-nav occurences

bartaz:
## Done

#1202 added @canonical/global-nav as a dependency, but it was still imported as `global-nav` (without namespace) causing failures on clean deployments.
This PR fixes this by properly namespacing all occurrences of global-nav.

## QA

- `rm -rf node_modules`
- `npm install`
- Everything should build without errors, site with global nav should be displayed

### DIFF
--- a/src/common/index.js
+++ b/src/common/index.js
@@ -2,7 +2,7 @@
 import '../../node_modules/normalize.css/normalize.css';
 import './style/base.css';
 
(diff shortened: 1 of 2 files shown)


bartaz:
@Lukewh deploying to staging DID catch this, because I only found out by attempting QA on staging before prod release

PR #1206, opened 2019-03-19
Load all snaps (when there is more than one batch)

### LINKED ISSUE #1181: App disappeared from my app list
## Summary

I need to trigger a build of atom, as there's been an upstream update. I went to build.snapcraft.io and logged in. Atom is no longer in the list. It certainly was, because I've built it numerous times.  I have had to re-add it again (which is spinning forever)...

image

## Process

I don't know how to reproduce this, but I do sometimes see apps gone from the list which should be there.

Visited build.snapcraft.io, logged in, expected to see all the applications I've added. Actually saw atom missing.

bartaz:
## Done

As discussed in https://github.com/canonical-websites/build.snapcraft.io/issues/1181#issuecomment-444050083 we now correctly load all the snaps (not just the first batch).

## QA

To be able to reproduce the original issue and check if it's fixed without the need to have ~80 repos configured small local code change is needed to reduce the API batch size in dev environment.

In [`src/server/handlers/launchpad.js`](https://github.com/canonical-websites/build.snapcraft.io/blob/master/src/server/handlers/launchpad.js#L384-L389) change `internalFindSnaps`  to add `size` parameter to "findByURLPrefixes" API call:

```js
  const result = await lpClient.named_get('/+snaps', 'findByURLPrefixes', {
    parameters: {
      url_prefixes: remainingPrefixes,
      owner: `/~${username}`
    },
    size: 5
  });
```

Size should be smaller then number of repos you have configured, to make sure more then one batch needs to be fetched.

#### Reproduce the issue locally

- check out master branch
- update code locally to reduce size of the batch (see above) to a number smaller then all of your configured repos
- sign in to load your "My repos" page
- not all repos should be visible (only the first batch with size defined before)

#### QA the fix

- update code locally to reduce size of the batch (see above) to a number smaller then all of your configured repos
- sign in to load your "My repos" page
- All the repos (more then one batch) should be loaded and displayed

## Issue / Card

Fixes #1181

### DIFF
--- a/src/server/handlers/launchpad.js
+++ b/src/server/handlers/launchpad.js
@@ -387,10 +387,17 @@ const internalFindSnaps = async (owner, token) => {
       owner: `/~${username}`
     }
   });
+
+  // The response may be split across multiple batches.  Gather them all.
+  const entries = [];
+  for await (const entry of result) {
+    entries.push(entry);
+  }
+
   // Split up results into separate cache entries so that they can help
   // speed things up for other users in the same organizations.
   const newSnapsByPrefix = remainingPrefixes.map((urlPrefix) => {
-    return result.entries.filter(
+    return entries.filter(
       (snap) => snap.git_repository_url.startsWith(urlPrefix)
     );
   });
@@ -399,7 +406,7 @@ const internalFindSnaps = async (owner, token) => {
       getUrlPrefixCacheId(urlPrefix), newSnaps, 3600
     )
   ));
-  snaps = snaps.concat(result.entries);
+  snaps = snaps.concat(entries);
   // ensure LP webhook is set up for all the snaps
   // but batch it by 10 not to overload LP servers
   const BATCH_SIZE = 10;


src/server/handlers/launchpad.js
```diff
@@ -387,10 +387,17 @@ const internalFindSnaps = async (owner, token) => {
       owner: `/~${username}`
     }
   });
+
+  // The response may be split across multiple batches.  Gather them all.
+  const entries = [];
+  for await (const entry of result) {
+    entries.push(entry);
+  }
```
cjwatson: The batching exists for a reason, and the same reason applies to build.snapcraft.io.  This is an unboundedly large set, and could easily result in timeouts.

Instead, we need to have corresponding batching for this build.snapcraft.io endpoint.  But in order to do that in a stable way, I think we need to be able to do cursor-based navigation of Launchpad's snap build batches, and Launchpad doesn't support that yet.  That's why I hadn't already done this.  I made a start on the Launchpad side last September, but my branch still needs more work.

cjwatson: Oh actually, this is just snaps for a prefix isn't it, not snap builds?  That's probably OK then, as while it may well go over a single batch it won't be arbitrarily long in practice.  Carry on.

bartaz:
@cjwatson Thanks for comments. Yes, this one is for snap list only, so it is based on assumption this list will rarely be extremely long.

For builds it's a different story, that we will hopefully fix one day and this will obviously need batching from the front-end side. And indeed, cursor based 'pagination' would make it easier to implement reliably.

PR #1258, opened 2020-04-01
Catch webhook db errors

### LINKED ISSUE #1257: Webhook deliveries fail when build was requested by new site
If a build was requested by the new snapcraft.io site, then (of course) no `BuildRequestAnnotation` row is created in build.snapcraft.io.  However, a webhook delivery is still made to build.snapcraft.io if the snap was created before the new site existed.  This delivery attempt returns HTTP 500 because it fails to create a `BuildAnnotation` row since it has no `BuildRequestAnnotation` to link it to.

Speaking as the author of the code that creates a `BuildAnnotation`, it was a mistake that it returns 500 in this case.  If you compare `src/server/handlers/launchpad.js:internalRequestSnapBuilds` with `src/server/handlers/webhook.js:handleLaunchpadSnapBuild`, you'll see that both claim to do best-effort creation of database rows, but only the former has an exception handler that makes it be actually best-effort.

The consequence of the HTTP 500s is that webhook delivery attempts get backed up in Launchpad; so it would be very helpful to fix this as part of the transition to the new site.

jkfran:
## Done

Fixes #1257

### DIFF
--- a/src/server/handlers/webhook.js
+++ b/src/server/handlers/webhook.js
@@ -104,14 +104,21 @@ const handleLaunchpadSnapBuild = async (req, res, owner, name, parsedBody) => {
       if (parsedBody.build_request) {
         // Record a build annotation so that we can determine why this build
         // was dispatched.  As above, this is best-effort.
-        await db.transaction(async (trx) => {
-          await db.model('BuildAnnotation')
-            .forge({
-              build_id: getLinkId(parsedBody.snap_build),
-              request_id: getLinkId(parsedBody.build_request)
-            })
-            .save({}, { method: 'insert', transacting: trx });
-        });
+        try {
+          await db.transaction(async (trx) => {
+            await db.model('BuildAnnotation')
+              .forge({
+                build_id: getLinkId(parsedBody.snap_build),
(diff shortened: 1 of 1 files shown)


src/server/handlers/webhook.js
```diff
@@ -104,14 +104,21 @@ const handleLaunchpadSnapBuild = async (req, res, owner, name, parsedBody) => {
...
+              .forge({
+                build_id: getLinkId(parsedBody.snap_build),
+                request_id: getLinkId(parsedBody.build_request)
+              })
+              .save({}, { method: 'insert', transacting: trx });
+          });
+        } catch (error) {
+          logger.error(
+            'Error saving build request annotation for ' +
+            `${parsedBody.build_request}: ${error}`
```
cjwatson, line 119: The error is in saving a build annotation, not a build request annotation, so this error message isn't quite right.  By the same token, it probably ought to refer to `parsedBody.snap_build` rather than `parsedBody.build_request`.

jkfran: It makes sense. Let me know if it's okay now.

PR #1260, opened 2020-04-15
Remove poller cronjob

jkfran:
## Done

Remove poller cronjob


jkfran:
Hey @cjwatson , could you confirm this is correct?

cjwatson (changes requested):
You could also remove a few other bits: the rendering of `snap-build-poller_secrets.j2` in `configure`, the associated calls to `hookenv.config` (searching for `poller` should find them), and the actual template files in `charm/templates/snap-build-poller_*`.

We'd ideally also remove the `poller_*` configuration items from `charm/config.yaml`, but we'd have to first ensure that the Mojo spec isn't attempting to set them any more, so it probably isn't possible to do that in the same PR.

jkfran:
Hi @cjwatson ,

I did this changes, I couldn't find more associated calls to hookenv.config. Please let me know.

Should I open another PR to update `charm/config.yaml` after this one?

Thank you!

cjwatson (approved):
lp:~ubuntuone-pqm-team/canonical-mojo-specs/mojo-ols-snap-build would have to be updated to remove poller references before removing anything from `charm/config.yaml`.  Is it worth bothering with that sort of extra cleanup given that this service is presumably on a gradual path to decommissioning now anyway?

jkfran:
@cjwatson No, I guess it doesn't.

Thank you for your time reviewing this.

Lukewh (approved):

